PR #50, opened 2022-09-09
Build: Support for building nested plugins

### LINKED ISSUE #33: Feature: Support plugin mono-repos
We have had people ask on slack and in Github to make toolkit work within a mono-repo.  Rather than invest time in getting toolkit to support mono-repos I think adding mono-repo support to this package would get quite a bit of buy in from plugin developers.

We could likely use this [monorepo of plugins](https://github.com/jackw/grafana-federated-pluginss) as reference to add this feature. All plugins within it  are "scaffolded" or "migrated" by this project. They have then been manually tweaked to build within a mono-repo. However right now tests and e2e tests have not been catered for. We can probably ignore the webpack config extensions to each plugin too.

mckn:
This PR will add support for building nested plugins to the generated webpack configuration.

I have changed the exported webpack config to return a `Promise<Configuration>` instead of a `Configuration`. Will this cause any issues for plugins overriding the configuration?

Fixes #33

### DIFF
--- a/templates/common/.config/webpack/utils.ts
+++ b/templates/common/.config/webpack/utils.ts
@@ -1,6 +1,10 @@
 import fs from 'fs';
 import path from 'path';
+import util from 'util';
+import glob from 'glob';
 import { SOURCE_DIR } from './constants';
+
+const globAsync = util.promisify(glob);
 
 export function getPackageJson() {
   return require(path.resolve(process.cwd(), 'package.json'));
@@ -14,4 +18,24 @@ export function getPluginId() {
 
 export function hasReadme() {
   return fs.existsSync(path.resolve(process.cwd(), SOURCE_DIR, 'README.md'));
+}
+
+export async function getEntries(): Promise<Record<string, string>> {
+  const parent = `..${path.sep}`;
+  const pluginsJson = await globAsync('**/src/**/plugin.json');
+  
+  const plugins = await Promise.all(pluginsJson.map(pluginJson => {
+    const folder = path.dirname(pluginJson);
+    return globAsync(`${folder}/module.{ts,tsx,js}`);
+  }));
+
+  return plugins.reduce((result, modules) => {
+    return modules.reduce((result, module) => {
+      const pluginPath = path.resolve(path.dirname(module), parent);
+      const pluginName = path.basename(pluginPath);
+  
+      result[`${pluginName}/module`] = path.join(parent, module);
+      return result;
+    }, result);
+  }, {});
 }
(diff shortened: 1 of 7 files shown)


templates/common/.config/webpack/webpack.config.ts
```diff
@@ -12,10 +12,10 @@ import path from 'path';
 import ReplaceInFileWebpackPlugin from 'replace-in-file-webpack-plugin';
 import { Configuration } from 'webpack';
 
-import { getPackageJson, getPluginId, hasReadme } from './utils';
-import { SOURCE_DIR, DIST_DIR, ENTRY_FILE } from './constants';
+import { getPackageJson, getPluginId, hasReadme, getEntries } from './utils';
+import { SOURCE_DIR, DIST_DIR } from './constants';
 
-const config = (env): Configuration => ({
+const config = async (env): Promise<Configuration> => ({
```
jackw, line 18: If we choose to do this we will need to update the docs as well to include how to use webpack-merge with a promise based webpack config.

templates/common/.config/webpack/utils.ts
```diff
@@ -15,3 +19,23 @@ export function getPluginId() {
 export function hasReadme() {
   return fs.existsSync(path.resolve(process.cwd(), SOURCE_DIR, 'README.md'));
 }
+
+export async function getEntries(): Promise<Record<string, string>> {
+  const parent = `..${path.sep}`;
```
jackw, line 24: I don't think there's a need for a separator as `path.resolve` and `path.join` should handle that internally.

templates/common/package.json
```diff
@@ -50,7 +51,8 @@
     "typescript": "^4.4.0",
     "webpack": "^5.69.1",
     "webpack-cli": "^4.9.2",
-    "webpack-livereload-plugin": "^3.0.2"
+    "webpack-livereload-plugin": "^3.0.2",
+    "glob": "^8.0.3"
```
jackw, line 55: Can we move this after `eslint-webpack-plugin` so npm/yarn won't move it in the future if/when a dev runs `yarn add anotherPackage` after scaffolding their plugin?

templates/common/.config/webpack/webpack.config.ts
mckn: So the question is, should I rewrite this so it is using the sync functions? Then we don't need to change so the configuration function returns a Promise.

mckn: On the other hand. We probably need to do this sooner or later 🙈 

What docs do we have that uses the `webpack-merge`?

jackw: The [readme](https://github.com/grafana/create-plugin/blob/main/templates/common/.config/README.md) contains an example of how to extend this webpack config. I say we go with this change now as doing it later is only going to be more painful.

jackw (approved):
🚀

PR #52, opened 2022-09-27
Bugfix: building a single plugin as module directly in the dist folder

mckn:
When I introduced #50 I broke the support for single plugin repositories by always put the plugin `dist/${pluginName}/module.js`. This PR will fix that issue by always putting a single plugin repo directly below `dist/module.js`.

We should probably add tests for the most simple scenarios. Will open another PR about that.


leventebalogh (approved):
LGTM 👍

jackw (approved):
Nice! 🚀

PR #57, opened 2022-09-28
App: Adding a proxy backend for apps

### LINKED ISSUE #55: Incorrect backend template for app plugins with a backend
When generating an app plugin with a backend it gets scaffolded with a datasource plugin backend instead of an app plugin backend. We should probably change this to be scaffolded with something like:

https://github.com/grafana/incident/tree/main/plugin/pkg

mckn:
Fixes #55


mckn:
kill it with fire

PR #66, opened 2022-10-06
Templates: Set timezone to UTC for jest

### LINKED ISSUE #60: enhancement: add runtime timezone setting to jest-config.js
When using snapshots (or any tests with Date), the runtime TZ of the machine is used by jest.

adding

```CODE
process.env.TZ = 'UTC';
```

before the module.export will force the tests to generate UTC timestamps, and will work inside git workflows/drone and locally.

briangann:
Snapshots and tests that use Date() will fail when generated locally in different timezones.  This forces jest to use UTC so the results will be reliable.

fixes #60

### DIFF
--- a/templates/common/jest.config.js
+++ b/templates/common/jest.config.js
@@ -1,3 +1,7 @@
+// force timezone to UTC to allow tests to work regardless of local timezone
+// generally used by snapshots, but can affect specific tests
+process.env.TZ = 'UTC';
+
 module.exports = {
   // Jest configuration provided by Grafana scaffolding
   ...require('./.config/jest.config'),


templates/common/jest.config.js
```diff
@@ -1,3 +1,7 @@
+// force timezone to UTC to allow tests to work regardless of local timezone
+// generally used by snapshots, but can affect specific tests
+process.env.TZ = 'UTC';
```
jackw, line 3: I'm wondering if it would be easier for developers to find and maybe adjust the value if we put it in the package.json test script instead. e.g. `"test": "TZ=UTC yarn jest"`. WDYT?

briangann: from what i could tell vscode was not running `yarn test` and would use the local timezone, you'd have to launch vscode with the env var set.

i saw this when when i tried a remote dev environment over ssh that had UTC, and my local machine was EDT, and then in the drone/ci builds it was in UTC, but the snapshots were EDT.

not 100% sure where it should go.

academo: I remember fixing this the same way jack suggests but I don't use vscode and it does make sense that vscode launches the jest process on its own not using scripts.

academo (approved):

Issue #43, opened 2022-09-05
Switch to pluginvalidator2


tolzhabayev:
`Lint Plugin` action should be switched to using `plugincheck2`

**Source code:**
https://github.com/grafana/create-plugin/blob/main/templates/github/ci/.github/workflows/release.yml#L120-L126

**To consider:**
- Should we use a special `-config` in the `Lint Plugin` flow e.g. https://github.com/grafana/plugin-validator/blob/master/config/publishing.yaml?

PR #67, opened 2022-10-06
Templates: Use plugincheck2 in release workflow

jackw:
Moving the release workflow to use plugincheck2 and the publishing config.

Fixes: #43

### DIFF
--- a/templates/github/ci/.github/workflows/release.yml
+++ b/templates/github/ci/.github/workflows/release.yml
@@ -120,7 +120,7 @@ jobs:
       - name: Lint plugin
         run: |
           git clone https://github.com/grafana/plugin-validator
-          pushd ./plugin-validator/cmd/plugincheck
+          pushd ./plugin-validator/cmd/plugincheck2
           go install
           popd
           plugincheck2 -config ./plugin-validator/config/publishing.yaml $\{{ steps.metadata.outputs.archive }}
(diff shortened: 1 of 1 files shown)


templates/github/ci/.github/workflows/release.yml
```diff
@@ -120,10 +120,10 @@ jobs:
       - name: Lint plugin
         run: |
           git clone https://github.com/grafana/plugin-validator
-          pushd ./plugin-validator/cmd/plugincheck
+          pushd ./plugin-validator/cmd/plugincheck2
```
jackw, line 123: I'm confused by this path. Shouldn't it be `./plugin-validator/pkg/cmd/plugincheck2` as per the repo https://github.com/grafana/plugin-validator/tree/master/pkg/cmd/plugincheck2 ??

academo: yes it should be that. I wonder if this `pushd` is failing and the script goes on? Though `pushd` exits with code 1 when the path is not found.

tolzhabayev: It should be the path as outlined by @jackw

while we at it git clone https vs ssh?

jackw: Ok, I've updated the path. What's the benefit of using ssh over https here @tolzhabayev ? 🤔

tolzhabayev (approved):
LGTM!

templates/github/ci/.github/workflows/release.yml
tolzhabayev: No idea, just used to `git clone` using `git@` <- ignore my comment it does not seem to add any value

Issue #48, opened 2022-09-08
Migrate: Clean up dependency declarations post migration


jackw:
Yarn (and I'm guessing all package managers) sort dependencies alphabetically after adding another package to a project (e.g. `yarn add @types/lodash`). During the migration of a plugin we "add" dependencies into the package.json but the result isn't sorted.

PR #68, opened 2022-10-06
Migrate: Sort package.json fields

jackw:
This PR adds a clean up function to make sure the package.json is in a state that package managers would expect (alphabetised dependencies and devDependencies) and scripts are sorted to make it easier for a developer to scan.

| Before | After |
| --- | --- |
| Screenshot 2022-10-06 at 12 11 08 | Screenshot 2022-10-06 at 12 09 49 |
 

Fixes: #48

### DIFF
--- a/src/utils/utils.npm.ts
+++ b/src/utils/utils.npm.ts
@@ -187,3 +187,20 @@ export function updateNpmScripts() {
 
   writePackageJson(packageJson);
 }
+
+export function cleanUpPackageJson() {
+  const packageJson = getPackageJson();
+  packageJson.scripts = sortKeysAlphabetically(packageJson.scripts);
+  packageJson.dependencies = sortKeysAlphabetically(packageJson.dependencies);
+  packageJson.devDependencies = sortKeysAlphabetically(packageJson.devDependencies);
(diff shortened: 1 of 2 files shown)


src/utils/utils.npm.ts
```diff
@@ -187,3 +187,20 @@ export function updateNpmScripts() {
...
+
+export function cleanUpPackageJson() {
+  const packageJson = getPackageJson();
+  packageJson.scripts = sortKeysAlphabetically(packageJson.scripts);
+  packageJson.dependencies = sortKeysAlphabetically(packageJson.dependencies);
+  packageJson.devDependencies = sortKeysAlphabetically(packageJson.devDependencies);
+  writePackageJson(packageJson);
+}
+
+function sortKeysAlphabetically(obj: Record<string, string>) {
```
jackw, line 199: Maybe use a lib (`ramda` / `lodash` / whatever) instead of writing custom util functions like this?

academo: I'm not a fan of introducing full dependencies for single simple functions. Let's keep it like this for now.

academo (approved):

PR #72, opened 2022-10-18
Move to monorepo to introduce sign-plugin

jackw:
This PR is a starting point to move plugin signing out of toolkit and into its own package. Currently we have a workflow in create-plugin that includes signing a plugin. This has a dependency on toolkit which I'm keen to remove before we launch create-plugin@v1.0.0.

~Consensus is needed on the following points for us to move forward with this PR:~ (Based on discussions and comments in the PR consensus seems to be this is a good direction.)

- [x] Make this repo a monorepo
- [x] Introduce sign-plugin package
- [x] Update workflows to cater for multiple packages
- [x] Update NPM bump and release workflows so create-plugin can still be released
- [x] Merge all open pull requests prior to merging this
- [ ] Define what to do with version parameter of [ManifestInfo](https://github.com/grafana/create-plugin/pull/72/files#diff-e4729107c017626b3b1d757d6e150c786eb52ee6cb9b26a6d7593a64378a9071R18-R20) and adjust the code accordingly

~Alternatively we take the contents of sign-plugin in this PR and put it in its own repo.~

**Notes for reviewers**
I've lifted most of the signing code from toolkit into this repo. I haven't made any changes other than removing the dependency on grafana/data and setting the toolkit version to an empty string just to get it working. I'm thinking we can clean all this up iteratively.

**Follow Up Tasks**

- [ ] Rename repo to `grafana/plugin-tools`
- [ ] Change the references in this PR's readme and other files from `create-plugin` to `plugin-tools`
- [ ] Define a release strategy for multiple packages


mckn:
Love it! Go go go 🚀

mckn (commented):
I think this looks good! Should the `example.mp4` be added to git?

academo:
* I like `grafana/plugin-tools` as the repo name.

leventebalogh:
I like the idea as well! 🚀

jackw:
> Should the example.mp4 be added to git?
It was already in the repo but looking at the docs it doesn't appear to be referenced anywhere. I'm thinking we can 🔪 it. WDYT @mckn ?

mckn:
> > Should the example.mp4 be added to git?
> 
> It was already in the repo but looking at the docs it doesn't appear to be referenced anywhere. I'm thinking we can 🔪 it. WDYT @mckn ?

go go go 🔥

packages/sign-plugin/README.md
```diff
@@ -0,0 +1,62 @@
...
+export GRAFANA_API_KEY=<YOUR_API_KEY>
+npx @grafana/sign-plugin
+```
+
+### Sign a private plugin
+
+In your plugin directory, run the following to create a MANIFEST.txt file in the dist directory of your plugin.
+
+```bash
+npx @grafana/sign-plugin --rootUrls https://example.com/grafana
```
leventebalogh, line 43: Not something necessarily for now: maybe it could be good to add some explanation on what the `--rootUrls` flag is exactly and how they should use it? (I don't think it's straightforward for everyone by the first sight.)

packages/sign-plugin/src/utils/pluginValidation.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('pluginValidation', () => {
+  describe('plugin.json', () => {
+    test('missing plugin.json file', () => {
+      expect(() => getPluginJson(`${__dirname}/mocks/missing-plugin.json`)).toThrowError();
+    });
+  });
+
+  describe('validatePluginJson', () => {
+    test('missing plugin.json file', () => {
```
leventebalogh, line 11: Just a nit typo, probably not important at all. 
```suggestion
    test('missing "id" field in the plugin.json file', () => {
```
(I also know that the code is coming from the Grafana repo as it was and that these tests are not covering too much, we probably improve these in an upcoming PR(s) anyway.)

leventebalogh (approved):
Amazing work! 🚀

academo (approved):
Thanks for the setup!

packages/sign-plugin/README.md
jackw: Noted in #74 ! 🙏

PR #75, opened 2022-10-19
Do not use TS_NODE_PROJECT for webpack configuration

academo:
Fixes https://github.com/grafana/create-plugin/issues/71

Instead of using an environment variable it uses an overwrite in the tsconfig for ts-node (which transpiles the webpack config).

the transpiler `ts-node/transpilers/swc-experimental` was removed from the config as it was creating problems with ts-node and it is a minimal optimization in speed for this specific case.

With these changes the generated scripts run natively in windows. 

image

### DIFF
--- a/packages/create-plugin/templates/common/.config/README.md
+++ b/packages/create-plugin/templates/common/.config/README.md
@@ -109,12 +109,12 @@ We need to update the `scripts` in the `package.json` to use the extended Webpac
 
 **Update for `build`:**
 ```diff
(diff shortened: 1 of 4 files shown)


jackw (approved):
Awesome work on this @academo ! 🚀

wz2b:
I tested this using `npx @grafana/create-plugin@0.2.1 update` and it worked perfectly.  I didn't have to touch any of my own sources, and the plugin builds the same on windows and linux now (tested both).  Thanks for the effort on this, and especially for fixing it so quickly - I really appreciate not having to wedge in temporary hacks while waiting for this to be fixed.  I'm astonished and happy as a clam at how quickly you were able to address this.

PR #78, opened 2022-10-19
Fix npm bump running test

academo:
It also adds `build` as a step to make sure the package is good for publishing


jackw (approved):
LGTM! 🚀

Issue #54, opened 2022-09-28
Don't clear backend binaries in dist folder when running webpack


mckn:
When running webpack we should not remove the built backend binaries from the dist folder because that will require the backend to be rebuilt everytime

sd2k:
FYI this is still an issue for plugins that declare a different `executable` name without the `gpx_` prefix.

academo:
@sd2k could you please open a new issue for this case?

PR #85, opened 2022-10-20
Prevent webpack to clean binaries from go

academo:
Fixes https://github.com/grafana/plugin-tools/issues/54

Adds a condition for webpack to not clean built binaries


mckn (approved):
Looks good to me!

PR #88, opened 2022-10-20
Update backend versions

academo:


marefr:
Pushed an update of running `go mod tidy`. Temporary had to change to `module github.com/grafana/plugin-tools/packages/create-plugin/templates/backend` and then run `go mod tidy` and revert the `module github.com/grafana/plugin-tools/packages/create-plugin/templates/backend`. A bit crazy perhaps, but maybe we can figure out a process/document how you would do this?

marefr (approved):
LGTM

Issue #84, opened 2022-10-20
Do not prompt to generate nor generate backend for panel plugins.


academo:
We don't support backend for panel plugins and the current generated one doesn't work because the plugin json is incomplete.

PR #89, opened 2022-10-20
Do not generate or ask to generate a backend in panel plugins

academo:
Fixes https://github.com/grafana/plugin-tools/issues/84

* It stops asking the user if they want to generate a backend for panel plugin. This was generating some backend code but not the correct plugin.json. Nevertheless we don't support panel plugins with backend.
* It changes the internal tools generate-(panel|datasource|app) to use node-plop directly instead of using the plopjs way to bypass cli arguments. This due to plopjs not being able to handle conditional inputs from the CLI args.


packages/create-plugin/scripts/generate-app.js
```diff
@@ -0,0 +1,24 @@
+const nodePlop = require('node-plop');
```
academo, line 1: Note that I decided to copy-paste the scripts rather than trying to abstract common parts into a separate function to keep them simple and clear.

One could think of a single-script that handle CLI arguments but that will require a far more complex logic that we will later have to maintain and understand so I decided against.

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -69,6 +69,7 @@ export default function (plop: NodePlopAPI) {
         type: 'confirm',
         message: 'Do you want a backend part of your plugin?',
         default: false,
+        when: (answers: CliArgs) => answers.pluginType !== PLUGIN_TYPES.panel,
```
academo, line 72: This is the important change in the generator part.

jackw (approved):
LGTM!

Issue #90, opened 2022-10-20
Do not add `backend: true` or `executable` fields when choosing to not generate a backend.


academo:
If you choose to not generate a backend for app and datasource plugins, create-plugin stills generates a plugin.json with the required fields and this will cause the plugin to fail loading in grafana.

PR #91, opened 2022-10-20
Generate plugin.json backend fields based on user responses

academo:
Fixes https://github.com/grafana/plugin-tools/issues/90


packages/create-plugin/fixtures/test-template/src/plugin.json
```diff
@@ -0,0 +1,35 @@
+{
```
academo, line 1: this fixture is to replace an existing failing test due to changes in the plugin.json

marefr (approved):
LGTM

Issue #80, opened 2022-10-19
[create-plugin] updates prompts to confirm to delete files when the list is empty


academo:
We should not prompt to confirm to delete 

![image](https://user-images.githubusercontent.com/227916/196715485-9af98227-3790-4301-a52a-e0b0282ead4a.png)

PR #94, opened 2022-10-21
Change update prompt message to clarify what is updating

academo:
Closes https://github.com/grafana/plugin-tools/issues/80

Create-plugin overwrites all the `.config` folder when `update` runs. This PR changes the prompt to confirm the update that was previously suggestion a list of files was to be displayed


leventebalogh (approved):
🚀

PR #95, opened 2022-10-21
Fix: Making plugin id schema compliant

### LINKED ISSUE #92: Generated plugin.json does not contain `keywords` field
`keywords` is a required field based on our plugin.json schema.

tolzhabayev:
Fixing plugin id's to make them schema compliant

### DIFF
--- a/packages/create-plugin/scripts/generate-panel.js
+++ b/packages/create-plugin/scripts/generate-panel.js
@@ -11,8 +11,8 @@ async function generatePanel() {
   const plop = await nodePlop(plopFile);
   const generator = plop.getGenerator('create-plugin');
   await generator.runActions({
-    pluginName: 'myplugin',
-    orgName: 'myorg',
+    pluginName: 'my-plugin',
+    orgName: 'my-org',
     pluginDescription: 'Auto-generated panel',
     pluginType: 'panel',
     hasGithubWorkflows: true,
--- a/packages/create-plugin/scripts/generate-panel.js
+++ b/packages/create-plugin/scripts/generate-panel.js
@@ -11,8 +11,8 @@ async function generatePanel() {
   const plop = await nodePlop(plopFile);
   const generator = plop.getGenerator('create-plugin');
   await generator.runActions({
-    pluginName: 'my-plugin',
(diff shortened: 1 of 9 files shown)


tolzhabayev:
duplicate of https://github.com/grafana/plugin-tools/pull/93

academo (approved):

packages/create-plugin/scripts/generate-panel.js
```diff
@@ -11,8 +11,8 @@ async function generatePanel() {
   const plop = await nodePlop(plopFile);
   const generator = plop.getGenerator('create-plugin');
   await generator.runActions({
-    pluginName: 'my-plugin',
-    orgName: 'my-org',
+    pluginName: 'myplugin',
```
jackw, line 14: Rather than change these shouldn't we do something with `normalize_id` so whatever the input is the output is schema compliant?

tolzhabayev: good idea, will update my PR

academo: These inputs are equivalent to user input.

packages/create-plugin/src/utils/utils.handlebars.ts
```diff
@@ -12,8 +12,12 @@ export const ifEq = (a: any, b: any, options: HelperOptions) => {
 
 export const normalizeId = (pluginName: string, orgName: string, type: PLUGIN_TYPES) => {
   const re = new RegExp(`-?${type}$`, 'i');
-  const newPluginName = pluginName.replace(re, '');
-  return changeCase.paramCase(orgName) + '-' + changeCase.paramCase(newPluginName) + `-${type}`;
+  const pluginNameRegex = new RegExp('[^0-9a-z]', 'g');
```
academo, line 15: These two regex are the same you can reuse one. what about `nameRegex``

tolzhabayev: Right, https://github.com/grafana/plugin-tools/pull/95/commits/8cb1fc172605d6dedc615a4c0dfc9a22c5899cb8

PR #98, opened 2022-10-21
App: disable backend by default in yarn generate-app

tolzhabayev:
Adjusts the script that is run when calling `yarn generate-app` to **not** include the backend by default. I think this case needs a bit more love.


academo (approved):
Keep in mind these scripts are not suppose to be user-scripts but scripts to help developing create-plugin itself.

PR #108, opened 2022-10-24
Jest: Ignore ol library

andresmgot:

The package @grafana/data 9.2.1, includes the "open layers" `ol` library. This fails to be imported by jest:

```
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
...
    Details:

    <plugin_path>/node_modules/ol/format/WKT.js:19
    import Feature from '../Feature.js';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (node_modules/@grafana/ui/dist/index.js:66:11)
```

To avoid this, Grafana main repo ignores the package: 

https://github.com/grafana/grafana/blob/5361efc225e7d4b06322b0b9b2005102aa786a26/jest.config.js#L12-L14

So I am copying the same approach here.

Fixes #

### DIFF
--- a/packages/create-plugin/templates/common/.config/jest.config.js
+++ b/packages/create-plugin/templates/common/.config/jest.config.js
@@ -32,4 +32,7 @@ module.exports = {
       },
     ],
   },
+  transformIgnorePatterns: [
+    'node_modules/(?!(ol)/)', // <- exclude the open layers library
+  ],
 };


jackw (approved):
Nice work @andresmgot ! 🚀 

I think as more packages transition away from supplying a cjs build we'll probably need to re-evaluate doing this with a regex.

Issue #62, opened 2022-09-30
create-plugin migrate appears to be using plugin.name instead of id for docker-compose.yml


briangann:
When running the migration, I ended up with a new docker-compose.yml file, and it used the plugin "name" instead of the "id" in a couple of locations.

It also detects that I had a docker-compose.yaml file and wanted to delete it.

Maybe detect if the repo is using .yaml and skip the migration for it? or rename/merge in changes?

PR #109, opened 2022-10-24
Use normalize_id for docker-compose.yml instead of only pluginName

academo:
Fixes: https://github.com/grafana/plugin-tools/issues/62

Using the pluginId instead of only the pluginName

Example of a newly generated docker-compose file:

```yml
version: '3.0'

services:
  grafana:
    container_name: 'myorg-myplugin-panel'
    build:
      context: ./.config
      args:
        grafana_version: ${GRAFANA_VERSION:-9.1.2}
    ports:
      - 3000:3000/tcp
    volumes:
      - ./dist:/var/lib/grafana/plugins/myorg-myplugin-panel
      - ./provisioning:/etc/grafana/provisioning
```


jackw (approved):
Awesome stuff! 🚀

Issue #111, opened 2022-10-24
Jest tests are failing due to some node_modules not being transpiled before loading


academo:
Similar to https://github.com/grafana/plugin-tools/pull/108 other dependencies like react-colorful or esm-browser are causing problems because they need to be transpiled before jest can loaded.

## Context

When jest load tests it also needs to load the code dependencies from node_modules. Those dependencies sometimes need transformation before jest can load them otherwise nodejs can't run them. e.g.: because they are written in typescript or they are [ESM](https://nodejs.org/api/esm.html) that some nodejs version don't understand.

Jest has a [configuration list](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring) of regex that tells it which packages should not transform, by default jest won't transform any module inside node_modules.

This [previous PR](https://github.com/grafana/plugin-tools/pull/108) created an exception for the package `ol`: [open layers](https://www.npmjs.com/package/ol) but there are much more packages that will have this problem and fail to run inside jest.

## Suggested solutions:
* Create and maintain a list of packages that should be transformed by jest and keep the list up to date
* Ask jest to transform all packages regardless if they need transformation or not
* Let the users maintain the list of packages on their own

grafanabot:
:rocket: Issue was released in [`@grafana/create-plugin@0.5.3`](https://github.com/grafana/plugin-tools/releases/tag/%40grafana/create-plugin%400.5.3) :rocket:

PR #112, opened 2022-10-24
Instruct jest to transform all packages (including all inside node_modules)

academo:
Closes https://github.com/grafana/plugin-tools/issues/111

This is an initial solution to the problem described in https://github.com/grafana/plugin-tools/issues/111

It is not feaseable for now to create an maintain a list of packages that need transformation and we'll investigate later in the future a better performant solution.

For now, even though slower, this guarantees tests to run inside jest with ESM and commonjs code.

### DIFF
--- a/packages/create-plugin/templates/common/.config/jest.config.js
+++ b/packages/create-plugin/templates/common/.config/jest.config.js
@@ -32,7 +32,5 @@ module.exports = {
       },
     ],
   },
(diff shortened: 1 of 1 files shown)


jackw (approved):
Awesome work on this @academo ! 🚀

andresmgot (approved):
:+1:

PR #119, opened 2022-10-26
Create-Plugin: Ignore grafana dependencies when migrating

### LINKED ISSUE #49: Migrate: Don't change grafana dependency versions
When migrating a plugin we shouldn't tamper with the versions of `@grafana` dependencies that already exist in a plugins package.json. Doing so will force plugin developers to have to fix all sorts of things (type errors, breaking changes in APIs, changes to e2e-selectors) for the tasks to work again. It will likely cause their plugin to stop working in the minimum version of Grafana that it was previously working in which I doubt a plugin developer wants or expects from the migration.

jackw:

This PR prevents create-plugin from updating any `@grafana/` dependencies that are already listed in a plugins `package.json` when running `npx @create-plugin migrate`. Doing so should ease the pain for a plugin developer migrating without needing to edit source code to deal with typescript errors in build or test.

This feels a little tricky to implement as we rely heavily on the templates for location of dependencies and delete `@grafana/**` dependencies that are listed as `devDependencies` when they should be `dependencies` etc. This led me to opt for to storing references to the "original" package versions then, when calculating the NPM updates, use these references instead to keep the moving / deleting logic without changing the package versions.

| Before | After |
| --- | --- |
| Screenshot 2022-10-26 at 16 15 07 | Screenshot 2022-10-26 at 16 15 20 |

Fixes #49

Bear in mind create-plugin will _still_ introduce `@grafana` dependencies with the "latest" version if they are not present in the plugins `package.json`. An example of this is clock-panel that has no e2e setup however during migration `e2e` and `e2e-selectors` are added along with a docker compose that reference 9.1.2. I don't really think we should get too smart with this and I'm leaning towards us adding a "next steps" document for post migration that covers these sorts of things to aid developers to make the most out of the setup create-plugin migration will provide.

### DIFF
--- a/packages/create-plugin/src/utils/utils.npm.ts
+++ b/packages/create-plugin/src/utils/utils.npm.ts
@@ -3,12 +3,13 @@ import fs from 'fs';
 import semver from 'semver';
 import { readJsonFile } from './utils.files';
 import { renderTemplateFromFile, getTemplateData } from './utils.templates';
-import { TEMPLATE_PATHS } from '../constants';
+import { GRAFANA_FE_PACKAGES, TEMPLATE_PATHS } from '../constants';
 
 type UpdateSummary = Record<string, { prev: string | null; next: string | null }>;
 
 type UpdateOptions = {
   onlyOutdated?: Boolean;
+  ignoreGrafanaDependencies?: boolean;
 };
 
 type PackageJson = {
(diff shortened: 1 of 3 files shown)


academo (approved):
LGTM!

packages/create-plugin/src/utils/utils.npm.ts
```diff
@@ -100,12 +102,44 @@ export function getPackageJsonUpdates(options: UpdateOptions = {}) {
...
+  }
+
   return {
     dependencyUpdates,
     devDependencyUpdates,
   };
 }
 
+function ignoreGrafanaDependencies(dependencyUpdates: UpdateSummary, prevGrafanaDependencies: Record<string, string>) {
+  const whatever = Object.keys(prevGrafanaDependencies);
```
leventebalogh, line 127: Whatever! 😬

leventebalogh (approved):
LGTM 👍

PR #114, opened 2022-10-25
Chore: Update ci scripts to follow new outputs format

tolzhabayev:
**What this PR does / why we need it**:
This PR bumps github actions dependencies and switches to the new outputs format as set-output is deprecated https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #121, opened 2022-10-27
CI: Introduce commands workflow to run on Linux, Mac and Windows

jackw:
Introduces a new "commands" workflow to validate the commands in the bundled packages work on all flavours of OS. This should help us catch issues with cross operating system compatibility without the need for firing up virtual machines locally.

It also bumps the versions of the actions used as per #114

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,16 +12,16 @@ jobs:
     name: Run unit tests
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           ref: ${{ github.ref }}
-      - name: Setup .npmrc file for NPM registry
-        uses: actions/setup-node@v2
+      - name: Setup node
+        uses: actions/setup-node@v3
         with:
           node-version: '16'
-          registry-url: 'https://registry.npmjs.org'
+          cache: 'yarn'
       - name: Install dependencies
-        run: yarn
+        run: yarn install --frozen-lockfile
       - name: Check types
         run: yarn typecheck
       - name: Lint
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,18 +28,3 @@ jobs:
         run: yarn lint
       - name: Unit tests
         run: yarn test:ci
-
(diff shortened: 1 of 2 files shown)


mckn (approved):
Great stuff!! Love it <3

.github/workflows/test.yml
```diff
@@ -12,34 +12,77 @@ jobs:
...
+          path: |
+            packages/create-plugin/create-plugin.tgz
+            packages/sign-plugin/sign-plugin.tgz
+  commands:
+    needs: test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
leventebalogh, line 48: Should we extend it with `macos-latest` at least? (I am not sure `windows-latest` would work out of the box as well, but maybe it should?)

.github/workflows/test.yml
```diff
@@ -12,34 +12,77 @@ jobs:
...
-          GRAFANA_API_KEY: ${{ secrets.GRAFANA_API_KEY }}
-        run: node ../../sign-plugin/dist/bin/run.js --rootUrls http://www.example.com --signatureType private
-        working-directory: ./packages/create-plugin/generated
+      # - name: '@grafana/sign-plugin - build'
+      #   run: yarn workspace @grafana/sign-plugin run build
+      # - name: '@grafana/sign-plugin - sign generated plugin'
+      #   env:
+      #     GRAFANA_API_KEY: ${{ secrets.GRAFANA_API_KEY }}
+      #   run: node ../../sign-plugin/dist/bin/run.js --rootUrls http://www.example.com --signatureType private
+      #   working-directory: ./packages/create-plugin/generated
```
leventebalogh, line 88: Should we get rid of this commented out code?

leventebalogh (approved):
Nice! 👏 

I think it would make sense to merge this as it is (even with just a single OS in the matrix if the others wouldn't work out of the box), and then we have a base to extend with future PRs?

jackw:
I'm closing this for now. Until we find a way around bypassing conditional prompts I can't see a way forward for this.

PR #124, opened 2022-10-27
Removing ts-ignore

tolzhabayev:
Cleanup our app template https://github.com/grafana/grafana-plugin-examples/pull/99#discussion_r1006855026


academo (approved):
Thanks for the follow up

Issue #116, opened 2022-10-26
.gitignore is not present in the generated folder


xnyo:
### Package Name

create-plugin

### What happened?

When running `yarn create @grafana/plugin`, the generated folder does not have a `.gitignore` file, even if it should as it's present in [templates/common](https://github.com/grafana/plugin-tools/tree/main/packages/create-plugin/templates/common).

This happens with every template (app, datasource and panel).

### What you expected to happen

`.gitignore` should be present in the generated folder.

### How to reproduce it (as minimally and precisely as possible)

```sh
╰─❯ yarn create @grafana/plugin
yarn create v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Installed "@grafana/create-plugin@0.3.2" with binaries:
      - create-plugin
[#######################################################################################################################] 308/308?
? What is going to be the name of your plugin? test
? What is the organization name of your plugin? test
? How would you describe your plugin? 
? What kind of plugin would you like?  app
? Do you want a backend part of your plugin? No
? Do you want to add Github CI and Release workflows? No
? Do you want to add a Github workflow for automatically checking "Grafana API compatibility" on PRs? No
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/.eslintrc
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/.prettierrc.js
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/Dockerfile
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/jest-setup.js
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/jest.config.js
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/README.md
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/tsconfig.json
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/types/custom.d.ts
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/webpack/constants.ts
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/webpack/utils.ts
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.config/webpack/webpack.config.ts
✔  ++ /home/giuseppe/grafana/plugins/test-test-app/.eslintrc
[...]

PR #126, opened 2022-10-28
Manually rename gitignore to prevent yarn and npm pack from removing them

academo:

It makes sure that generated plugins contain a `.gitignore` file.

Npm and yarn will not include `.gitignore` files when packing packages and this means the release version of create-plugin will not generate this file. The suggested fixes of creating a `.npmignore` and `.yarnignore` files to prevent this behaviour don't work.

This PR implements a simple exception to include the gitignore file without the dot prefix and rename it when the plugin is generated.

More details of this behaviour https://github.com/npm/npm/issues/3763#issuecomment-22255799

Fixes # https://github.com/grafana/plugin-tools/issues/116

If you want to test this PR you should first run `yarn build` and then `yarn pack` and inspect the packed file to reproduce the error (note the lack of `.gitignore` in the package).


jackw (approved):
Awesome work! 🚀

PR #134, opened 2022-11-03
Create-Plugin: Implement DataSourceWithBackend for datasource plugins with backend

### LINKED ISSUE #25: Datasource should implement DataSourceWithBackend when backend is opted-in
- The datasource.ts should implement the `DataSourceWithBackend` if the user opted in for `backend` when scaffolding. This is so we get the default implementations of health checks etc.

xnyo:

Changes `datasource.ts` (only when backend is opted-in) with an implementation of `DataSourceWithBackend`.

This ensures the `query` and `testDatasource` methods are called from the backend.

The call to the backend is handled by `DataSourceWithBackend` already and does not require any special overriding.

Fixes #25 


mckn (approved):
Great work on this!!

PR #142, opened 2022-11-14
Fix: include uppercase characters when normalising plugin id

jackw:
Users can type all sorts into the prompts when scaffolding a new plugin. Right now the `normalizeId` function strips uppercase characters which creates incorrect IDs (see path of directory created below):

```
success Installed "@grafana/create-plugin@0.5.0" with binaries:
      - create-plugin
? What is going to be the name of your plugin? My Plugin
? What is the organization name of your plugin? HeyWesty
? How would you describe your plugin?
? What kind of plugin would you like?  panel
? Do you want to add Github CI and Release workflows? Yes
? Do you want to add a Github workflow for automatically checking "Grafana API compatibility" on PRs? Yes
✔  ++ /Users/jackwestbrook/Sandbox/plugin-tools-tests/eyesty-ylugin-panel/.config/.eslintrc
```

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@0.5.1-canary.142.d32fe7f.0
  # or 
  yarn add @grafana/create-plugin@0.5.1-canary.142.d32fe7f.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/utils/tests/utils.handlebars.test.ts
+++ b/packages/create-plugin/src/utils/tests/utils.handlebars.test.ts
@@ -31,5 +31,15 @@ describe('Handlebars helpers', () => {
         expect(actual).toEqual(`myorg-myplugin-${type}`);
       }
     );
+
+    test.each([PLUGIN_TYPES.app, PLUGIN_TYPES.datasource, PLUGIN_TYPES.panel])(
+      'should handle uppercase characters',
+      (type: PLUGIN_TYPES) => {
+        const pluginName = `MyPlugin`;
+        const orgName = 'MyOrg';
+        const actual = normalizeId(pluginName, orgName, type);
+        expect(actual).toEqual(`myorg-myplugin-${type}`);
+      }
+    );
   });
 });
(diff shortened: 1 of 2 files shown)


tolzhabayev:
Should'nt we instead `.toLowerCase()` them?

tolzhabayev:
I think we should in:
- `normalize_id` make everything lowecase as we use it in plugin.json -> id which is the plugin slug, that has to be lowecase, no spaces, etc. https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/panel/src/plugin.json#L5
- Outside of `normalize_id` orgName being any case should be allowed https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/panel/src/plugin.json#L10
- Same for `pluginName` it can be any case

leventebalogh (commented):
LGTM 👍 

Why are we using the `change-case`, `upper-case`, etc packages instead of `string.toLowerCase()`? To cater for some unicode character mappings?

Nit: maybe we could just return with this?
```javascript
return changeCase.lowerCase(`${newOrgName}-${newPluginName}-${type}`);
```

tolzhabayev (changes requested):
see above

tolzhabayev (approved):
forget about my messages above, we already do lowercase in return

leventebalogh (approved):

PR #143, opened 2022-11-14
Bumping workflow version to use node 16 instead of deprecated node 12

tolzhabayev:

Removes this deprecation warning, which you can see [here](https://github.com/grafana/plugin-tools/actions/runs/3460203067)

SCR-20221114-f44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,11 @@ jobs:
     name: Run unit tests
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           ref: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


tolzhabayev:
🤔 what labels are correct in this case?

leventebalogh (approved):
LGTM 👍

leventebalogh:
I guess we should have a `chore` label or something similar? 🤔

jackw:
> 🤔 what labels are correct in this case?

In regard to a release label? We shouldn't need any labels. It'll be ignored as it doesn't affect any of the packages.

> I guess we should have a chore label or something similar? 🤔

Have added one and attached it to this PR!

jackw (approved):
LGTM! 🚀

PR #145, opened 2022-11-14
Create-Plugin: Ignore known ES modules in jest config

### LINKED ISSUE #111: Jest tests are failing due to some node_modules not being transpiled before loading

jackw:
The previous [PR](https://github.com/grafana/plugin-tools/pull/112) that addressed issues with loading es modules by transforming all modules in node_modules introduced an issue where SWC will error if a package contains code the typescript compiler doesn't appreciate.

One example of this is reassignment of a class (something we do in [grafana/data](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/datasource.ts#L646)).

I'm not sure if there's a better way to solve this but I'm now leaning towards maintaining a list of packages as a solution mentioned in #111.

Fixes #111

### DIFF
--- a/packages/create-plugin/templates/common/.config/jest.config.js
+++ b/packages/create-plugin/templates/common/.config/jest.config.js
@@ -5,6 +5,8 @@
  */
 
 const path = require('path');
+// Jest will throw errors if it tries to load an es module with it being transformed first
+const esModules = ['ol', 'react-colorful'].join('|');
 
 module.exports = {
   moduleNameMapper: {
@@ -35,5 +37,5 @@ module.exports = {
       },
     ],
   },
-  transformIgnorePatterns: [],
+  transformIgnorePatterns: [`node_modules/(?!(${esModules})/)`],
 };


packages/create-plugin/templates/common/.config/jest.config.js
```diff
@@ -5,6 +5,8 @@
  */
 
 const path = require('path');
+// Jest will throw errors if it tries to load an es module with it being transformed first
+const esModules = ['ol', 'react-colorful'].join('|');
```
leventebalogh, line 9: Maybe it would make sense to update the comment and the name of the variable to reflect more what problems one can solve with adding package names here and when it should be used? Probably it would also make sense to export it so plugin devs can extend it in their own config more easily?

```suggestion
// Just an idea, maybe we can find a better name
const esModuleNpmPackages = ['ol', 'react-colorful'].join('|');
```

leventebalogh (approved):
Nice! 👏

leventebalogh (approved):
🚀

jackw:
> @jackw This causes some issues, maybe something missing in the migration script?
> 
> 
> 
> ```
> 
> Error: Cannot find module './jest/utils'
> 
> Require stack:
> 
> - /drone/src/.config/jest.config.js
> 
> ```
> 
> 
> 
> https://drone.grafana.net/grafana/clock-panel/35/1/5
> 
> https://github.com/grafana/clock-panel/pull/110

Thanks for the heads up @tolzhabayev I'll get it fixed! 🙏

PR #146, opened 2022-11-15
Create-Plugin: Backend template: Switch from Info to Debug level

### LINKED ISSUE #140: Enhancement: Backend template logging improvements
### Package Name

create-plugin

### What happened?

Logging being made with Info level.

https://github.com/grafana/plugin-tools/blob/a5e2abc924dec0f54d8ee49f095f214f844e2d3a/packages/create-plugin/templates/backend/pkg/plugin/datasource.go#L47
https://github.com/grafana/plugin-tools/blob/a5e2abc924dec0f54d8ee49f095f214f844e2d3a/packages/create-plugin/templates/backend/pkg/plugin/datasource.go#L97

Logging being made with `Debug` level since the frequency of QueryData calls can be high and could create a large amount of logged messages.

N/A

### Environment

```shell
N/A
```

### Additional context

_No response_

xnyo:

- Switched the log level in the backend template from `Info` to `Debug`
- Adds a comment to remind to not log sensitive information at `Info` level

Fixes #140

### DIFF
--- a/packages/create-plugin/templates/backend/pkg/plugin/datasource.go
+++ b/packages/create-plugin/templates/backend/pkg/plugin/datasource.go
@@ -44,7 +44,9 @@ func (d *Datasource) Dispose() {
 // The QueryDataResponse contains a map of RefID to the response for each query, and each response
 // contains Frames ([]*Frame).
 func (d *Datasource) QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
-	log.DefaultLogger.Info("QueryData called", "request", req)
+	// when logging at a non-Debug level, make sure you don't include sensitive information in the message
+	// (like the *backend.QueryDataRequest)
+	log.DefaultLogger.Debug("QueryData called", "request", req)
 
 	// create response struct
 	response := backend.NewQueryDataResponse()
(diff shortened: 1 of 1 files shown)


packages/create-plugin/templates/backend/pkg/plugin/datasource.go
```diff
@@ -44,7 +44,9 @@ func (d *Datasource) Dispose() {
 // The QueryDataResponse contains a map of RefID to the response for each query, and each response
 // contains Frames ([]*Frame).
 func (d *Datasource) QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
-	log.DefaultLogger.Info("QueryData called", "request", req)
+	// when logging at a non-Debug level, make sure you don't include sensitive information in the message
+	// (like the *backend.QueryDataRequest)
+	log.DefaultLogger.Debug("QueryData called", "request", req)
```
wbrowne, line 49: In this case I'd remove the `*backend.QueryDataRequest` param completely just in case people don't think to read the comment, don't understand it or just blindly delete it 🙈

xnyo: Like this?
```suggestion
	log.DefaultLogger.Debug("QueryData called")
```

wbrowne: Yeah I think that should be fine! You could also add something like:
```go
log.DefaultLogger.Debug("QueryData called", "numQueries", len(req.Queries))
```

to demonstrate how to pass key values to the log. Totally up to you 😄

xnyo: Good point! I think demonstrating how to pass key/value to the log is valuable :+1: :+1:

wbrowne (approved):
👏

PR #148, opened 2022-11-17
CI: Silence slack from pinging entire channel

jackw:
Sets the [`atTarget`](https://intuit.github.io/auto/docs/generated/slack#attarget) property for the auto slack plugin to prevent from pinging the entire slack channel where there is a new release.

Fixes #

I've created a PR in the auto repo that updates the `atTarget` so we can [disable pinging](https://github.com/intuit/auto/pull/2272). Whilst we wait to see how that pans out I guess we can continue with this approach for now.

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -9,6 +9,11 @@
     "all-contributors",
     "first-time-contributor",
     "released",
-    "slack"
+    [
+      "slack",
+      {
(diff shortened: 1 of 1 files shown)


leventebalogh (approved):
🚢🇮🇹

PR #152, opened 2022-11-18
CI: Speed up builds

jackw:
CI builds are slow due to multiple yarn installs happening. This PR addresses this by using the cache action to store the node_modules and generated/node_modules | yarn.lock files to speed up ci.

| Before | After |
| --- | --- |
| Screenshot 2022-11-18 at 11 03 39 | Screenshot 2022-11-18 at 11 03 56 |

Also spruces up the readme a bit with badges and grafana logo. 😃 

Fixes #


tolzhabayev (changes requested):
https://stackoverflow.com/a/62244232

Solution 1 says:
you don't need actions/cache@v3 anymore and since you arleady use setup-node with cache: yarn you can just `yarn --prefer-offline` instead.

So basically:
remove https://github.com/grafana/plugin-tools/pull/152/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR21-R30

and change this line https://github.com/grafana/plugin-tools/pull/152/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR33

to

`yarn install --immutable --prefer-offline`

jackw:
Nice! Thanks for the feedback @tolzhabayev I've updated the PR with the suggested changes.

tolzhabayev:
This part can also go I think?

https://github.com/grafana/plugin-tools/blob/934365e6f1ab736a1d0b962db119df0d214c24c5/.github/workflows/ci.yml#L21-L27

jackw:
> This part can also go I think?
> 
> https://github.com/grafana/plugin-tools/blob/934365e6f1ab736a1d0b962db119df0d214c24c5/.github/workflows/ci.yml#L21-L27

Turns out you're right. Have removed it. 👍

tolzhabayev (approved):
Send 🇮🇹

PR #155, opened 2022-11-22
Docs - make sure that angular is not supported

tolzhabayev:
Makes it clear that we will not migrate any angular plugins.


jackw (approved):
LGTM! 🚀

PR #158, opened 2022-11-23
Create Plugin: Fix incorrect templates being used for scaffolding

### LINKED ISSUE #157: Bug: Scaffolding a datasource with backend doesn't use DataSourceWithBackend
### Package Name

create-plugin

### What happened?

When choosing to scaffold a datasource with backend the output doesn't include the correct backend datasource template due to the default frontend datasource template being scaffolded first.

For the backend datasource template to be included in the scaffolded plugin

1. Run `npm init @grafana/plugin`
2. Fill in the prompts choosing `datasource` and `yes` to backend part of plugin
3. See error:
```
✖  ++ File already exists
 -> /Users/jackwestbrook/Sandbox/plugin-tools-tests/org-name-datasource/src/datasource.ts
 ``` 
4. View ./src/datasource.ts in newly scaffolded plugin to see wrong template.

### Environment

```shell
System:
    OS: macOS 13.0.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 1.24 GB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.1/bin/npm
  Browsers:
    Brave Browser: 107.1.45.127
    Chrome: 107.0.5304.110
    Chrome Canary: 86.0.4239.0
    Firefox: 104.0.2
    Firefox Nightly: 81.0a1
    Safari: 16.1
    Safari Technology Preview: 16.4
```

### Additional context

Screenshot 2022-11-23 at 16 39 33

jackw:

Right now when we scaffold the different types of plugins in DEV we use the `force` property on plop actions so we can regenerate the plugin without needing to delete the plugin directory first. However for production cases we don't want to trash a developers plugin if they accidentally run the command.

This behaviour difference has a bug in production which prevents a pluginType or backend template (that should be used as an override) from being written to the same location. E.g If a user scaffolds a backend datasource there is `templates/datasource/src/datasource.ts` and `templates/backend/src/datasource.ts` and they should both be scaffolded to `src/datasource.ts`. Currently the first template found will be written and the second will fail causing the scaffolded plugin to have incorrect source code.

This PR attempts to address this by filtering through the destination paths of the combined "common", "pluginType" and "backend" templates to only keep the last entry for a given destination path.

Fixes #157

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@0.6.3-canary.158.10dc5db.0
  # or 
  yarn add @grafana/create-plugin@0.6.3-canary.158.10dc5db.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/commands/generate.plopfile.ts
+++ b/packages/create-plugin/src/commands/generate.plopfile.ts
@@ -101,7 +101,7 @@ export default function (plop: NodePlopAPI) {
         templateData: { pluginId },
       });
 
-      // Copy over files from the plugin type specific folder, e.g. "tempaltes/app" for "app" plugins ("app" | "panel" | "datasource").
+      // Copy over files from the plugin type specific folder, e.g. "templates/app" for "app" plugins ("app" | "panel" | "datasource").
       const pluginTypeSpecificActions = getActionsForTemplateFolder({
         folderPath: TEMPLATE_PATHS[pluginType],
         exportPath,
@@ -112,6 +112,21 @@ export default function (plop: NodePlopAPI) {
       const backendActions = hasBackend
         ? getActionsForTemplateFolder({ folderPath: backendFolderPath, exportPath })
         : [];
+
+      // Common, pluginType and backend actions should only contain the final "override"
+      // template file in the array otherwise plop will error when trying to replace and
+      // scaffolded plugins will have the wrong files.
+      const filteredActions = [...commonActions, ...pluginTypeSpecificActions, ...backendActions]
+        .reduceRight((acc, file, idx) => {
+          const exists = acc.some((f) => f.path === file.path);
+          if (exists && file.type !== 'modify') {
+            return acc;
+          }
+
+          acc.push(file);
+          return acc;
+        }, [])
+        .reverse();
 
       // Copy over Github workflow files (if selected)
       const ciWorkflowActions = hasGithubWorkflows
@@ -125,14 +140,7 @@ export default function (plop: NodePlopAPI) {
       // Replace conditional bits in the Readme files
       const readmeActions = getActionsForReadme({ exportPath });
 
-      return [
-        ...commonActions,
-        ...pluginTypeSpecificActions,
-        ...backendActions,
-        ...ciWorkflowActions,
-        ...readmeActions,
-        ...isCompatibleWorkflowActions,
-      ];
+      return [...filteredActions, ...ciWorkflowActions, ...readmeActions, ...isCompatibleWorkflowActions];
     },
   });
 }
@@ -196,7 +204,7 @@ function getActionsForTemplateFolder({
     templateFile: f,
     // The target path where the compiled template is saved to
     path: path.join(exportPath, getExportPath(f), getExportFileName(f)),
-    // We would like to override generated files in development mode
+    // Support overriding templates. E.g datasource/src/datasource.ts with backend/src/datasource.ts
     force: IS_DEV,
     // We would still like to scaffold as many files as possible even if one fails
     abortOnFail: false,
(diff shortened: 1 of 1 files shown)


packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -113,6 +113,22 @@ export default function (plop: NodePlopAPI) {
         ? getActionsForTemplateFolder({ folderPath: backendFolderPath, exportPath })
         : [];
 
+      // Common, pluginType and backend actions can contain the same destination filenames.
+      // This filtering removes the duplicates to prevent plop erroring and makes sure the correct
+      // template is scaffolded.
```
leventebalogh, line 118: Maybe add something to the comment like "last actions take precedence" - to explain the `.reduceRight(...).reverse()`?

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -113,6 +113,22 @@ export default function (plop: NodePlopAPI) {
...
+      const pluginActions = [...commonActions, ...pluginTypeSpecificActions, ...backendActions]
+        .reduceRight((acc, file) => {
+          const exists = acc.some((f) => f.path === file.path);
+          // if the file.path already exists and isn't a modification exclude it.
+          if (exists && file.type !== 'modify') {
```
leventebalogh, line 123: What happens if the existing file also has `file.type = 'modify'`? Are we overriding it in that case, and not keeping the precedence of the last action?

leventebalogh: (After talking to Jack I actually understand now, resolving.)

packages/create-plugin/src/commands/generate.plopfile.ts
jackw: Decided to take a different approach and order the array spread items backwards so we `reduce` rather than `reduceRight(...).reverse()`. Updated the comment as suggested to explain precedence.

packages/create-plugin/src/commands/generate.plopfile.ts
jackw: We're looking to only exclude additional entries in the array that have `file.type ==='add'` if there is already an entry with `file.type === 'add'` and the file.paths match. I've changed the logic and the comment to try to make it easier to reason with.

jackw: I'm also wondering if we should extract the whole function and write some tests to assert these things are doing their intended job! 😄

leventebalogh: Hehe, we could do that, too, however maybe in this case it would be better to write some "E2E" tests for the scaffolding scenarios, and have a case for testing this functionality as well? (This of course doesn't mean that we couldn't / shouldn't have tests for the units themselves)

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -113,6 +113,22 @@ export default function (plop: NodePlopAPI) {
...
+        const actionExists = acc.find((f) => f.path === file.path);
+        // return early to prevent multiple add type actions being added to the array
+        if (actionExists && actionExists.type === 'add' && file.type === 'add') {
+          return acc;
+        }
+        acc.push(file);
+        return acc;
+      }, []);
+
+      console.log([...backendActions, ...pluginTypeSpecificActions, ...commonActions].length, pluginActions.length);
```
academo, line 130: extra console log?

jackw: Oops. Tis gone now. Thanks for pointing this out!

PR #173, opened 2022-12-08
Templates: Use getDefaultQuery method to set query default values

sunker:
Rather than merging props.query with defaults in the editor and merging target with defaults in the datasource.query method, we should suggest that developers use the [`getDefaultQuery`](https://github.com/grafana/grafana/blob/75cceec62c3e99ba15b0da4958c28e97eea31ea1/packages/grafana-data/src/types/datasource.ts#L360) method. Using this method, defaults will be applied before the query editor is rendered the first time and before the query runner executes the query the first time.


mckn (approved):
LGTM! Thanks for opening this PR 🙏🏻

PR #175, opened 2022-12-18
Signing: Parameterise distribution directory

tomasbasham:
For build pipelines that do not create a distribution directory named exactly `dist` the current signing process will not work. This commit allows the invocation of the signing command to take a flag specifying the path to the distribution directory.

In our particular use case we have implemented a build pipeline that aligns with the majority of our other Go projects. This creates several, deterministically named, distribution directories - one for each platform being built. As such we don't have a single directory named `dist`. This change allows for the distribution directory to be specified on the command line. This should be covered by the current tests in place since if the directory does not exist then, just like if `dist` does not exist, the tests should fail.


packages/sign-plugin/README.md
```diff
@@ -33,6 +33,13 @@ export GRAFANA_API_KEY=<YOUR_API_KEY>
 npx @grafana/sign-plugin
 ```
 
+If the plugin distribution directory differs from the default `dist`, specify the path to use with the `--disDir` flag.
```
jackw, line 36:
```suggestion
If the plugin distribution directory differs from the default `dist`, specify the path to use with the `--distDir` flag.
```

jackw (commented):
Thanks so much for contributing to Grafana @tomasbasham we really appreciate it. ⭐ 

Code looks good to me however there's a typo in the updated docs. Would you mind addressing that before we merge?

packages/sign-plugin/README.md
tomasbasham: Nice spot. I've pushed a fix for this.

jackw (approved):
Awesome work! 🥇

PR #182, opened 2023-01-09
Create Plugin: Make sure nested plugins dist directory structure is correct

jackw:
**What this PR does / why we need it**:
Nested plugins are producing dist directory structures that prevent Grafana from loading the plugins.

### Before

```
./dist
├── CHANGELOG.md
├── LICENSE
├── README.md
├── datasource
│   ├── img
│   │   └── logo.svg
│   └── plugin.json
├── grafana-create-tool-nested-plugins-error
│   ├── module.js
│   └── module.js.map
├── img
│   └── logo.svg
├── plugin.json
└── src
    ├── module.js
    └── module.js.map
```

### After

```
./dist
├── CHANGELOG.md
├── LICENSE
├── README.md
├── datasource
│   ├── img
│   │   └── logo.svg
│   ├── module.js
│   ├── module.js.map
│   └── plugin.json
├── img
│   └── logo.svg
├── module.js
├── module.js.map
└── plugin.json
```

**Which issue(s) this PR fixes**:

Fixes #180 

**Special notes for your reviewer**:

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/create-plugin@0.8.2-canary.182.1320a78.0
  # or 
  yarn add @grafana/create-plugin@0.8.2-canary.182.1320a78.0
  ```
</details>

PR #183, opened 2023-01-11
CI: Update workflows to use permissions rather than PAT

jackw:
Removes the need for a personal access token in favour of using the github token and permissions.

I cherry picked these commits from #182 which verifies it's working for publishing to NPM but that PR is held up with potential docs changes.

Fixes #


leventebalogh (approved):
LGTM 👍

PR #187, opened 2023-01-24
Sign Plugin: Return non zero exit code if error occurs

### LINKED ISSUE #186: Bug: signing failure doesn't return non zero exit code
### Package Name

sign-plugin

### What happened?

When the signing process fail, it should return non zero exit code for the CI systems to work.

CI returns 0 even with signing failure. Instead they should return non-zero exit codes

Sign with incorrect api key and check `exit $?` for status.

image
image
image
image

### Environment

```shell
"@grafana/*"

  System:
    OS: macOS 13.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 149.67 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - /usr/local/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.11.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 109.0.5414.87
    Firefox: 108.0.2
    Safari: 16.1
```

jackw:

Returns a non zero exit code if an error occurs. Useful for CI.

Fixes #186

<details>
  <br />

  ```bash
  npm install @grafana/sign-plugin@0.0.3-canary.187.5dd9cd6.0
  # or 
  yarn add @grafana/sign-plugin@0.0.3-canary.187.5dd9cd6.0
  ```
</details>

### DIFF
--- a/packages/sign-plugin/src/commands/sign.command.ts
+++ b/packages/sign-plugin/src/commands/sign.command.ts
@@ -36,5 +36,6 @@ export const sign = async (argv: minimist.ParsedArgs) => {
     console.log('Signed successfully');
   } catch (err) {
     console.warn(err);
+    process.exitCode = 1;
   }
 };


jackw:
> Don't we need exit code on this failure?
> 
> https://github.com/grafana/plugin-tools/blob/5dd9cd6107dd31e45606f52a2b071306de935fe3/packages/sign-plugin/src/commands/sign.command.ts#L14

From testing, this failure will result in exit code 1 as it throws outside the try/catch. We could probably remove the try/catch entirely from this file as an alternative solution but I'd rather look at that when we find the time to add tests and rewrite most of the sign-plugin package.

academo (approved):

PR #188, opened 2023-01-24
Create Plugin: Update jest config to support Grafana 9.4 packages

jackw:

grafana/grafana/pull/61609 bumps the d3 library to latest to resolve a security vulnerability. However latest d3 only provides an es module which causes jest to complain. This PR updates the list of es modules to make sure it's compiled by swc when jest is run.

Fixes #


leventebalogh (approved):
🚢🇮🇹

PR #197, opened 2023-02-02
CreatePlugin: Change default docker image to grafana-enterprise

mckn:
We got a question about why the `GF_ENTERPRISE_*` environment variables wasn't picked up properly after migrating to the `create-plugin` configuration files. The reason for this is that we are using the `grafana/grafana` docker image which doesn't contain the enerprise version of Grafana.

This PR will change so that all the docker related configuration that is scaffolded by `create-plugin` will use `grafana/grafana-enterprise` as the default docker image. It also adds a way to configure this without adding any changes to the `./config` folder which might cause issues while updating.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@0.10.0-canary.197.f95c2c8.0
  # or 
  yarn add @grafana/create-plugin@0.10.0-canary.197.f95c2c8.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/.config/README.md
+++ b/packages/create-plugin/templates/common/.config/README.md
@@ -139,3 +139,26 @@ We need to update the `scripts` in the `package.json` to use the extended Webpac
 -"dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
 +"dev": "webpack -w -c ./webpack.config.ts --env development",
 ```
+
+### Configure grafana image to use when running docker
+
+By default `grafana-enterprise` will be used as the docker image for all docker related commands. If you want to override this behaviour simply alter the `docker-compose.yaml` by adding the following build arg `grafana_image`. 
+
+**Example:**
+
+```yaml
+version: '3.7'
+
+services:
+  grafana:
+    container_name: 'myorg-basic-app'
+    build:
+      context: ./.config
+      args:
+        grafana_version: ${GRAFANA_VERSION:-9.1.2}
+        grafana_image: ${GRAFANA_IMAGE:-grafana}
+```
+
+In this example we are assigning the environment variable `GRAFANA_IMAGE` to the build arg `grafana_image` with a default value of `grafana`. This will give you the possibility to set the value while running the docker-compose commands which might be convinent in some scenarios.
+
+---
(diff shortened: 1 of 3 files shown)


packages/create-plugin/templates/common/.config/README.md
```diff
@@ -139,3 +139,26 @@ We need to update the `scripts` in the `package.json` to use the extended Webpac
 -"dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
 +"dev": "webpack -w -c ./webpack.config.ts --env development",
 ```
+
+### Configure grafana image to use when running docker
```
jackw, line 143: Nice! 👏 

Could we also copy this to the [advanced config](https://github.com/grafana/plugin-tools/blob/main/docusaurus/docs/advanced-configuration.md) for the documentation website please?

packages/create-plugin/templates/common/.config/README.md
```diff
@@ -139,3 +139,26 @@ We need to update the `scripts` in the `package.json` to use the extended Webpac
...
+version: '3.7'
+
+services:
+  grafana:
+    container_name: 'myorg-basic-app'
+    build:
+      context: ./.config
+      args:
+        grafana_version: ${GRAFANA_VERSION:-9.1.2}
+        grafana_image: ${GRAFANA_IMAGE:-grafana}
```
jackw, line 159: I'm wondering if we should set the default `grafana_image` in the [docker-compose.yaml](https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/common/docker-compose.yaml#L8-L9) to `grafana` and then update this doc to mention switching to `grafana-enterprise` with the diff for using `grafana-enterprise`?

e.g.

```
args:
  grafana_version: ${GRAFANA_VERSION:-9.1.2}
-  grafana_image: ${GRAFANA_IMAGE:-grafana}
+  grafana_image: ${GRAFANA_IMAGE:-grafana-enterprise}
```
Originally I thought switching to enterprise would be an easy solution to this but now I'm thinking it's more internal to Grafana and might confuse the community?

jackw (approved):
LGTM! 🚀 

Left a couple of nits.

packages/create-plugin/templates/common/.config/README.md
mckn: Will leave it as it is according to our chat, ok?

packages/create-plugin/templates/common/.config/README.md
mckn: Done

jackw:
:shipit:

PR #206, opened 2023-03-02
Fix datasource without backend template and update test CI

academo:
- Update scripts to generate a datasource plugin with and without backend
- Remove unused import from datasource template
- Update ci to include app with backend and datasource without backend
- Use grafana 9.3.8 in dependencies

Fixes #


packages/create-plugin/scripts/generate-datasource-without-backend.js
```diff
@@ -0,0 +1,24 @@
+const nodePlop = require('node-plop');
+const path = require('path');
+const fs = require('fs');
+
+async function generatePanel() {
```
leventebalogh, line 5: **Nit:** probably `generateDatasource()` instead, although `generate()` would be probably better as it's "copy-paste safer" :)

leventebalogh (approved):
Nice 👍

Issue #204, opened 2023-02-24
Backend binaries with renamed executables are removed in build process


sd2k:
Similar to https://github.com/grafana/plugin-tools/issues/54 in which webpack cleans up the `dist` directory during a build, including backend binaries that may have just been put there. 

https://github.com/grafana/plugin-tools/pull/85 fixed this for backend executables with the default name (`gpx_whatever`) but if the plugin specifies a different executable in plugin.json (e.g. `"executable": "ml_backend"`) and the backend binaries have different names, they're silently removed from the `dist` directory during the `yarn build` command.

PR #207, opened 2023-03-02
Keep custom executable name if defined

academo:

Keeps compiled binaries that don't follow the `gpx_.*` prefix in the `./dist` directory when webpack runs.

Fixes https://github.com/grafana/plugin-tools/issues/204

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@1.2.0-canary.207.948a872.0
  npm install @grafana/sign-plugin@1.1.0-canary.207.948a872.0
  # or 
  yarn add @grafana/create-plugin@1.2.0-canary.207.948a872.0
  yarn add @grafana/sign-plugin@1.1.0-canary.207.948a872.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -128,7 +128,6 @@ const config = async (env): Promise<Configuration> => ({
   output: {
     clean: {
       keep: new RegExp(`gpx_.*|${pluginJson.executable || 'gpx_'}.*`),
-
     },
     filename: '[name].js',
     library: {
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -13,8 +13,10 @@ import path from 'path';
 import ReplaceInFileWebpackPlugin from 'replace-in-file-webpack-plugin';
 import { Configuration } from 'webpack';
 
-import { getPackageJson, getPluginId, hasReadme, getEntries } from './utils';
+import { getPackageJson, getPluginJson, hasReadme, getEntries } from './utils';
 import { SOURCE_DIR, DIST_DIR } from './constants';
+
+const pluginJson = getPluginJson();
 
 const config = async (env): Promise<Configuration> => ({
   cache: {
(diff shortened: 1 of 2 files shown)


packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -125,7 +127,7 @@ const config = async (env): Promise<Configuration> => ({
 
   output: {
     clean: {
-      keep: /gpx_.*/,
+      keep: new RegExp(`gpx_.*|${pluginJson.executable || 'gpx_'}.*`),
```
academo, line 130: I'm keeping `gpx_.*` for backwards compatibility and intentionally using a redundant fallback. This might be me being too cautious but there's no harm in a duplicated regex or if this might cover some weird edge cases where you have both gpx_ prefixed binaries and custom binaries.

jackw: This change looks good but I'm not sure it will work with nested plugins. 🤔 

If an App plugin has a `plugin.json` in the root `src` directory but it is the datasource `plugin.json` that declares the executable e.g `../gpx_zabbix-plugin` I think this will continue to fail. Should we worry much? 

`output.clean.keep` can also accept a function. Could we do anything smarter there to understand if the file should be removed or not.

academo: @jackw good point on nested plugins.  what if Instead I just check that it doesn't have any of the target platforms suffix? e.g. amd64, arm, etc?

jackw: Yeah I think that should work. From what I know we append `darwin`, `windows`, and `linux` to whatever is set in the plugin.json `executable` field. 👍

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -125,7 +127,7 @@ const config = async (env): Promise<Configuration> => ({
 
   output: {
     clean: {
-      keep: /gpx_.*/,
+      keep: new RegExp(`.*?_(amd64|arm(64)?)(.exe)?`),
```
academo, line 130: @jackw I implemented the change here.

I kept the `getPluginJson` method because it can be useful for future features.

jackw (approved):
LGTM! 🚀

PR #213, opened 2023-03-06
Bumped mage action to v2 to avoid the warning "Node.js 12 actions are…

romain-gaillard:
… deprecated"

The ci.yml and release.yml workflow files use an outdated version of the mage action. This outdated version uses Node 12 as a runtime, which leads to a warning when the workflows are used. This commit bumps the mage action version to 2, which uses to Node 16 as a runtime.

**What this PR does / why we need it**:  We need it to get rid of the warning shown by GitHub:
"Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: magefile/mage-action@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/."

**Which issue(s) this PR fixes**: None

**Special notes for your reviewer**: Tested successfully.

### DIFF
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -59,14 +59,14 @@ jobs:
 
       - name: Test backend
         if: steps.check-for-backend.outputs.has-backend == 'true'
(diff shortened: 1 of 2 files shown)


academo (approved):
Thanks!

PR #215, opened 2023-03-07
Chore: Update `@grafana/eslint-config`

leventebalogh:
### What changed?
Updated the `@grafana/eslint-config` to the latest version and also cleaned up the uneccessary ESLint related deps in our package.json.

### Why?
Why not? Also the outdated `eslint-plugin-jsdoc` in our deps required <17 versions of node to run the project with.


package.json
```diff
@@ -21,19 +21,13 @@
...
     "@auto-it/first-time-contributor": "^10.37.6",
     "@auto-it/slack": "^10.37.6",
-    "@grafana/eslint-config": "^2.5.0",
+    "@grafana/eslint-config": "^5.1.0",
     "@grafana/tsconfig": "^1.2.0-rc1",
     "@swc/core": "^1.2.162",
     "@swc/jest": "^0.2.20",
     "@types/jest": "^27.4.1",
     "@types/node": "18.11.7",
-    "@typescript-eslint/eslint-plugin": "^4.33.0",
```
academo, line 30: Are you sure we don't need these dependencies anymore? I can see them as part of the plugins common package.json template.

jackw: I'm pretty sure they aren't needed as the dependencies in the package.json template are for after a plugin is scaffolded where as this worktree `package.json` has the common devDependencies for the workspaces `create-plugin` and `sign-plugin`. 

This fixes the node 18 install issue and cleans up as `@grafana/eslint-config` has most of these as nested deps too.

package.json
```diff
@@ -21,19 +21,13 @@
...
-    "@typescript-eslint/parser": "^4.33.0",
     "auto": "^10.37.6",
-    "eslint": "^7.32.0",
-    "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-jsdoc": "^36.1.0",
-    "eslint-plugin-prettier": "^4.0.0",
```
jackw, line 36: Was this ever used? Asking as the latest eslint-config doesn't use it or include it.

jackw (approved):
Nice! 🚀 

Left a comment about the eslint-prettier-plugin.

package.json
leventebalogh: That's a very good point, and I am not sure when it was used - maybe in the beginning. But it looks to me like we don't need it anymore 🤔

package.json
leventebalogh: Yea, as Jack says those are the dependencies of the scaffolded plugins, while these are the dependencies of the tools themselves. I think we could clean up the deps of the scaffolded plugins as well, but that's a separate step.

Issue #163, opened 2022-11-29
Bug: Create Plugin update command updates Grafana dependencies


jackw:
### Package Name

create-plugin

### What happened?

When running `npx @grafana/create-plugin update` if the user asks to update dependencies it will include the `grafana` npm packages. This could introduce painful upgrade paths for devs related to type changes or api changes when they only want to keep their configs up to date.

### What you expected to happen

To be offered the choice to update grafana dependencies or to leave them as they are.

### How to reproduce it (as minimally and precisely as possible)

N/A

### Environment

```shell
N/A
```

PR #217, opened 2023-03-07
Create Plugin: Don't upgrade grafana dependencies when running update command

academo:

The update command is prompting to update npm dependencies but it is always suggesting the grafana packages to be updated. This introduces an option to opt-out of updating these packages.

Closes https://github.com/grafana/plugin-tools/issues/163

This PR tries to use the existing logic to handle dependencies and intentionally does not fix underlying issues. I'll create a second pr with  a refactor of all the code after this is merged.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@1.2.1-canary.217.2441a25.0
  # or 
  yarn add @grafana/create-plugin@1.2.1-canary.217.2441a25.0
  ```
</details>


packages/create-plugin/src/constants.ts
```diff
@@ -95,6 +95,7 @@ export const TEXT = {
   updateNpmDependenciesPrompt: '**Would you like to update the following dependencies in the `package.json?`**',
   updateNpmDependenciesSuccess: 'Successfully updated the NPM dependencies.',
   updateNpmDependenciesAborted: 'No NPM dependencies have been updated.',
+  updateGrafanaNpmDependenciesPrompt: '**Would you like to update the Grafana packages (@grafana/*) in the `package.json?`**',
```
sympatheticmoose, line 98: Given @jackw 's comments in the issue:
> This could introduce painful upgrade paths for devs related to type changes or api changes when they only want to keep their configs up to date

It feels like we'd want the default behaviour to be 'no'? Or at least to advise on the implications of updating these?

leventebalogh:
I would probably lean towards just skipping this extra prompt and don't update the `@grafana` dependencies (similar to what have in the migration command right now). 

I have a feeling plugin authors don't always want to update to use the latest version of all the `@grafana` packages (depending on how much migration work they wanna do in a single step). It would also make the code simpler for now (less nesting). It would solve the current issue, and we could refactor this part in the future if we have a better idea about how we imagine the whole update being used.

academo:
@jackw @leventebalogh @sympatheticmoose I updated the PR to always skip grafanaDependencies when updating the configuration.

packages/create-plugin/src/constants.ts
academo: I pushed an update to always skip the dependencies. I agree that's a better default behaviour.

jackw (approved):
Awesome work! 👌

sympatheticmoose:
Agree with the approach 👍 

I could be wrong, but it feels like theres still a hanging question on how to help a user update the grafana dependencies on those less frequent occasions. Maybe worth a follow up issue for visibility/feedback?

sympatheticmoose (approved):

mckn (approved):
LGTM

mountainash:
RE: https://github.com/grafana/plugin-tools/pull/217#issuecomment-1488548075

This is valid! We are using `npx @grafana/create-plugin@latest update` regularly to ensure we our plugin is up-to-date with Grafana suggestions, yet it's core dependencies are now outdated (in our case `v9.3.8`).

I'd tried changing the version in the `docker-compose.yml` then running the CLI with `update` arg. Changing the `dependencies.grafanaDependency` in `plugin.json` and running the CLI with `update` arg. Also tried `GRAFANA_VERSION=10.2.2 npx @grafana/create-plugin@latest update`. There's nothing in the docs at https://grafana.com/developers/plugin-tools/migration-guides/update-create-plugin-versions about how to do it. Until I found this... quite strange that this decision was made, even when most of the work was done for an optional choice.

What would be the suggested way to do this for a plugin developer? Manually updating the `package.json` version numbers?

mckn:
> RE: [#217 (comment)](https://github.com/grafana/plugin-tools/pull/217#issuecomment-1488548075)
> 
> This is valid! We are using `npx @grafana/create-plugin@latest update` regularly to ensure we our plugin is up-to-date with Grafana suggestions, yet it's core dependencies are now outdated (in our case `v9.3.8`).
> 
> I'd tried changing the version in the `docker-compose.yml` then running the CLI with `update` arg. Changing the `dependencies.grafanaDependency` in `plugin.json` and running the CLI with `update` arg. Also tried `GRAFANA_VERSION=10.2.2 npx @grafana/create-plugin@latest update`. There's nothing in the docs at https://grafana.com/developers/plugin-tools/migration-guides/update-create-plugin-versions about how to do it. Until I found this... quite strange that this decision was made, even when most of the work was done for an optional choice.
> 
> What would be the suggested way to do this for a plugin developer? Manually updating the `package.json` version numbers?

I think the recommended workflow would be something like:

1. You scaffold you plugin using the `@grafana/create-plugin` tool with you favourite package manager of choice.
2. If you want to update the Grafana dependencies you do that with your package manager (npm, yarn etc...).

We could probably provide a command for this but on the other hand you should manage your Grafana dependencies as any other dependency which will be managed with your package manager.

mountainash:
> We could probably provide a command for this but on the other hand you should manage your Grafana dependencies as any other dependency which will be managed with your package manager.

Thanks for the feedback @mckn - that's a good way of thinking about the current flow for a new plugin.

But then I don't really get the use case of the "update" function (eg. `npx @grafana/create-plugin@latest update` - seems there'd be quite some overlap.

We were using the "update" function to keep our plugin in-sync with the "official" Grafana supported dependencies and versions. If use another tool - we'd likely be using newer "un-verified" dependency versions.

mckn:
> > We could probably provide a command for this but on the other hand you should manage your Grafana dependencies as any other dependency which will be managed with your package manager.
> 
> Thanks for the feedback @mckn - that's a good way of thinking about the current flow for a new plugin.
> 
> But then I don't really get the use case of the "update" function (eg. `npx @grafana/create-plugin@latest update` - seems there'd be quite some overlap.
> 
> We were using the "update" function to keep our plugin in-sync with the "official" Grafana supported dependencies and versions. If use another tool - we'd likely be using newer "un-verified" dependency versions.

Great feedback! Thanks 🙏🏻 

The update function should be responsible for keeping the scaffolded `.config` folder up to date. So it will update build scripts, base configurations etc. We should probably explain this a bit more in the documentation.

Regarding the `@grafana`-packages. Grafana try to follow semver when shipping new versions of the Grafana packages. So it should be safe to update a patch versions. A minor version should also be safe to update. With that said, Grafana is a big project with lots of contributors so from time to time there have been breaking changes within a minor version but we are trying our best to prevent that from happening. Upgrading to a major version would require the plugin author to do some proper testing to verify that no breaking changes have been introduced for the APIs you plugin use.

My point is that it is only you as a plugin developer that can verify the compatibility between your plugin and a Grafana version because it depends on which APIs your plugin are using and how they are being used.

I hope this makes sense.

mountainash:
All great advice, but the follow statement doesn't match the current functionality of the `update` command.

> The update function should be responsible for keeping the scaffolded .config folder up to date. So it will update build scripts, base configurations etc. We should probably explain this a bit more in the documentation.

It **does** modifications to `package.json`. It updates versions and even (optionally) downgrades versions (if you push ahead of what it wants). 

... and this is the crux of this issue.

mckn:
> All great advice, but the follow statement doesn't match the current functionality of the `update` command.
> 
> > The update function should be responsible for keeping the scaffolded .config folder up to date. So it will update build scripts, base configurations etc. We should probably explain this a bit more in the documentation.
> 
> It **does** modifications to `package.json`. It updates versions and even (optionally) downgrades versions (if you push ahead of what it wants).
> 
> ... and this is the crux of this issue.

hmm, that is true but only for dev dependencies. Right? Again - this is great feedback that we should reflect on. The dev dependencies is related to the .config folder.

PR #219, opened 2023-03-09
Fix: GH bug report template

jackw:
The bug report template [on main](https://github.com/grafana/plugin-tools/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml) appears to be broken. The changes in [this branch](https://github.com/grafana/plugin-tools/blob/jackw/fix-issue-template/.github/ISSUE_TEMPLATE/bug_report.yml) appear to fix it.


leventebalogh (approved):
Nice catch 🙌

PR #226, opened 2023-03-22
Create Plugin: Support different Node package managers

### LINKED ISSUE #156: Provide npm option
Background

- yarn v1 is very slow and [has not been updated for over 1.5 years](https://github.com/yarnpkg/yarn/releases).
- And yarn v2 is totally different and lack of ecosystem.
- npm has a lot of improvements in last a few years, so yarn does not have clear advantages than before.

For example, simply switching from yarn v1 to npm v7, speeds up a lot based in my [past experiment](https://github.com/Hongbo-Miao/hongbomiao.com/pull/1935).

(Left side is yarn v1, right side is npm v7)

image

Today Node.js 18 comes with npm v8 which is even better. (Latest npm version is v9)

It would be great to provide generating plugin in npm option, or totally move to npm (?) To developers, one less dependency is also a good thing especially yarn v1 is very old and npm is faster today. 😃

This commit will help a little bit https://github.com/Hongbo-Miao/hongbomiao.com/pull/5893/commits/41da617088d3bfde72099e77eed054ae534a9e5e for npm option.

BTW, for this repo itself, it may speed up a lot when switching to npm too. 👻

jackw:

This PR aims to give plugin devs the option to use a node package manager other than Yarn 1. The associated issue highlights that Yarn 1 is indeed deprecated and will reach end of life at some point (it's not been updated in over 1.5 years). It also raises the question over using a different package manager to speed up tasks.

Rather than force a package manager on developers I've opted to let the user decide what they want to use. 

#### A fresh scaffold
- Rely on npm_config_user_agent to get package manager and version
- Insert into package.json packageManager on successful scaffold

#### An update or migration to a plugin
- check for package.json packageManager prop
- check for lock files
- fallback to yarn 1
- Insert into package.json packageManager if missing in step 1

Ultimately every plugin that uses create-plugin either to scaffold, update or migrate will have the `packageManager` property added to its [package.json](https://nodejs.org/docs/latest-v18.x/api/packages.html#packagemanager). This is experimental but currently used by corepack. Hopefully it will make it easy for our tooling to know which package manager is used by a plugin.

I've opted not to introduce support for yarn berry at this time.

### How much faster is NPM / PNPM over Yarn?

I've created a [test repo here](https://github.com/jackw/grafana-plugins-playground#grafana-plugins-playground) that tries to benchmark some plugin related frontend tasks regarding the different package managers. There are also some [successful workflow runs](https://github.com/jackw/grafana-plugins-playground/actions) and a [PR open](https://github.com/jackw/grafana-plugins-playground/pull/1) to test is-compatible works across all three package managers.

Based on these I'm not sure NPM is actually any faster than Yarn 1. PNPM however is a 🚀 .

~I've also not opted to do anything with updates or migrations right now as I'd like some opinion on:~

- ~Should we offer support for multiple package managers? (Similar to the current state of this PR.)~
- ~Should we only support NPM?~
- ~Should we instead move users to yarn berry with a node-linker setting of `node_modules` to match as closely as possible to current workflows?~
- ~How will this impact plugin validation?~

Based on feedback below I've gone ahead with supporting all three package managers (NPM, PNPM, Yarn 1).

Fixes #156

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@1.3.0-canary.226.755718c.0
  # or 
  yarn add @grafana/create-plugin@1.3.0-canary.226.755718c.0
  ```
</details>


tolzhabayev:
My 2 cents:
- Support npm only due to https://github.com/github/roadmap/issues/657 and it's wide support and adoption because it will keep being the 1st class citizen for npmjs

academo:
Thanks for putting this together @jackw !


>    Should we offer support for multiple package managers? (Similar to the current state of this PR.)
Updates don't handle anything related to the package manager. The migration script also doesn't modify anything related to package manager as far as I remember. Update and migrate are both based on the templates content.

>    Should we only support NPM?
If we merge this PR we'll have to modify the update and migrate scripts to maintain the existing package manager or forcing everyone that runs update/migrate to start using NPM. If we have to take a decision like this my vote is to either keep only supporting yarn or support all 3 yarn, npm and pnpm

>    Should we instead move users to yarn berry with a node-linker setting of node_modules to match as closely as possible to current workflows?
Please no.

>    How will this impact plugin validation?
Plugin validation is only affected by package managers when OSV scanner is involved and osv scanner supports 3 package managers in this PR. Levitate `isCompatible` only supports yarn and npm but we don't use it in the plugin validation.

leventebalogh:
Nice one @jackw! 👏  
Pretty cool that you have put together that test repo with the benchmarks.

**I would vote for supporting all three of them (yarn 1, npm, pnpm)**, and to not support Yarn berry. By seamlesslynallowing folks to use whatever main package manager they like the most, I think we could prevent some future requests / questions. An argument could be that like this we need to give support for 3 different package managers, however I don't think that's necessarily true, we _just_ need to make sure that our starter works with all 3 of them.
 
Sigstore: I believe that we should be able to provide a script to sign the plugin and interact with Sigstore in a way that is package-manager independent.

Update and migrate: as we discussed and as @academo pointed it out as well, we somehow need to detect what package manager the project is using before generating the new templates. But I guess this is not part of this draft PR yet only because you wanted to get some input on the idea first.

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -191,19 +217,30 @@ function getActionsForTemplateFolder({
...
+    }
+
     return path.extname(f) === '.hbs' ? path.basename(f, '.hbs') : path.basename(f);
   }
   function getExportPath(f: string) {
     return path.relative(folderPath, path.dirname(f));
   }
 
+  // Remove the npmrc file if not running `pnpm`.
+  if (templateData.packageManagerName !== 'pnpm') {
```
academo, line 238: Nit: move this logic up to after line 220 for clarity

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -0,0 +1,105 @@
+import { basename } from 'path';
+import { spawnSync } from 'child_process';
+import { sync as findUpSync } from 'find-up';
+import { getPackageJson } from './utils.npm';
+
+const NPM_LOCKFILE = 'package-lock.json';
+const PNPM_LOCKFILE = 'pnpm-lock.yaml';
+const YARN_LOCKFILE = 'yarn.lock';
+const SUPPORTED_PACKAGE_MANAGERS = ['npm', 'yarn', 'pnpm'];
+const LEGACY_PACKAGE_MANAGER = { packageManagerName: 'yarn', packageManagerVersion: '1.22.19' };
```
academo, line 10: Nit: `FALLBACK_PACKAGE_MANAGER`. or `DEFAULT_PACKAGE_MANAGER` we are not really deprecating yarn 1. just making it the default among other options.

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return undefined;
+  } catch (error) {
+    console.error('Failed to find package manager from lock file. Have you installed dependencies?');
+    throw Error(error);
+  }
+}
+
+function getPackageManagerFromPackageJson(): PackageManager | undefined {
```
academo, line 97: my paraonid mind tells me that we can't really trust the package.json to contain the correct package manager and specially since this is also an experimental feature https://nodejs.org/docs/latest-v18.x/api/all.html#all_packages_packagemanager

my suggestion is to rely on lock files and fallback to this.

leventebalogh: Hmm, I am not sure, as we are persisting this field in the package.json during scaffolding / updating / migrating I would say we can rely on it. As we are using the same signature (`"<package manager name>@<version>"`) as Node.js and Corepack, I don't think this should conflict (we only override this property if it was not set before)?

leventebalogh (approved):
Left a few nits, but overall LGTM. Nice one 👏

packages/create-plugin/src/utils/utils.packageManager.ts
jackw: Replaced with `DEFAULT_PACKAGE_MANAGER`. 👍

packages/create-plugin/src/commands/generate.plopfile.ts
jackw: Moved it up as suggested. 👍

packages/create-plugin/src/utils/utils.packageManager.ts
jackw: As @leventebalogh mentions we're setting this ourselves based on either the lock files or the npm_user_agent if it doesn't already exist. Whilst corepack is experimental this package.json property is already widely used in projects (courtesy of yarn berry and pnpm) which gives me the confidence to use it here too. It's also being used by the pnpm github action to know which version of pnpm to install when running the workflows.

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    case 'pnpm':
+      return 'pnpm install --frozen-lockfile --prefer-offline';
+
+    default:
+      return 'npm ci';
+  }
+}
+
+function getPackageManagerFromLockFile(): PackageManager | undefined {
+  const closestLockfilePath = findUpSync([YARN_LOCKFILE, PNPM_LOCKFILE, NPM_LOCKFILE]);
```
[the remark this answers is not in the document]
jackw: I opted for this approach as create-plugin is already being used in some mono-repo setups and I'd like to make sure that wherever this command is run it's gonna look up the file tree for the closest lock file. I'm unsure if `glob` can achieve the same functionality. 🤔

academo (approved):

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -0,0 +1,105 @@
...
+const NPM_LOCKFILE = 'package-lock.json';
+const PNPM_LOCKFILE = 'pnpm-lock.yaml';
+const YARN_LOCKFILE = 'yarn.lock';
+const SUPPORTED_PACKAGE_MANAGERS = ['npm', 'yarn', 'pnpm'];
+const DEFAULT_PACKAGE_MANAGER = { packageManagerName: 'yarn', packageManagerVersion: '1.22.19' };
+
+export type PackageManager = {
+  packageManagerName: string;
+  packageManagerVersion: string;
+};
```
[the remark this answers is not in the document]
jackw: I did this originally but then I got fed up of having to rename these properties to match the template data variables `packageManagerName` and `packageManagerVersion` so decided to do it here instead.

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -119,11 +129,27 @@ export default function (plop: NodePlopAPI) {
...
-        ? getActionsForTemplateFolder({ folderPath: TEMPLATE_PATHS.isCompatibleWorkflow, exportPath })
+        ? getActionsForTemplateFolder({
+            folderPath: TEMPLATE_PATHS.isCompatibleWorkflow,
+            exportPath,
+            templateData: {
+              packageManagerName,
+              packageManagerInstallCmd,
+              packageManagerVersion,
+            },
+          })
```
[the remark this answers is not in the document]
jackw: Tis done! 👍

packages/create-plugin/src/commands/generate.plopfile.ts
```diff
@@ -191,15 +217,26 @@ function getActionsForTemplateFolder({
   exportPath: string;
   templateData?: Record<string, string>;
 }) {
-  const files = glob.sync(`${folderPath}/**`, { dot: true });
+  let files = glob.sync(`${folderPath}/**`, { dot: true });
+
+  // Remove the npmrc file if not running `pnpm`.
```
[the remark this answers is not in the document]
jackw: Have updated comment explaining the npmrc file is for pnpm settings.

packages/create-plugin/src/utils/utils.npm.ts
```diff
@@ -236,6 +238,14 @@ export function updateNpmScripts() {
   writePackageJson(packageJson);
 }
 
+export function writePackageManager({ packageManagerName, packageManagerVersion }: PackageManager) {
```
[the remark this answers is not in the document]
jackw: Tis done! 👍

packages/create-plugin/src/utils/utils.packageManager.ts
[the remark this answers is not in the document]
leventebalogh: Fair enough, let's leave it as it is 👍

PR #232, opened 2023-04-17
Chore: Update dev-dependencies

leventebalogh:
### What changed?
- **updated the `devDependencies`** for the scaffolded plugins (most of these are just for "clarity", as they were already defined using carets `^` which already allowed all possible minor and patch versions)
- **"downgraded" and fixed typescript to version `4.8.4`** - this is temporary, and the reason is to hide the typescript version mismatch because of our grafana-eslint package

### Major updates
- **`webpack-cli` (4->5)**
  - no changes that should affect us

### What's next with Typescript?
The goal is to update typescript to the latest possible version, too. In order to do this first we need to update `@typescript-eslint/eslint-plugin` (so it supports TS >5) [in eslint-config-grafana](https://github.com/grafana/eslint-config-grafana/blob/master/package.json#L24), release a new version of it, and then update it in the `create-plugin`.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@1.3.2-canary.232.f3fd3ea.0
  # or 
  yarn add @grafana/create-plugin@1.3.2-canary.232.f3fd3ea.0
  ```
</details>


jackw (approved):
LGTM 🚀

PR #257, opened 2023-05-25
Scripts: Add error-handling to generator scripts

leventebalogh:
### What changed?

Currently the errors inside the plop generator are "swallowed", we are not seeing them in the console (which caused us some headaches while trying to add the new app-scenes template).

This PR is fixing this by logging the generator errors to the console.


jackw (approved):
LGTM! 🚀 

I removed the release related fields as I don't think these dictate a release as we only use them in CI / dev, right?

leventebalogh:
Aaa, true! You are right, thanks 👍

PR #259, opened 2023-05-26
CI: Use a GH app token

### LINKED ISSUE #222: Github Actions: Move away from grafanabot token and use a Github App
Utilize https://github.com/marketplace/actions/github-app-token instead of a stored grafanabot token.

```[tasklist]
### Setup github app
- [ ] contact it helpdesk to create a new app
- [ ] configure the app
```

```[tasklist]
### Replace token stored in secrets
- [ ] release pipeline
- [ ] integration tests
```

jackw:
This PR replaces the usage of the GH bot token with a GH app token in workflows which I believe is the preferred way to auth these days.

Fixes #222

I'm not sure how to test this fully other than merge and wait for releases to succeed/break. 😬

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,6 +13,13 @@ jobs:
     name: Deploy docs to GitHub Pages
     runs-on: ubuntu-latest
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PEM }}
+
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Setup node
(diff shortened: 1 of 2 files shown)


dsotirakis (approved):

.github/workflows/deploy.yml
```diff
@@ -13,6 +13,13 @@ jobs:
     name: Deploy docs to GitHub Pages
     runs-on: ubuntu-latest
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92
```
leventebalogh, line 18: Why do we fix the action to this exact version? Does it have any security implications to use a 3rd party action for working with the app-id and private-keys?

leventebalogh (approved):
LGTM, but had one question.

.github/workflows/deploy.yml
jackw: Yep, these [github docs examples](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions) suggest using actions pinned to specific versions.

PR #260, opened 2023-05-26
Tooling: Use NX instead of Lerna for running tasks

jackw:
This PR replaces Lerna with NX for task running. The main incentive here is that local development can benefit from caching to prevent running commands with the same input. Lerna will still be required for versioning and publishing the packages in this repo.

If we're happy to merge this I think we could apply for an [OSS plan for nx cloud](https://nx.app/pricing) and start using the cache in CI.

Fixes #

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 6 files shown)


leventebalogh (approved):
LGTM 👍

academo (approved):
Looks good!

PR #268, opened 2023-05-31
Create Plugin: Lint Deprecation Warnings

### LINKED ISSUE #254: Create Plugin: Make it easier to find deprecated API usage
It might be useful to introduce [the eslint deprecation plugin](https://www.npmjs.com/package/eslint-plugin-deprecation) to create-plugin linting so devs can more easily see which deprecated APIs they're using.

jackw:
Adds an eslint deprecation plugin so plugin developers spot deprecations as they upgrade their `@grafana` dependencies.

Here's an example of what the lint task will print out for deprecations.

```shell
yarn lint
yarn run v1.22.19
$ eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .

/Users/jackwestbrook/dev/grafana/plugin-tools/packages/create-plugin/generated/src/components/AppConfig/AppConfig.tsx
  104:27  warning  'GrafanaTheme' is deprecated. use GrafanaTheme2                    deprecation/deprecation
  126:42  warning  'datasourceRequest' is deprecated. Use the fetch function instead  deprecation/deprecation

✖ 2 problems (0 errors, 2 warnings)

✨  Done in 6.13s.
```

**Note**: It's worth pointing out that whilst this change brings value it also increases the time it takes for the lint command to run.

Fixes #254 

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@2.3.0-canary.268.2121be4.0
  # or 
  yarn add @grafana/create-plugin@2.3.0-canary.268.2121be4.0
  ```
</details>

### DIFF
--- a/docusaurus/docs/create-a-plugin/develop-a-plugin/work-with-frontend.mdx
+++ b/docusaurus/docs/create-a-plugin/develop-a-plugin/work-with-frontend.mdx
@@ -7,7 +7,7 @@ keywords:
   - plugins
   - plugin
   - frontend
-  - CLI 
+  - CLI
 sidebar_position: 1
 ---
 
@@ -39,7 +39,7 @@ Launches the e2e test runner. Make sure to have an instance of Grafana running b
 
 ### <SyncCommand cmd="run lint" />
 
-Runs the linter against source code to verify that you have adhered to the formatting and styling rules.
+Runs the linter against source code to verify that you have adhered to the formatting and styling rules. This command will also report warnings for usage of deprecated code.
 
 ### <SyncCommand cmd="run lint:fix" />
 
--- a/docusaurus/docs/create-a-plugin/extend-a-plugin/extend-configurations.md
+++ b/docusaurus/docs/create-a-plugin/extend-a-plugin/extend-configurations.md
@@ -34,6 +34,23 @@ Edit the `.eslintrc` file in the project root to extend the ESLint configuration
 }
 ```
 
+The following example can be used to disable deprecation notices.
+
+```json
+{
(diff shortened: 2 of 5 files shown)


tolzhabayev (approved):

sympatheticmoose:
@jackw what's the latest with this, was there any more discussion on pro/con of increased lint run time?

tolzhabayev:
@jackw feels like a great addition from my POV, we can then track to see if the usage of deprecated APIs is going down - which I expect it would.

jackw:
@sympatheticmoose @tolzhabayev I've revisited this and I believe the timings aren't a big deal. No one runs these commands manually unless there is a problem and an extra few seconds in CI is (imho) worth it for this feature.

The false positives I was seeing before seem to only affect `atob` and `btoa` node types which can be worked around by adding `window.`. It seems to be a pain point for lots of people. I don't think it'll really impact that many and generally speaking it might be better to reference it to window anyhow.

tolzhabayev:
@jackw if they don't run it locally - will they really spot them? But regardless I think it would be a great addition.

sympatheticmoose (approved):
Generally happy - I know @josmperez is looking at CLI reference docs, but there might be some other places we could update to advertise the feature such as https://grafana.com/developers/plugin-tools/create-a-plugin/develop-a-plugin/work-with-frontend

jackw:
> if they don't run it locally - will they really spot them? But regardless I think it would be a great addition.
Actually I'm wrong, the local dev build script runs linting on changed files. They'd see the deprecation warnings in their IDE, terminal and CI.

PR #276, opened 2023-06-06
Fix 'Install dependencies' command in generated frontend README

sd2k:
It should be yarn/npm install, without the 'run' in between.


jackw (approved):
Thanks so much for spotting and fixing this @sd2k ! 🥇

PR #277, opened 2023-06-07
Docs: Fix snippets rendering after initial load

jackw:
Due to the dynamic import of snippets in CodeSnippets component all pages using the component would appear after the initial render. This PR moves the loading out of the component to the call site where we can rely on docusaurus tooling to generate the HTML for us. Other benefit is we can use markdown to define the language in the snippet rather than "burying" it in the snippet filename. It also adds the `@snippets` alias to make it easier to import snippets.

Fixes #


leventebalogh (approved):
LGTM 👍

PR #278, opened 2023-06-08
Chore: Migrate to NPM

jackw:
**What this PR does / why we need it**:
With the intention to move away from Yarn 1 (without dropping support) this PR migrates the repository to NPM. This allows CI to test plugin generation using NPM instead of Yarn 1. I also took the opportunity to cache the generated app package-lock.json and node_modules folder to try to [speed up CI a bit](https://github.com/grafana/plugin-tools/actions/workflows/ci.yml?query=is%3Asuccess).

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/create-plugin@1.6.3-canary.278.cc64dde.0
  npm install @grafana/sign-plugin@1.0.2-canary.278.cc64dde.0
  # or 
  yarn add @grafana/create-plugin@1.6.3-canary.278.cc64dde.0
  yarn add @grafana/sign-plugin@1.0.2-canary.278.cc64dde.0
  ```
</details>

PR #280, opened 2023-06-09
CI: Introduce Omit Commits

jackw:
This PR introduces Auto Omit Commits package so we can keep certain PRs out of changelogs and versioning decisions.

The [NPM migration PR](https://github.com/grafana/plugin-tools/pull/278) is a good example where we have to edit the dev scripts in the packages but we probably don't want to create noise in the changelog or publish a new patch version based on those changes.

PR also bumps Auto dependencies and sets `atTarget: false` to remove tagging in Slack messages.

Fixes #


leventebalogh (approved):
🚀

PR #297, opened 2023-07-10
Chore: Add comment to PR template explaining labelling

jackw:
Tries to expose some info for PRs regarding which labels to add. Might not be the right place for this but it seems very hidden away right now and very few actually know which labels to add when.

Fixes #


jackw:
Closing this as we feel it would be better addressed with a github action.

Issue #299, opened 2023-07-10
[meta] Cypress 12 migration


academo:
This a meta issue to track progress of the migration to cypress 12.

```[tasklist]
### Migration path to cypress 12
- [ ] Merge https://github.com/grafana/grafana/pull/60117 for Grafana 10.1.x
- [ ] Write migration guide to upgrade from grafana/e2e <=10.0.x  to 10.1.x
- [ ] Update create-tool to detect and correctly ask to update the cypress configuration if grafana/e2e 10.1.x is detected and an old cypress configuration is detected.
- [ ] Update create-tool to generate plugins using grafana >=10.1.x and include the cypress 12 configuration
```

jackw:
This issue has been superseded by the plugin-e2e work.

PR #302, opened 2023-07-11
Do not generate videos for cypress e2e tests

academo:

Adds a cypress.json  file to the common template to not generate videos when running.

The current (older) version of Cypress is slow in specific scenarios in CI when generating videos.

A effort is in place to [migrate to cypress 12](https://github.com/grafana/plugin-tools/issues/299) which should bring a solution to the video generation problem


jackw (approved):
LGTM! 🚀

PR #311, opened 2023-07-20
Website: Add shared alias for includes

jackw:
Adds an alias to docsusaurus to make it easier to embed shared docs in other pages. Once merged docs writers can add shared markdown content to `docusaurus/docs/shared` and then include it in any page like so:

```
import SharedContent from '@shared/path/to/shared/content.md';

<SharedContent />
```

Fixes #


josmperez (approved):
Thanks, this looks like it will be pretty easy to work with.

PR #317, opened 2023-07-28
(chore): migrate to autogen sidebar

sympatheticmoose:
This draft PR shows the required changes to move to an autogenerated sidebar

This recreates the current structure on the site today:

**Current - expanded all**
Screenshot 2023-07-28 at 16 09 14

**New - start**
Screenshot 2023-07-28 at 16 08 52

**New - expanded all**
Screenshot 2023-07-28 at 16 17 33

PR #318, opened 2023-07-28
(chore): tidy sidebars

sympatheticmoose:
Alternative to https://github.com/grafana/plugin-tools/pull/317

Attempts to reorganise the file to make additions more intuitive, shows a way to override the default display name


docusaurus/website/sidebars.js
```diff
@@ -13,55 +13,62 @@
...
+    },
+    {
+      type: 'category',
+      label: 'Development',
+      collapsed: true,
+      items: [
+        'frontend',
+        'backend',
+        {
+          type: 'doc',
```
sympatheticmoose, line 48: This section is unnecessary but shows how we could override the default title metadata of a page to change its appearance in the ToC, without changing the target document. Another alternative would be `sidebar_label: string` in the target document.

PR #320, opened 2023-07-28
Docs: Structure edit with some style corrections of plugin-tools repo

josmperez:
Following the merger of a PR converting site to auto-generated TOC, it became necessary to change the filenames and in some cases folder structures, and associated metadata and links, in order to match the Doc Plan (in other words, to align the existing structure to developers/plugins site.  

I believe that this is the *smallest possible* PR that moves the reorganization project forward without negatively impacting user experience. I have used atomic commits to ease the review process. I have not changed any content except some hypertext link names. I have not moved over any content from developers/plugins at this time.

Note that I was unable to clear one build error which pertains to the rename of the file getting-started to get-started (to match Grafana doc style). Nevertheless, the PR runs and can be previewed.


sympatheticmoose:
I've fixed the link on the homepage, we'll need to review the links on the grafana docs again after any changes like this https://github.com/grafana/grafana/pull/72582

docusaurus/docs/create-a-plugin/_category_.json
```diff
@@ -1,6 +1,6 @@
 {
     "position": 3,
-    "label": "Development",
+    "label": "Create a plugin",
```
sympatheticmoose, line 3: I find it quite confusing to have a Create a plugin section then have Create a plugin page under Get Started

docusaurus/docs/advanced-usage/_category_.json
```diff
@@ -1,6 +0,0 @@
-{
-    "position": 5,
-    "label": "Advanced Usage",
```
sympatheticmoose, line 3: Slightly concerned by losing this section as it may clutter other sections. But not a blocker for me personally.

sympatheticmoose (approved):
Open question on the content of Getting Started vs Create a plugin, otherwise LGTM 👍

docusaurus/docs/create-a-plugin/_category_.json
sympatheticmoose: WDYT @jackw ?

docusaurus/docs/advanced-usage/_category_.json
josmperez: In the doc plan, I relocated the content under the "Advanced" category to the main section for development. My feeling was that "advanced" categories have the potential to be catchalls and not terribly useful... and "usage" seemed a bit odd because the topics pertained to development. I'm open to keeping this section if we think it would be useful, but perhaps we can find a better word than "usage".

docusaurus/docs/create-a-plugin/_category_.json
josmperez: What if we merged the content of both docs/getting-started/ and docs/getting-started/creating-a-plugin? There is already some overlap and I'm not sure that it makes sense to keep these topics separate at this point.

sympatheticmoose: Sure - I'm happy enough with that suggestion 👍

josmperez: Done. creating-a-plugin.mdx is deleted, content incorporated into get-started.mdx

jackw (approved):
Great work on this @josmperez ! 👍

Issue #313, opened 2023-07-20
Bug: go_plugin_build_manifest is deleted during building the frontend


fr-ser:
### Which package(s) does this bug affect?

- [X] Create Plugin
- [ ] Sign Plugin

### Package versions

```shell
1.9.0
```

### What happened?

The go_plugin_build_manifest file, which is created during the build stage of a backend plugin in the dist folder, is removed during the npm run build stage for the frontend.

It seems like that file is missing from the exception list here: https://github.com/grafana/plugin-tools/blob/870397fbc1788f673cc2cbafb7d4b1157a78badd/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts#L130

### What you expected to happen

The go_plugin_build_manifest file - similar to the backend code itself - remains.

### How to reproduce it (as minimally and precisely as possible)

1. Create backend datasource plugin
2. build backend datasource plugin
3. See go_plugin_build_manifest file in the dist folder
4. build frontend plugin (npm run build)
5. see the go_plugin_build_manifest gone

### Environment

```shell
System:
    OS: macOS 13.4.1
    CPU: (8) arm64 Apple M1
    Memory: 122.17 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.20.0 - ~/.asdf/installs/nodejs/16.20.0/bin/node
    npm: 8.19.4 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 114.0.5735.198
    Firefox: 106.0.3
    Safari: 16.5.2
  npmPackages:
    @grafana/data: 9.5.2 => 9.5.2
    @grafana/e2e: 9.5.2 => 9.5.2
    @grafana/e2e-selectors: 9.5.2 => 9.5.2
    @grafana/eslint-config: ^6.0.0 => 6.0.0
    @grafana/runtime: 9.5.2 => 9.5.2
    @grafana/tsconfig: ^1.2.0-rc1 => 1.2.0-rc1
    @grafana/ui: 9.5.2 => 9.5.2
```

PR #322, opened 2023-07-31
Do not remove go_plugin_build_manifest on frontendbuild

academo:

If you build backend before frontend the go manifest file will be removed. This is problematic for certain CI pipelines that follow backend then frontend build steps.

Fixes https://github.com/grafana/plugin-tools/issues/313


jackw (approved):
LGTM! 🚀

PR #331, opened 2023-08-04
sign-plugin: GRAFANA_ACESS_POLICY_TOKEN has been added instaed of GRAFANA_API_KEY

### LINKED ISSUE #323: Support `GRAFANA_ACCESS_POLICY_TOKEN` as default in `sign-plugin` tool
Context:
Adding support for Grafana Cloud Access Policies as the primary method to sign plugins. We want to introduce a new env variable and encourage people to adjust the way they are signing their plugins from using API Keys to Cloud Access Policy Tokens. The longer why can be found in the parent EPIC that this issue is linked to.

**How does it work now:**

Currently the plugin signing command is expecting the environment variable called `GRAFANA_API_KEY` to exist or it will fail with an error saying that you have to set it before signing your plugin.

Here is the corresponding code: 
https://github.com/grafana/plugin-tools/blob/46287dc854793a40fa1a1cd7966570a50ec60e4c/packages/sign-plugin/src/utils/manifest.ts#L78

Here is the API call where this key / token is used:
https://github.com/grafana/plugin-tools/blob/46287dc854793a40fa1a1cd7966570a50ec60e4c/packages/sign-plugin/src/utils/manifest.ts#L90

Here is the corresponding documentation:
https://grafana.com/docs/grafana/latest/developers/plugins/publish-a-plugin/sign-a-plugin/#sign-a-public-plugin

**What do we want to achieve / acceptance criteria:**
- The signing package should first try to read a new env variable called `GRAFANA_ACCESS_POLICY_TOKEN` and use it for signing purposes
- If this env variable is not set, fallback to `GRAFANA_API_KEY` and use it instead
- The rest of the code can stay as is as the new token is provided in the same way to the same endpoint

**Additional todos**
- Add a warning message when `GRAFANA_API_KEY` is used that says:
```
The usage of GRAFANA_API_KEY is deprecated, please consider using GRAFANA_ACCESS_POLICY_TOKEN instead. For more info visit https://grafana.com/docs/grafana/latest/developers/plugins/publish-a-plugin/sign-a-plugin
```

**How to test that you changes are working?**
- Every PR that has a `release` label added to it will get a canary / preview package available. You can use this package to test locally if the signing works following the documentation and using a private plugin as an example. You would need a access policy token for that (ask @tolzhabayev for one or create one yourself).

SCR-20230802-cwm

Ukochka:

- The signing package should first try to read a new env variable called GRAFANA_ACCESS_POLICY_TOKEN and use it for signing purposes
- If this env variable is not set, fallback to GRAFANA_API_KEY and use it instead and show warning message, that usage of GRAFANA_API_KEY here is deprecated
- Provided GitHub actions for sign/releasing of plugins are updated
- Our integration test for sign plugin is updated

Fixes #323

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
 
       - name: '@grafana/sign-plugin - sign generated plugin'
         env:
-          GRAFANA_API_KEY: ${{ secrets.GRAFANA_API_KEY }}
+          GRAFANA_ACCESS_POLICY_TOKEN: ${{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
         run: node ../../sign-plugin/dist/bin/run.js --rootUrls http://www.example.com --signatureType private
         working-directory: ./packages/create-plugin/generated
 
--- a/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GRAFANA_API_KEY: $\{{ secrets.GRAFANA_API_KEY }} # Requires a Grafana API key from Grafana.com.
+      GRAFANA_ACCESS_POLICY_TOKEN: $\{{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
     steps:
       - uses: actions/checkout@v3
 {{#if_eq packageManagerName "pnpm"}}
@@ -65,7 +66,7 @@ jobs:
 
       - name: Sign plugin
         run: {{ packageManagerName }} run sign
-        if: $\{{ env.GRAFANA_API_KEY != '' }}
+        if: $\{{ env.GRAFANA_ACCESS_POLICY_TOKEN != '' }}
 
       - name: Get plugin metadata
         id: metadata
--- a/packages/sign-plugin/src/utils/manifest.ts
+++ b/packages/sign-plugin/src/utils/manifest.ts
@@ -76,18 +76,26 @@ export async function buildManifest(dir: string): Promise<ManifestInfo> {
 
 export async function signManifest(manifest: ManifestInfo): Promise<string> {
   const GRAFANA_API_KEY = process.env.GRAFANA_API_KEY;
-  if (!GRAFANA_API_KEY) {
-    throw new Error(
-      'You must create a GRAFANA_API_KEY env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
-    );
+  const GRAFANA_ACCESS_POLICY_TOKEN = process.env.GRAFANA_ACCESS_POLICY_TOKEN;
+
+  if(!GRAFANA_ACCESS_POLICY_TOKEN) {
+    if (!GRAFANA_API_KEY) {
+      throw new Error(
+        'You must create a GRAFANA_ACCESS_POLICY_TOKEN env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
+      );
+    }
+    else {
+      console.warn(`\x1b[33m%s\x1b[0m`,'The usage of GRAFANA_API_KEY is deprecated, please consider using GRAFANA_ACCESS_POLICY_TOKEN instead. For more info visit https://grafana.com/docs/grafana/latest/developers/plugins/publish-a-plugin/sign-a-plugin')
+    }
   }
 
   const GRAFANA_COM_URL = process.env.GRAFANA_COM_URL || 'https://grafana.com/api';
   const url = GRAFANA_COM_URL + '/plugins/ci/sign';
 
+  const token = GRAFANA_ACCESS_POLICY_TOKEN ? GRAFANA_ACCESS_POLICY_TOKEN : GRAFANA_API_KEY;
   try {
     const info = await postData(url, manifest, {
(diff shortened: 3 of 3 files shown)


packages/create-plugin/templates/github/ci/.github/workflows/release.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GRAFANA_API_KEY: $\{{ secrets.GRAFANA_API_KEY }} # Requires a Grafana API key from Grafana.com.
+      GRAFANA_ACCESS_POLICY_TOKEN: $\{{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
```
academo, line 14: this template will generate plugins with both variable environments but we probably want them to start using `GRAFANA_API_KEY`. What about leaving the `GRAFANA_API_KEY` line commented and adding a comment on top saying that it is "deprecated"?

packages/sign-plugin/src/utils/manifest.ts
```diff
@@ -76,18 +76,26 @@ export async function buildManifest(dir: string): Promise<ManifestInfo> {
 
 export async function signManifest(manifest: ManifestInfo): Promise<string> {
   const GRAFANA_API_KEY = process.env.GRAFANA_API_KEY;
-  if (!GRAFANA_API_KEY) {
-    throw new Error(
-      'You must create a GRAFANA_API_KEY env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
-    );
+  const GRAFANA_ACCESS_POLICY_TOKEN = process.env.GRAFANA_ACCESS_POLICY_TOKEN;
+
+  if(!GRAFANA_ACCESS_POLICY_TOKEN) {
```
academo, line 81: nit: To make this logic easier to read:

```typescript
if(!GRAFANA_ACCESS_POLICY_TOKEN && !GRAFANA_API_KEY) {
	throw new Error(
		'You must create a GRAFANA_ACCESS_POLICY_TOKEN env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
	);
}
if (GRAFANA_API_KEY) {
	console.warn(`\x1b[33m%s\x1b[0m`,'The usage of GRAFANA_API_KEY is deprecated, please consider using GRAFANA_ACCESS_POLICY_TOKEN instead. For more info visit https://grafana.com/docs/grafana/latest/developers/plugins/publish-a-plugin/sign-a-plugin')
}
```

academo (commented):
nice work! 

We should also update the README for the sign-plugin package https://github.com/grafana/plugin-tools/blob/main/packages/sign-plugin/README.md with instructions to use `GRAFANA_ACCESS_POLICY_TOKEN`

.github/workflows/ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
 
       - name: '@grafana/sign-plugin - sign generated plugin'
         env:
-          GRAFANA_API_KEY: ${{ secrets.GRAFANA_API_KEY }}
+          GRAFANA_ACCESS_POLICY_TOKEN: ${{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
```
academo, line 68: To make sure the GRAFANA_API_KEY still work. can you add a new step that uses `GRAFANA_API_KEY` ? just underneath this one. you can adjust the `name` so it is clear one works with policy token and a nother with api_key

academo (approved):

.github/workflows/ci.yml
jackw: I think there is still the previous step `use GRAFANA_API_KEY to sign generated plugin` below this one?

packages/sign-plugin/src/utils/manifest.ts
```diff
@@ -76,18 +76,24 @@ export async function buildManifest(dir: string): Promise<ManifestInfo> {
 
 export async function signManifest(manifest: ManifestInfo): Promise<string> {
   const GRAFANA_API_KEY = process.env.GRAFANA_API_KEY;
-  if (!GRAFANA_API_KEY) {
+  const GRAFANA_ACCESS_POLICY_TOKEN = process.env.GRAFANA_ACCESS_POLICY_TOKEN;
+
+  if(!GRAFANA_ACCESS_POLICY_TOKEN && !GRAFANA_API_KEY) {
     throw new Error(
-      'You must create a GRAFANA_API_KEY env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
+      'You must create a GRAFANA_ACCESS_POLICY_TOKEN env variable to sign plugins. Please see: https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/#generate-an-api-key for instructions.'
```
jackw, line 83: Should this url point to the docs for creating access policy tokens? Right now it points to creating an api key.

packages/create-plugin/templates/github/ci/.github/workflows/release.yml
jackw: As discussed on slack please update the step further down that warns about a missing `GRAFANA_API_KEY`.

jackw (approved):
Code LGTM. Great work! 🚀 

I've left a few nits regarding links to docs not sure if we want to address them before releasing this.

packages/sign-plugin/src/utils/manifest.ts
tolzhabayev: Yes it should point to creating a policy token.

Ukochka: Good catch! Thanks!

PR #339, opened 2023-08-10
Build: Enable NX cloud

jackw:

This PR introduces nx cloud for sharing caches locally and in CI.

I got a merge conflict on package-lock.json during merge. Ran `npm i` and it seems to have deleted ~5000 lines. Further to that I've ran `npm dedupe` to clean it up. 🤷 

Fixes #


.github/workflows/ci.yml
```diff
@@ -6,6 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
     name: Run unit tests
+    env:
+      NX_BRANCH: ${{ github.event.number || github.ref_name }}
```
tolzhabayev, line 10: what are we aiming at here? Branch name for different triggers?

jackw: This env var needs to be set so the nx cloud dashboard has a reference to the github branch. Apparently (I found this on their slack workspace) this is the way to do it. 🤔

tolzhabayev: I mean is `github.event.number` empty in some cases so that we want to fallback to `github.ref_name`? In which cases is that?

jackw: > I mean is github.event.number empty in some cases
Yeah, this workflow runs on push so I believe it's empty when doing anything that isn't associated with a PR (merging to main, tagging). In those cases we grab the reference to the branch/tag instead.

academo (approved):
Let's give it a try!

PR #341, opened 2023-08-11
Switching to dev bucket and manual dev workflow triggers

tolzhabayev:
Switching to the _real_ dev gcs bucket and switching dev deployment to manual triggers


jackw (approved):
Let's go! 🚀

PR #354, opened 2023-08-15
Add troubleshooting.md page for windows and other create-plugin issues

academo:

Adds a guide for windows troubleshooting

Fixes #


packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
+# Windows Troubleshooting
+
+Here are some things to try and check if you experience problems while working with create-plugin or sign-plugin in Windows.
+
+## I am getting: `Unsupported operating system 'Windows' detected. Please use WSL with create-plugin`
```
sympatheticmoose, line 5: Instead of having this as a question, I would just have it as the opening statement.

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+`uname -a`
+
+This command should return something like:
+
+`Linux ....-microsoft-standard-WSL2 #... GNU/Linux`
+
+If you receive an error or a different output, you are not using a WSL terminal.
+
+## I can confirm I am inside a WSL terminal, but I still get `Unsupported operating system 'Windows' detected.`
```
sympatheticmoose, line 23: maybe have this and the next one has h3 subpoints of the above as its progressive diagnosis?

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+You must install Node.js within your WSL environment. You can find a guide on how to install it [here](https://learn.microsoft.com/en-us/windows/dev-environment/javascript/nodejs-on-wsl). We recommend installing the latest LTS version of Node.js.
+
+## I have Node.js installed inside WSL, but I still get `Unsupported operating system 'Windows' detected.`
```
sympatheticmoose, line 27: do you think we need this as a distinct point or can we merge the guidance on "what you can do" into the above point saying to install the latest LTS version of Node.js?

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+### What you can do:
+
+- Check that you are using a version of Node.js >= 18 by running `node --version`.
+- Check if you are using a Linux binary for Node.js and npx by running the following commands: `which node` and `which npx` (note that the command is `which`, not `node --version`). You should receive output similar to `/usr/bin/node` and `/usr/bin/npx`. If the output of these commands is something like `..../Program Files/nodejs/npx`, it means you are using the wrong binary and should install or reinstall Node.js.
+- You can follow the [Microsoft guide](https://learn.microsoft.com/en-us/windows/dev-environment/javascript/nodejs-on-wsl) on how to correctly install Node.js.
+
+NOTE: If you installed Node.js using `sudo apt install nodejs` without following any guide or changing the APT repositories, it is probable that you are not using the latest Node.js LTS.
```
sympatheticmoose, line 37: doesn't feel required 🤷

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+## When I open Grafana with my newly created plugin, Grafana can't load the plugin.
```
sympatheticmoose, line 39: Would this present with a given error message like "Failed to load plugin"? Might be good to include if so.

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+- First, try restarting Grafana to ensure it detects the new plugin.
+- If you are using Native Windows (without WSL), you need to use WSL. Create-plugin does not support Native Windows.
```
sympatheticmoose, line 42: half with you on repeating this at every opportunity 😄 but given your first suggestion is "First <do X>", this isn't a follow up. What should the user try if a restart fails to resolve the issue?

packages/create-plugin/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+## When I run `yarn build` or `yarn dev` in my plugin, I don't see my changes reflected.
```
sympatheticmoose, line 44: I think we _kinda_ default to npm rather than yarn these days. I'd either switch or just say "build and dev commands for my plugin"

academo: Good point!

packages/create-plugin/TROUBLESHOOTING.md
academo: I agree with the points here and later but I intentionally left it that way (repetitive and a bit over-verbose) because developers looking for solutions won't read this page top to bottom but will scan through and expect to see their error message and their solution

josmperez:
Approved with suggestions from my [editorial pass](https://github.com/grafana/plugin-tools/pull/357).

academo:
@josmperez Thanks for the suggestions! In a future time it is much easier if you create suggestions inside this same PR rather than create a new PR for my PR which makes the process longer and harder to review. you can do this in github by using the suggestion tool when you write a comment in a line
image

I left a comment in your PR and approve it. Please merge it so I we can continue the review process here.

josmperez:
> Thanks for the suggestions! In a future time it is much easier if you create suggestions inside this same PR

Generally, I do, unless there are too many suggestions. In this case, I edited for contractions (Grafana style prefers to use contractions) so there too many to do it one by one.

BTW, unfortunately when I changed the filename we lost a precise record of each change relative to the original. In this repo, we changed from manual to auto sidebars, and so I'm trying to ensure that filenames match the title of the doc to make them easier to maintain. That's why I changed the filename.

academo:
@sympatheticmoose anythign missing in this one? can we merge it?

docusaurus/docs/create-a-plugin/extend-configurations.md
```diff
@@ -54,23 +54,7 @@ There are two files in the project root that belong to Jest: `jest-setup.js` and
...
-const { grafanaESModules, nodeModulesToTransform } = require('./.config/jest/utils');
-
-module.exports = {
-  // Jest configuration provided by @grafana/create-plugin
-  ...require('./.config/jest.config'),
-  // Inform Jest to only transform specific node_module packages.
-  transformIgnorePatterns: [nodeModulesToTransform([...grafanaESModules, 'packageName'])],
-};
-```
+If you see `SyntaxError: Cannot use import statement outside a module` when running Jest or `npm run test` see [Troubleshooting](../troubleshooting.md).
```
sympatheticmoose, line 57:
```suggestion
If you see `SyntaxError: Cannot use import statement outside a module` when running Jest or `npm run test` see [Troubleshooting](../troubleshooting.md#i-get-syntaxerror-cannot-use-import-statement-outside-a-module-when-running-jest-or-npm-run-test).
```

docusaurus/docs/troubleshooting.md
```diff
@@ -0,0 +1,77 @@
+---
+id: troubleshooting
+title: Troubleshooting
+sidebar_position: 6
+---
+
+## Troubleshooting
```
sympatheticmoose, line 7:
```suggestion
# Troubleshooting
```

docusaurus/docs/create-a-plugin/extend-configurations.md
sympatheticmoose: makes it easier

docusaurus/docs/troubleshooting.md
sympatheticmoose: other files have h1 so best to be consistent

PR #355, opened 2023-08-15
Add CSS to make links visible inside page content

academo:

Adds custom css to make links visible inside a doc's article

This is a temporal fix while saga provides the solution to the base css.

This is how they look with this css

image

Fixes #


sympatheticmoose (approved):
https://youtu.be/-ITMRHJnaJY?t=45

Perhaps temporary whilst we wait for guidance in https://github.com/grafana/grafana/issues/71069

PR #356, opened 2023-08-15
create-plugin: Enable webpack watchOption -> poll if WSL is detected

### LINKED ISSUE #353: Enable webpack `watchOption` -> `poll` if WSL is detected
When plugin developers work with WSL they have to deal with interoperability between the native windows file system and linux file system.

What often developers do is to run webpack (create plugin `yarn dev`) in their WSL, and use their native windows editor to edit the files.

This often results in WSL not picking up the file changes and thus not rebuilding the plugin as intended.

Webpack's poll option can help to alleviate this problem https://webpack.js.org/configuration/watch/#watchoptionspoll

We don't enable `poll` by default because it can lead to high CPU usage for file watching. Webpack also recommend to make sure not to watch big folders such as `node_modules`.

We can leverage the [detection](https://github.com/sindresorhus/is-wsl/blob/main/index.js) of a wsl environment and set this flag true by default for them so windows users can have a better experience at the cost of a bit more CPU usage.

Ukochka:
To be able to check if file on windows was changed while working in WSL we need to implement poll option for webpack

if we detect WSL we add watchOption -> poll to the config

Fixes #353

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@1.11.0-canary.356.0c5b11c.0
  # or 
  yarn add @grafana/create-plugin@1.11.0-canary.356.0c5b11c.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -200,7 +200,6 @@ const config = async (env): Promise<Configuration> => {
   }
 
   if(isWSL()) {
-    console.log('WSL detected, enabling polling for file changes');
     baseConfig.watchOptions = {
       poll: 3000,
       ignored: /node_modules/,
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -19,7 +19,7 @@ import { SOURCE_DIR, DIST_DIR } from './constants';
 const pluginJson = getPluginJson();
 
 const config = async (env): Promise<Configuration> => {
-    const baseConfig: Configuration = {
+  const baseConfig: Configuration = {
     cache: {
       type: 'filesystem',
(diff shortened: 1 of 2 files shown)


packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -13,189 +13,201 @@ import path from 'path';
...
-      const hasPrefix = (request) => request.indexOf(prefix) === 0;
-      const stripPrefix = (request) => request.substr(prefix.length);
-
-      if (hasPrefix(request)) {
-        return callback(undefined, stripPrefix(request));
-      }
-
-      callback();
+const config = async (env): Promise<Configuration> => {
+    const baseConfig: Configuration = {
```
academo, line 22: nit: this should be indented two spaces less.

academo (approved):

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
Ukochka: fixed!

PR #357, opened 2023-08-15
Docs: Edit Troubleshooting topic

josmperez:
Nice work! Doc review for style, readability. 

Notes:
- Grafana style, like most technical communications style guidelines, discourages use of FAQs. To the extent that some of them have slipped into our doc set over the years, we aim to move the content out of FAQs into the body of the document in a logical place. This is why I have removed "FAQ".
- Grafana style also prefers not to have troubleshooting topics whenever possible for similar reasons, but this isn't a strict rule. I think the contents of this topic are okay, at least for the time being. However, a better long-term approach would be to create a topic called "Install Grafana tools on Windows", located in the "Get started" section, for most of this content (and find suitable places for anything that doesn't fit into that category).


academo:
>Grafana style also prefers not to have troubleshooting topics whenever possible

Strange. it'd sound like ignoring that troubles can occur. The idea behind this troubleshooting guide is that search engines can index these common questions and people find straight answers in the documentation. I particularly dislike when projects hide their problems behind documentation.

docusaurus/docs/troubleshooting.md
academo, line 1: Because you renamed this file, you must also edit `docusaurus/docs/create-a-plugin/extend-configurations.md` that in the original PR now has a link to this file

academo (approved):

docusaurus/docs/troubleshooting.md
```diff
@@ -0,0 +1,77 @@
+---
+id: troubleshooting
+title: Troubleshooting
+sidebar_position: 6
+---
+
+## Troubleshooting
+
+Here are some things to try and check if you experience problems while working with the `create-plugin` or `sign-plugin` tools in Windows.
```
sympatheticmoose, line 9: The Jest error is nothing to do with Windows so I think we'd need to change this as our intro.
```suggestion
Here are some things to try and check if you experience problems while working with the `create-plugin` or `sign-plugin` tools.
```

sympatheticmoose:
@josmperez @academo a potential alternative then if we believe that search engines are the likely entry point, and that it is generally against the Grafana documentation guidelines would be to instead post the questions and answers on our community forum.

docusaurus/docs/troubleshooting.md
josmperez: Fixed.

josmperez: Re: "I particularly dislike when projects hide their problems behind documentation."

I don't want to create the misimpression that we are trying to hide problems. The issue isn't so much what to document as to how to fold in the content into an organizational structure that is more user-friendly. In this case, at some point in the future we should create a topic like "Install plugin tools on Windows" where this good stuff can live.

PR #368, opened 2023-08-17
Docs: Delete file that is duplicate of work-with-backend.md

josmperez:
Delete file that is duplicate of work-with-backend.md.


sympatheticmoose:
Good catch 👍

PR #377, opened 2023-08-28
Docs: Add portal home links

jackw:

- Fixes links in docs so they are using the same text-link colour (blue) across the site without impacting links in specific components inside markdown (e.g. alerts).
- Adds a `portal home` link to the header and footer. (Replaces the docs link in the header which felt a bit redundant given the sidebar is always there.)
- Removes confusing underscores from the answers in get-started which were being rendered in the html.

Fixes #

### DIFF
--- a/docusaurus/website/docusaurus.config.base.js
+++ b/docusaurus/website/docusaurus.config.base.js
@@ -68,7 +68,7 @@ const themeConfigNavbar = {
     src: 'img/logo.svg',
   },
   items: [
-    { href: '/../', label: 'Portal Home', position: 'right' },
+    { href: 'https://grafana.com/developers', label: 'Portal Home', position: 'right' },
     { href: 'https://community.grafana.com/c/plugin-development/30', label: 'Help', position: 'right' },
     {
       href: 'https://www.github.com/grafana/plugin-tools',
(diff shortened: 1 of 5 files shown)


tolzhabayev (approved):
:home:

docusaurus/website/docusaurus.config.base.js
```diff
@@ -68,12 +68,7 @@ const themeConfigNavbar = {
     src: 'img/logo.svg',
   },
   items: [
-    {
-      type: 'doc',
-      docId: 'get-started/get-started',
-      position: 'right',
-      label: 'Docs',
-    },
+    { href: 'https://grafana.com/developers', label: 'Portal Home', position: 'right' },
```
tolzhabayev, line 71: twice?

And probably would make sense to have one for dev and one for prod

sympatheticmoose:
Screenshot 2023-08-29 at 10 08 54

Does it break everything if we change baseURL to `/developers/` so the logo links back to the portal? Within the docs site we've also got the home icon on every single page to take you to `/developers/plugin-tools/` so wonder if this change (if possible) would make it feel more connected to the portal and be a more intuitive link back. 

Alternatively, and this is such a nit pick so I'm sorry, but I wonder if folks will recognise "Portal" so whether "Developer Portal" would be a better link text than "Portal Home"?

PR #381, opened 2023-08-31
Move to underscore from dots in urls

tolzhabayev:
Since we are running behind a reverse proxy in nginx and are actively hiding `index.html` fetches from urls to keep them pretty -> there is no way for nginx to know when to fetch a static asset and when to fetch a `${page}/index.html`

This leads to pages that have dots in url being broken if they have no trailing `/index.html` set. 

To make our lives easier - we should just avoid dots in the url while we are running in reverse proxy mode.


jackw (approved):
LGTM! 🚀

PR #374, opened 2023-08-21
Copy over `query_help.md` when building plugins

mure:
**What this PR does / why we need it**:

Data sources can expose plugin help in markdown: https://github.com/grafana/grafana/blob/546b2deb13ae5d86c03dbd5e639cf0c90296ce42/public/app/core/components/PluginHelp/PluginHelp.tsx#L14

But the webpack config in this repo doesn't copy it over when building.

**Special notes for your reviewer**:

I tested this with an app plugin with nested data sources.

PR #383, opened 2023-09-04
fix: update dockerfile with platform

### LINKED ISSUE #379: Bug: backend plugins fail on M2 Mac

- [X] Create Plugin
- [ ] Sign Plugin

### Package versions

```shell
grafana/create-plugin@1.12.1
```

### What happened?

Backend data source plugins are unavailable

Backend data source plugins to work when running the prescribed commands

After running create you get the [What's Next](https://github.com/grafana/plugin-tools/blob/6bdce9c2a08fa14990a816b7aa8a8fe7d20f277c/packages/create-plugin/src/commands/generate-actions/print-success-message.ts#L25) including `mage -v build:linux`

On a M2 Mac, this cmd generates gpx_fs_linux_amd64.

Then going through running `npm run server`, attempting to `Save & Test` the scaffolded plugin fails with "Plugin Unavailable"

This _might_ be based on the detection being "kinda flakey" on Macs https://github.com/grafana/grafana-plugin-sdk-go/blob/main/build/common.go#L211

If I just run `mage` to generate the `gpx_asd_linux_arm64` then it all works as it should. Confirmed by deleting every other one buildAll adds. 

### Environment

```shell
System:
    OS: macOS 13.5.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 112.22 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.16.0/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Chrome: 116.0.5845.110
    Edge: 116.0.1938.62
    Safari: 16.6
  npmPackages:
    @grafana/data: 9.5.3 => 9.5.3 
    @grafana/e2e: 9.5.3 => 9.5.3 
    @grafana/e2e-selectors: 9.5.3 => 9.5.3 
    @grafana/eslint-config: ^6.0.0 => 6.0.1 
    @grafana/runtime: 9.5.3 => 9.5.3 
    @grafana/tsconfig: ^1.2.0-rc1 => 1.2.0-rc1 
    @grafana/ui: 9.5.3 => 9.5.3
```

### Additional context

We could just suggest users run `mage` instead and get everything.

sympatheticmoose:

Fixes #379 

### DIFF
--- a/packages/create-plugin/templates/common/.config/Dockerfile
+++ b/packages/create-plugin/templates/common/.config/Dockerfile
@@ -1,7 +1,7 @@
 ARG grafana_version=latest
 ARG grafana_image=grafana-enterprise
 
-FROM grafana/${grafana_image}:${grafana_version}
+FROM --platform=linux/amd64 grafana/${grafana_image}:${grafana_version}
 
 # Make it as simple as possible to access the grafana instance for development purposes
 # Do NOT enable these settings in a public facing / production grafana instance


packages/create-plugin/templates/common/.config/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG grafana_version=latest
 ARG grafana_image=grafana-enterprise
 
-FROM grafana/${grafana_image}:${grafana_version}
+FROM --platform=linux/amd64 grafana/${grafana_image}:${grafana_version}
```
mckn, line 4: I think we can skip this one and keep everything in the `docker-compose` file.

packages/create-plugin/templates/common/docker-compose.yaml
```diff
@@ -5,6 +5,8 @@ services:
     container_name: '{{ pluginId }}'
```
mckn, line 5:
```suggestion
    container_name: '{{ pluginId }}'
    platform: "linux/amd64"
```

mckn (approved):
LGTM! Added a small suggestion on how we can keep all of it in one place.

sympatheticmoose:
Turns out we just needed it in the place suggested by @mckn, nice and simple and seems to work.

mckn (approved):
LGTM!

packages/create-plugin/templates/common/docker-compose.yaml
```diff
@@ -3,6 +3,7 @@ version: '3.0'
 services:
   grafana:
     container_name: '{{ pluginId }}'
+    platform: "linux/amd64"
```
mountainash, line 6: This change breaks prettier.

[It's set](https://github.com/grafana/plugin-tools/blob/main/.prettierrc) to single-quotes `platform: 'linux/amd64'`.

Maybe there should be a linter setup to run on PR's on this project? I also found another break at https://github.com/grafana/plugin-tools/pull/374#issuecomment-1884786130 ?

mckn: Thanks for reporting this and it should be fixed in: https://github.com/grafana/plugin-tools/pull/657

Also, we have opened an issue to prevent malformed templates into the repo.

PR #388, opened 2023-09-07
Templates/Webpack: Include full path in dev mode image output filename

domasx2:

:wave:  Having multiple images with same filename but under different folders in `img/` causes webpack to error in dev mode because output path only includes the filename. For example we have `img/technology/generic.svg` and `/img/cloud/generic.svg' and it tries to ouput both into `img/generic.svg`, causing a conflict.

This PR makes dev output filename use the full img path, fixing this issue

### DIFF
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -110,7 +110,7 @@ const config = async (env): Promise<Configuration> => {
             // Keep publicPath relative for host.com/grafana/ deployments
             publicPath: `public/plugins/${pluginJson.id}/img/`,
             outputPath: 'img/',
-            filename: Boolean(env.production) ? '[hash][ext]' : '[name][ext]',
+            filename: Boolean(env.production) ? '[hash][ext]' : '[file]',
           },
         },
         {


packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -110,7 +110,7 @@ const config = async (env): Promise<Configuration> => {
             // Keep publicPath relative for host.com/grafana/ deployments
             publicPath: `public/plugins/${pluginJson.id}/img/`,
             outputPath: 'img/',
-            filename: Boolean(env.production) ? '[hash][ext]' : '[name][ext]',
+            filename: Boolean(env.production) ? '[hash][ext]' : '[file]',
```
jackw, line 113: I'm not sure how intuitive it is to keep the folder structure in dev and not in prod. With the current suggestion will the same image file in multiple directories cause webpack to error when doing a prod build?
Would something like this work for both build envs?

- development: `[file]`
- production: `[path][hash][ext]`

domasx2: > will the same image file in multiple directories cause webpack to error when doing a prod build?

As long as images are not identical, prod build is fine because the hash is different

> Would something like this work for both build envs?
>
> development: [file]
> production: [path][hash][ext]

Sure, but I don't think it matters much as long as output files have unique names. Could also change dev to `[name][hash][ext]` to fix the issue. Which do you prefer?

jackw: Let's leave it as you have it now and get the fix shipped. 👍

jackw (approved):
Thanks for fixing this @domasx2 ! 🚀

PR #397, opened 2023-09-12
Excluding dependabot from steps that require secrets

tolzhabayev:
Excluding dependabot from steps that require secrets as dependabot to prevent issues like this from happening: https://github.com/grafana/plugin-tools/actions/runs/6160248058/job/16716747383


xnyo (approved):
LGTM!

PR #402, opened 2023-09-13
Changed so UI extensions section isnt collapsed by default

mckn:
Changed so UI extensions section is not collapsed by default.

Fixes #


sympatheticmoose (approved):
This makes it collapsed by default, which I think is good, title confused me!

PR #406, opened 2023-09-14
Docs: move error handling topic

josmperez:
I propose to move the error handling topic out of Introduction into Create a plugin/Develop a plugin. As it contains a mix of conceptual and how-to information, the latter seems more appropriate.


josmperez:
Oops, fixed in a new PR.

PR #407, opened 2023-09-14
Docs: move error handling topic

josmperez:
I propose to move the error handling topic out of Introduction into Create a plugin/Develop a plugin. As it contains a mix of conceptual and how-to information, the latter seems more appropriate.


sympatheticmoose (approved):
Looking at the relative content in Develop vs Extend, I actually think this might fit better in the latter. But happy to go with your preference 👍

PR #409, opened 2023-09-18
Docs: add new section to get-started

josmperez:
Add a new section to the top of "Get started" to introduce the documentation and describe the types of plugins available. New H1 says "Grafana plugin developer's guide" instead of "Get started".

### DIFF
--- a/docusaurus/docs/get-started/get-started.mdx
+++ b/docusaurus/docs/get-started/get-started.mdx
@@ -15,31 +15,20 @@ import ScaffoldNPM from '@snippets/createplugin-scaffold.npm.md';
 import ScaffoldPNPM from '@snippets/createplugin-scaffold.pnpm.md';
 import ScaffoldYarn from '@snippets/createplugin-scaffold.yarn.md';
 
-# Grafana plugin developer's guide
+# Get started with plugins
 
-Extend Grafana's built-in capabilities with plugins. Plugins enable Grafana 
-to accomplish specialized tasks, custom-tailored to your requirements. 
-By making a plugin for your organization, you can connect Grafana to other 
-data sources, ticketing tools, and CI/CD tooling.
+Extend Grafana's built-in capabilities with plugins custom-tailored 
+to your requirements. This guide is designed to help you quickly get up 
+and running with a working plugin.  
 
-You can create plugins for private use or contribute them to the open 
-source community by publishing to the Grafana 
-[plugin catalog](https://grafana.com/plugins). 
-This catalog has hundreds of other community and commercial plugins.
+From here, you can explore the rest of our documentation,
+[SDK for Go](../introduction/grafana-plugin-sdk-for-go), and 
+[plugin examples](https://github.com/grafana/grafana-plugin-examples/) to build 
+your knowledge and add more advanced capabilities. And when you're ready, you can
+publish your plugin in the Grafana [plugin catalog](https://grafana.com/plugins). 
 
-## Types of plugins
-
-You can create several types of plugins, including:
-
-* **Panel plugins** - Visualize data and navigate between dashboards.
-* **Data source plugins** - Link to new databases or other sources of data.
-* **App plugins** - Create rich applications for custom out-of-the-box experiences.
-
-## Plugin tools
-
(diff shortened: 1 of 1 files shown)


docusaurus/docs/get-started/get-started.mdx
```diff
@@ -15,6 +15,28 @@ import ScaffoldNPM from '@snippets/createplugin-scaffold.npm.md';
 import ScaffoldPNPM from '@snippets/createplugin-scaffold.pnpm.md';
 import ScaffoldYarn from '@snippets/createplugin-scaffold.yarn.md';
 
+# Grafana plugin developer's guide
```
sympatheticmoose, line 18: I think if we add this header, we need more of an index of links / journey. 

The hypothesis (this could be wrong) is that we want getting started (generally) to be a "quick start" so a developer can get something up and running, and then explore the working code and the rest of this site to gain deeper knowledge as and when appropriate. 

If we remove this header, I think we can still provide a basic introduction to improve the getting started experience and partly address the feedback that led to this PR.

docusaurus/docs/get-started/get-started.mdx
```diff
@@ -15,6 +15,28 @@ import ScaffoldNPM from '@snippets/createplugin-scaffold.npm.md';
...
+
+## Types of plugins
+
+You can create several types of plugins, including:
+
+* **Panel plugins** - Visualize data and navigate between dashboards.
+* **Data source plugins** - Link to new databases or other sources of data.
+* **App plugins** - Create rich applications for custom out-of-the-box experiences.
+
+## Plugin tools
```
sympatheticmoose, line 38: we could extend below the content here, before the quick start to elaborate. Something like this:

"This guide is designed to help you quickly get up and running with a working plugin. From here you can explore the rest of our documentation and [examples](https://github.com/grafana/grafana-plugin-examples/) to build your knowledge and add more advanced capabilities."

docusaurus/docs/get-started/get-started.mdx
```diff
@@ -15,6 +15,28 @@ import ScaffoldNPM from '@snippets/createplugin-scaffold.npm.md';
...
+to accomplish specialized tasks, custom-tailored to your requirements. 
+By making a plugin for your organization, you can connect Grafana to other 
+data sources, ticketing tools, and CI/CD tooling.
+
+You can create plugins for private use or contribute them to the open 
+source community by publishing to the Grafana 
+[plugin catalog](https://grafana.com/plugins). 
+This catalog has hundreds of other community and commercial plugins.
+
+## Types of plugins
```
sympatheticmoose, line 30: Unfortunately adding this here introduces duplication with https://grafana.com/developers/plugin-tools#what-type-of-plugin-would-you-like slightly further down. Maybe that's fine, but feels a bit strange. Also we should ensure the descriptions match.

sympatheticmoose (commented):
Made some comments and suggestions

docusaurus/docs/get-started/get-started.mdx
josmperez: I think what we probably need is a hybrid of an Intro and Get started with an emphasis on the Get started. So, 've changed the top header to "Get started with plugins" and then streamlined the following text so that there is a bit of an intro but not so much that it gets in the way of the user starting their journey.

docusaurus/docs/get-started/get-started.mdx
josmperez: In my revision I have removed the section on types of plugins. It's possible we may want to add a new topic "Types of plugins" to the Introduction section at some point, but for now I think the short descriptions we give lower on the page are fine.

docusaurus/docs/get-started/get-started.mdx
josmperez: Agreed, removed.

josmperez:
@sympatheticmoose thanks for helpful feedback, I'm taking this PR out of draft into a proposed submission.

josmperez:
Moving to draft. I have a different idea on how to address this in the "Introduction" section later.

josmperez:
Closing in favor of approach taken in https://github.com/grafana/plugin-tools/pull/479

PR #413, opened 2023-09-19
Docs: Add three indexes as temporary landing pages

josmperez:
Add indexes for Introduction, Tutorials, and Migration Guide sections using the DocCardList Docusaurus shortcode. The purpose of these index is to serve as temporary landing pages while we develop introductory content to help navigate users through this content. Also, they are helpful so that the old Hugo doc content can be migrated to parallel locations in the doc set.


jackw (approved):
LGTM! 🚀

PR #418, opened 2023-09-21
Docs: add plugin spellcheck internal doc

josmperez:
Add internal doc on plugin spellcheck feature.

Per internal Slack conversation, we are keeping this in draft. Moving from grafana repo to plugin-tools repo. Related internal discussion at https://raintank-corp.slack.com/archives/C024NL6AEJH/p1683014954787059?thread_ts=1682980023.733219&cid=C024NL6AEJH

### DIFF
--- a/docusaurus/docs/create-a-plugin/develop-a-plugin/plugin-spellcheck.md
+++ b/docusaurus/docs/create-a-plugin/develop-a-plugin/plugin-spellcheck.md
@@ -1,0 +1,79 @@
+---
+id: plugin-spellcheck
+title: Set up GitHub workflows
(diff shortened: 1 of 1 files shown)


sympatheticmoose (approved):
fine by me 👍 

@cletter7 with the file deleted and now moved here, I guess there's a change needed somewhere in the drone CI?

> We link from CI to this file in case spellcheck is not configured.

Also don't know if its worth updating the library since is a year old now? Maybe fine, but if the idea is that for new plugins which don't have this configured to begin with get a message to use it, they may want to start with a newer release.

PR #419, opened 2023-09-22
Cleanup commented out workflow code and remove the noindex

tolzhabayev:
Since we have started redirecting the docs content to developer portal, there is no need for noIndex anymore and we can let the content get into SERPs.

Referenced PR https://github.com/grafana/website/pull/15415


jackw (approved):
LGTM! 🚀

PR #422, opened 2023-09-22
Docs: Add plugin.schema.json

josmperez:
Transfering this file from its old location on the Grafana developers/plugins site. May or may not be needed. (?)


josmperez:
We are keeping this in grafana repo.

PR #423, opened 2023-09-25
Docs: Add workflows for tech docs

josmperez:
Add workflows for tech docs process -- see https://github.com/grafana/technical-documentation/wiki/Project-management#about-the-typedocs-label-workflow for details

### DIFF
--- a/.github/workflows/issue_commands.json
+++ b/.github/workflows/issue_commands.json
@@ -1,0 +1,10 @@
+[
+    {
+      "type": "label",
+      "name": "type/docs",
+      "action": "addToProject",
+      "addToProject": {
+        "url": "https://github.com/orgs/grafana/projects/69"
+      }
+    }
+  ]
--- a/.github/workflows/typedocs.yml
+++ b/.github/workflows/typedocs.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Actions
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: grafana/grafana-github-actions
           path: ./actions
--- a/.github/workflows/typedocs.yml
+++ b/.github/workflows/typedocs.yml
@@ -1,0 +1,24 @@
+name: Run commands when issues or PRs are labeled
+on:
+  issues:
+    types: [ labeled ]
+  pull_request:
+    types: [ labeled ]
+jobs:
+  main:
+    if: ${{ github.event.pull_request == false || github.event.pull_request.head.repo.full_name == github.repository }}
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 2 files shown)


.github/workflows/issue_commands.json
```diff
@@ -0,0 +1,10 @@
+[
+    {
+      "type": "label",
+      "name": "type/docs",
```
sympatheticmoose, line 4: I'd guess this wont work since we don't have that label, we have `documentation`

sympatheticmoose (changes requested):

.github/workflows/issue_commands.json
jackw: Yeah I don't think it'll work either. We need to add the label and probably best to then delete the `documentation` label.

jackw: I believe it would be better if there is a convention in the docs team to use `type/docs` across repos.

sympatheticmoose: done

.github/workflows/typedocs.yml
```diff
@@ -0,0 +1,24 @@
...
+    types: [ labeled ]
+  pull_request:
+    types: [ labeled ]
+jobs:
+  main:
+    if: ${{ github.event.pull_request == false || github.event.pull_request.head.repo.full_name == github.repository }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v3
```
sympatheticmoose, line 13: https://github.com/grafana/plugin-tools/pull/395 we should probably use 4 as we've just bumped this in another action

josmperez, line 13:
```suggestion
        uses: actions/checkout@v4
```

jackw (approved):
LGTM! 🚀

sympatheticmoose (commented):
We can do this, but it feels like overkill to me, its doing a lot just to add things to a project. Then its also 2 files being a mix of json/yaml.

I think the following would suffice in all honesty:

```
name: Add docs issues to project

on:
  pull_request:
    types:
      - labeled
  issues:
    types:
      - labeled
jobs:
  add-to-project:
    name: Add issue to project
    runs-on: ubuntu-latest
    steps:
      - uses: actions/add-to-project@v0.5.0
        with:
          project-url: https://github.com/orgs/grafana/projects/69
          github-token: ${{ secrets.ISSUE_COMMANDS_TOKEN }}
          labeled: type/docs
```
The token may have more privileges than it strictly requires, but if we're going to use it anyway 🤷

josmperez:
> We can do this, but it feels like overkill to me

It's not clear to me how to take action on this comment. Can you please clarify?

sympatheticmoose (approved):
go ahead :)

PR #431, opened 2023-10-02
Adjusting dependabot dependency labels

tolzhabayev:
Should theoretically control what labels are set, not sure it practically works but let's see.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,16 +2,16 @@
 
 version: 2
 updates:
-
+  # Check for updates to GitHub Actions every month
   - package-ecosystem: "github-actions"
     directory: "/"
-    schedule:
-      # Check for updates to GitHub Actions every month
+    schedule:      
       interval: "monthly"
     labels:
         - "github_actions"
         - "dependencies"
         - "no stalebot"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -2,9 +2,22 @@
...
 version: 2
 updates:
-
+  # Check for updates to GitHub Actions every month
   - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:      
+      interval: "monthly"
+    labels:
+        - "github_actions"
```
jackw, line 11: I think we should probably add this to the `omit-commits` plugins in `.autorc` so it is never mentioned in changelog etc. Looking at plugin source code. e.g.

```
[
  "omit-commits",
  {
    "labels": ["no-changelog", "github_actions"]
  }
],
```

tolzhabayev: https://github.com/grafana/plugin-tools/pull/431/commits/8f4f3c33f8cc8c42e38b048cc22e94c9e77360cd

jackw (approved):
LGTM! 🚀 

Hopefully the changes in dependabot.yml prevent the addition of the semver label. 🤞

Issue #310, opened 2023-07-19
Bug: plugin signature not working behind a proxy


papaux:
### Which package(s) does this bug affect?

- [ ] Create Plugin
- [X] Sign Plugin

### Package versions

```shell
@grafana/sign-plugin@1.0.3
```

### What happened?

After the upgrade to v1.0.3 of sign-plugin, we cannot anymore sign plugins when running behind a HTTPS proxy.

The issue has been introduced by this change, where it replaced `axios` by a plain request from `https` module:
https://github.com/grafana/plugin-tools/pull/292

Axios honors out of the box the `https_proxy` config from the environment if set, as highlighted here: https://www.scrapingbee.com/blog/nodejs-axios-proxy/

Which is not the case with https module, it requires explicit proxy config as far as I understand.

I confirmed by reverting to @grafana/sign-plugin@1.0.2 and the plugin signature works as expected.

Maybe there is another way to set the proxy, but I don't see it from the config here:
https://nodejs.org/api/http.html#httprequestoptions-callback

### What you expected to happen

Signature should work when behind a proxy.

### How to reproduce it (as minimally and precisely as possible)

1. Set the environment variable `https_proxy` to use a proxy server
2. Close all direct internet access
3. Run plugin signature with the old / new version

### Environment

```shell
N/A but environment should not matter.
```

academo:
Hi @papaux would you be able to contribute a PR to include support for https_proxy ?

Waxolunist:
Why not instead use again axios? This was an unneccessary change.

academo:
> Why not instead use again axios? This was an unneccessary change.

Axios was causing different problems  in other scenarios see https://github.com/grafana/plugin-tools/issues/290

I opened a PR to enable https_proxy support here https://github.com/grafana/plugin-tools/pull/436

PR #436, opened 2023-10-03
Allows to sign behind an https proxy

academo:

ALlows to sign plugins behind an https proxy. nodejs does not handle the https_proxy environment variable natively. using `proxy-agent` allow us to have a full support of different proxies in the package with minimal code changes

Fixes https://github.com/grafana/plugin-tools/issues/310

Fixes #

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6758,6 +6758,11 @@
         "node": ">= 10"
       }
     },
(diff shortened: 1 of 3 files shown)


jackw (approved):
LGTM! 🚀 

If this is a bug fix I think we should probably label as patch. Don't forget to add `release` label if you'd like this PR to publish a new version of the package.

PR #447, opened 2023-10-04
<FeatureArea>: Add new flag "bundle" to sign command

### LINKED ISSUE #411: advanced build action and documentation
This issue is to provide a more advanced github action that creates zips per architecture, alongside updating our documentation to advise on when it makes sense to choose this option and how it can be used with our plugin catalog submission process

gabrielthomasjacobs:

When trying to bundle a plugin for submission, the process is manual and can be tedious; especially when making typos or forgetting a step.

This adds a new flag `--bundle` to the sign command. When running the sign command with the bundle flag, this will:

1. Copy the dist folder contents into a folder matching the plugin id (lifted from dist/plugin.json)
2. Updates the permissions of the executable files to 0755 (*nix systems only)
3. Zip that folder matching the syntax {pluginId}-{version}.zip, satisfying the submission requirements 
4. Delete the folder created in step 1
5. Calculates and logs the md5 and sha1 of the zipped folder, to make submission simple

Some notes:
+ This doesn't fix #411, but would be a great first-step if accepted; the next step would be to allow a flag value, e.g. `--bundle="multiple"` and then handle that case. If this approach is acceptable, then I will add that functionality as well.

+ scaffolding and zipping works on *nix and windows systems.   

+ Setting permissions works on *nix systems only. Setting permissions on windows *won't* work, even using WSL, due to how windows manages permissions.

Extra feedback request:

+ Since *nix permissions don't work on Windows and can't persist to a unix environment, should we instead throw an error and instruct the user to run the command on a unix machine instead or that the flag only works on unix machines? The only benefit to a windows user really is that the folder name/structure would be present to imitate. If so, I'll also strip out the windows-specific portions.

+ I placed this as a flag for the sign command, since the workflow is going to be sign -> bundle for the vast majority of the user workflows and the sign command already has the dist dir set. I can break this out into a fully new command though if that would be preferred.

### DIFF
--- a/packages/sign-plugin/src/commands/bundle.command.ts
+++ b/packages/sign-plugin/src/commands/bundle.command.ts
@@ -1,4 +1,4 @@
-import { chmodSync, mkdirSync, readdirSync, copyFileSync, rmSync, readFileSync } from "fs";
+import { mkdirSync, readdirSync, copyFileSync, rmSync, readFileSync } from "fs";
 import { platform } from 'os';
 import { getPluginJson } from "../utils/pluginValidation";
 import { resolve, join } from 'path';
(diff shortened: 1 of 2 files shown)


sympatheticmoose (changes requested):
Hi @gabrielthomasjacobs apologies for the delay in getting back to you. This is a good idea and thank you for taking the initiative to propose the fix. I'd recommend raising issues for enhancements first, this way we can provide some feedback or guidance before any code gets written. 

This probably goes beyond the scope of responsibilities for `sign-plugin`, but could be provided through a different mechanism - our recommendation would be a script which could be distributed within `.config`.  I've raised https://github.com/grafana/plugin-tools/issues/491 to track this request.

gabrielthomasjacobs:
Ok, thanks @sympatheticmoose for the feedback and for creating #491, I'll subscribe to that ticket and wait for triage/feedback!

sympatheticmoose:
Hey @gabrielthomasjacobs - wrt the adv build action and the zips per architecture - I'm curious if this was a feature of interest to you and if you had a use case for it?

gabrielthomasjacobs:
Hey @sympatheticmoose, I'm planning on using the github actions to handle the building/releasing as described in https://github.com/grafana/plugin-actions/blob/main/build-plugin/action.yml

As for the multiple zips, it would be a nice to have, but it's my understanding that community plugins aren't currently supported for this type of distribution.

PR #449, opened 2023-10-05
Chore: Run prettier

zoltanbedi:
I ran prettier in the whole repo but I'm mostly interested in the things under the template/common. Also bumped the nvmrc from 16 to 18 hope that makes sense.


sunker (approved):
Makes sense and looks good to me!

packages/create-plugin/templates/common/npmrc
```diff
@@ -1,14 +1,18 @@
...
 
 # PNPM 8 changed the default resolution mode to "lowest-direct" which is not how we expect resolutions to work
+
 resolution-mode="highest"
 
 # Make sure the default patterns are still included (https://pnpm.io/npmrc#public-hoist-pattern)
-public-hoist-pattern[]="*eslint*"
-public-hoist-pattern[]="*prettier*"
+
+public-hoist-pattern[]="_eslint_"
```
jackw, line 9: @zoltanbedi I think this will break pnpm scaffolded plugin support.

jackw:
@zoltanbedi I'm sorry but would you mind if we revert this and do it in smaller chunks please? I think we also should consider getting a precommit hook in place otherwise this type of PR will need to happen again and again. For templates it might be better to run a command post scaffold to make sure they are "prettier approved". WDYT?

PR #450, opened 2023-10-06
Revert prettier formatting

zoltanbedi:


jackw (approved):
Thanks for dealing with this so quickly @zoltanbedi ! 🚀

PR #456, opened 2023-10-11
clarify that preloading is only applicable to app plugins

sunker:

Clarify that the `preload` setting is only applicable for app plugins.


sympatheticmoose (approved):
Nice - this fixes the issue 👍 

We could potentially clarify more about when exactly it loads

mckn (approved):
LGTM!

jackw (approved):
LGTM! 🚀

PR #463, opened 2023-10-16
CI: Split gh workflow generate commands into a matrix

jackw:
This PR splits the `generate-<pluginType>` commands into a matrix to try to improve the times and also so we can run commands as individual steps against the scaffolded plugin rather than doing it as a one liner of `npm exec nx run @grafana/create-plugin:generate-app-backend && (cd ./packages/create-plugin/generated && npm install && npm run build && npm run lint && mage -v build:linux)`.

To speed up these tasks further we need to figure out what is going on with `npm ci` which seems to be taking ages.

| Before |
| --- |
| Screenshot 2023-10-17 at 12 09 29 |

| After |
| --- |
| Screenshot 2023-10-17 at 12 09 41 |

Fixes #


.github/workflows/ci.yml
```diff
@@ -45,54 +34,119 @@ jobs:
...
 
+  get-commands:
+    name: Get scaffold commands
+    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
+    runs-on: ubuntu-latest
+    outputs:
+      scaffoldCommands: ${{ steps.set-matrix.outputs.scaffoldCommands }}
+    steps:
+      - id: set-matrix
+        run: echo "scaffoldCommands=[{\"cmd\":\"generate-app\",\"hasBackend\":false},{\"cmd\":\"generate-app-backend\",\"hasBackend\":true},{\"cmd\":\"generate-panel\",\"hasBackend\":false},{\"cmd\":\"generate-datasource\",\"hasBackend\":false},{\"cmd\":\"generate-datasource-backend\",\"hasBackend\":true},{\"cmd\":\"generate-scenes-app\",\"hasBackend\":false}]" >> $GITHUB_OUTPUT
```
andresmgot, line 45: I am not sure I understand why you want this to be generated with a bash command. I don't see anything dynamic?

.github/workflows/ci.yml
```diff
@@ -45,54 +34,119 @@ jobs:
...
+
+  generate-plugins:
+    name: Test plugin scaffolding
+    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
```
andresmgot, line 49: maybe it's better to use a label for this? Also, why would you want to skip the CI?

.github/workflows/ci.yml
```diff
@@ -45,54 +34,119 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        scaffoldCommand: ${{ fromJson(needs.get-commands.outputs.scaffoldCommands) }}
+    env:
+      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_BRANCH: ${{ github.event.number || github.ref_name }}
+    steps:
+      - run: echo ${{ matrix.scaffoldCommand.cmd }}
+      - run: echo ${{ matrix.scaffoldCommand.hasBackend }}
```
andresmgot, line 60: Nit:
```suggestion
      - run: echo "Command: ${{ matrix.scaffoldCommand.cmd }}"
      - run: echo "Has backend: ${{ matrix.scaffoldCommand.hasBackend }}"
```
But you can also remove this, seems more for debugging purposes

.github/workflows/ci.yml
```diff
@@ -45,54 +34,119 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          # We need to fetch all branches and commits so that Nx affected has a base to compare against.
+          fetch-depth: 0
+      - uses: nrwl/nx-set-shas@v3
+      - name: Setup .npmrc file for NPM registry
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
```
andresmgot, line 71: should this be 20?

.github/workflows/ci.yml
jackw: Thanks for questioning this. I hadn't really put much thought into doing it any other way. Maybe it would make more sense if we read from a json file?

`echo "scaffoldCommands=$(jq -c . < ./scaffoldCommands.json)" >> $GITHUB_OUTPUT`

.github/workflows/ci.yml
jackw: Yep! Will update it. 👍

.github/workflows/ci.yml
andresmgot: can't you simply define the matrix as?:

```
    strategy:
      matrix:
        cmd: [generate-app, generate-app-backend, generate-panel, generate-datasource, generate-datasource-backend, generate-scenes-app]
        hasBackend: [true, false]
```

I guess you don't want weird combinations like `cmd: generate-datasource-backend, hasBackend: false` so I think you can also do:

```
    strategy:
      matrix:
        include:
        - cmd: generate-app
          hasBackend: false
        - cmd: generate-app-backend
          hasBackend: true
          ...
```
according to https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

andresmgot:
LGTM!

PR #466, opened 2023-10-16
Docs: Refactor 'Get started' section

### LINKED ISSUE #468: Docs: Need better organization in 'Get started' section
Existing content arrangement is confusing and sometimes redundant.

josmperez:
Refactor 'Get started' section for usability, reducing redundancy. Highlights:
- Tighten up 'Get started' main page to make it more breezy
- Move CLI reference to its own page
- Remove duplication of Folder Structure
- Move prerequisites to Development Environment topic

Fixes https://github.com/grafana/plugin-tools/issues/468

### DIFF
--- a/docusaurus/docs/get-started/cli-commands.md
+++ b/docusaurus/docs/get-started/cli-commands.md
@@ -1,0 +1,36 @@
+---
+id: cli-commands
+title: Key CLI commands
+description: The CLI commands you need to know for plugin development.
+keywords:
+  - grafana
+  - plugins
+  - plugin
+  - create-plugin
+  - CLI
+  - command line
+  - commands
+sidebar_position: 3
+---
+
+# Key CLI commands
+
+After `create-plugin` has finished, you can run the following built-in commands in the shell:
+
+## <SyncCommand cmd="run dev" />
+
+Builds the plugin in _development mode_ and runs in _watch mode_. Rebuilds the plugin whenever you make changes to the code. You'll see build errors and lint warnings in the console.
+
+## <SyncCommand cmd="run test" />
+
+Runs the tests and watches for changes.
+
+## <SyncCommand cmd="run build" />
+
+Creates a production build of the plugin that optimizes for the best performance. Minifies the build and includes hashes in the filenames.
+
+<h2>
+  <code>mage -v</code>
+</h2>
+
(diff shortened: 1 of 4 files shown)


docusaurus/docs/get-started/cli-commands.md
sympatheticmoose, line 1: I would _typically_ expect CLI commands to come under `/reference/`

Do you think we would have a full reference as well as this kind of quick start?

docusaurus/docs/get-started/cli-commands.md
```diff
@@ -0,0 +1,36 @@
...
+  - create-plugin
+  - CLI
+  - command line
+  - commands
+sidebar_position: 3
+---
+
+# Key CLI commands
+
+After `create-plugin` has finished, you can run the following built-in commands in the shell:
```
sympatheticmoose, line 18: You can't run these without running install first

sympatheticmoose (commented):
I like the idea of this, but I'm worried that `getting-started` is becoming just "run create" and we're not guiding through the user through the actual running of the scaffolded plugin. We _kind_ of have this in `cli-commands` but we don't link to this and I actually think you want the end to end on a single page. I would expect get started to give me a quick start to scaffolding a new plugin, building it, and running both the frontend and backend in Grafana so I can see it in action.

docusaurus/docs/get-started/cli-commands.md
josmperez: You raise a good and important point: what exactly are we trying to achieve on this (front) page? To overly simplify a bit, it's my understanding that a "Quick Start" is basically an installation and configuration guide, and a "Getting Started" is that, plus basic usage, and some conceptual information as well. 

So, since this page is labeled both "Get started" and "Quick Start", we have an ambiguous purpose. There are many ways to resolve ths, and I think the approach you suggest is one that would work fine for me by making the page a true "Get started" page (although I was going for more of a "Quick Start" guide). The way you use these two page labels may differ from mine, and that's fine so long as we agree on the ultimate purpose of the page.

I will make the changes to this PR tomorrow.

Also, regarding CLI commands under /reference/, Jack has helped me by creating a comprehensive list of CLI commands that gives us what we need to create a true reference. I will also add that page too soon.

josmperez:
Replaced with https://github.com/grafana/plugin-tools/pull/479

PR #476, opened 2023-10-17
Docs: Fix broken links in package files

josmperez:
Fix broken links in package files.


jackw (approved):
LGTM! 🚀 

I changed the label from `no-changelog` to `patch` so when a new release goes out this will get a mention in the changelog as they are updates to files in the create-plugin code.

PR #489, opened 2023-10-24
Website: Fix tracking issue with missing canonical links

jackw:

@josmperez has reported some odd behaviour with docusaurus crashing in dev. Looking at the stack trace it appears the tracking code will error if it cannot find a canonical link on the page. This PR fixes that by checking the link exists before attempting to get the `href` attribute.

image

Fixes #

### DIFF
--- a/docusaurus/website/src/theme/tracking/index.ts
+++ b/docusaurus/website/src/theme/tracking/index.ts
@@ -16,7 +16,7 @@ const isRudderstackSetup = (config: RudderStackTrackingConfig) =>
 // Enqueue all rudderstack requests until it is loaded and consent is granted
 let rudderstack = {};
 let rudderQueue = [];
-const rudderMethods = ["page", "track", "identify", "reset"];
+const rudderMethods = ['page', 'track', 'identify', 'reset'];
 for (const method of rudderMethods) {
   rudderstack[method] = (...args) => {
     rudderQueue.push([method].concat(args));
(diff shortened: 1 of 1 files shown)


docusaurus/website/src/theme/tracking/index.ts
```diff
@@ -61,12 +61,13 @@ export function trackPage() {
...
   if (rudderQueue) {
     const { href, pathname } = location;
+    const url = Boolean(document.querySelector("link[rel='canonical']"))
+      ? document.querySelector("link[rel='canonical']").getAttribute('href')
+      : href;
     const properties = {
-      url:
-        document.querySelector("link[rel='canonical']").getAttribute("href") ||
-        href,
+      url: url,
```
tolzhabayev, line 68: nit: shouldn't it just be `url,` ?

docusaurus/website/src/theme/tracking/index.ts
```diff
@@ -16,7 +16,7 @@ const isRudderstackSetup = (config: RudderStackTrackingConfig) =>
 // Enqueue all rudderstack requests until it is loaded and consent is granted
 let rudderstack = {};
 let rudderQueue = [];
-const rudderMethods = ["page", "track", "identify", "reset"];
+const rudderMethods = ['page', 'track', 'identify', 'reset'];
```
tolzhabayev, line 19: is there a prettier config or something we should collectively use to have the same styling? otherwise it feels like we will have changes that are just styling back and forth depending on who commited them / how their IDE is setup.

tolzhabayev (approved):

docusaurus/website/src/theme/tracking/index.ts
jackw: There is a prettierrc config file in this repo. The reason for this addition in this PR is due to added a git hook recently to make sure files are fixed on commit.

docusaurus/website/src/theme/tracking/index.ts
jackw: It sure can be. I've updated it in https://github.com/grafana/plugin-tools/pull/489/commits/d1d2515521456406b1c1eefeb620133ac6d78185.

josmperez (approved):

Issue #473, opened 2023-10-17
Consider customizing DocCardLists (category indexes)


josmperez:
Background: When designing the website, we decided not to use the Docusaurus DocCardList (category index) feature because it seemed to add nothing of value. However, when creating redirects from the old Hugo site this created the problem that many Hugo index pages had nothing to link to. Subsequently, I created DocCardLists in a few categories where we were most in need of a hook to hang a redirect onto.

This issue captures two things to consider going forward
- Do we add DocCardLists to the rest of the category pages where they don't exist? And subcategory pages where there's more than one subcategory? This would have the advantage of creating a **consistent** user experience.
- Whether we do or don't expand our use of DocCardLists, do we develop custom code logic so that the default text that appears in the cards is of custom length? This would avoid the problem that the first sentences of the topics are all truncated, which proides a less than ideal user experience. 

Google "docusaurus doccardlist custom" for the latest discussion of how Docusaurus may be extended to allow custom category index pages.

sympatheticmoose:
I think the DocCardLists were a great interim solution to satisfy the requirements on redirects. Longer term I would opt for something simpler like the following, but also displaying the metadata description. 
![image](https://github.com/grafana/plugin-tools/assets/19860021/f9b3c08a-c7e3-424e-b253-80c970f3abe2)

The DocCards take up so much space for such low utility, the docs images look like they've failed to load and the folder images are not in keeping with the rest of our look and feel. 

If we have a solution which generates a list of the folder's contents displaying the title and description, I think that would suffice. As part of adding these though, I would like us to consider the additional content we would put on this index/introductory page.

josmperez:
This issue has been mainly fixed by the PR which added DocCardLinks, and[ this pending PR](https://github.com/grafana/plugin-tools/pull/526) which implemented the fix. However, we need to implement creating links based on directories as well as files.

josmperez:
Closing. To track related issue in https://github.com/grafana/plugin-tools/pull/526.

PR #493, opened 2023-10-25
Website: Add a DocLinkList component

jackw:
Minimises the space taken up by "index" pages and prevents truncation of descriptions.

## before
Screenshot 2023-10-25 at 14 57 26

## after
Screenshot 2023-10-25 at 14 56 01

Related #473

### DIFF
--- a/docusaurus/website/src/components/DocLinkList/DocLinkList.tsx
+++ b/docusaurus/website/src/components/DocLinkList/DocLinkList.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import clsx from 'clsx';
+import { useCurrentSidebarCategory, filterDocCardListItems } from '@docusaurus/theme-common';
+import Link from '@docusaurus/Link';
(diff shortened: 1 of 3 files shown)


sympatheticmoose:
@jackw so to use this we'd just replace `<DocCardList />` with `<DocLinkList />` for the current indexes?

sympatheticmoose (approved):
I like 😄 but definitely lets wait for Joe's input. 

I also wonder whether we should start documenting things like this either in https://github.com/grafana/plugin-tools/blob/main/CONTRIBUTING.md#contribute-documentation or https://github.com/grafana/plugin-tools/blob/main/docusaurus/website/README.md

jackw:
> @jackw so to use this we'd just replace `<DocCardList />` with `<DocLinkList />` for the current indexes?

Exactly and as it's also added to the `theme/MDXcomponents` file there is no need to import it in the current indexes. 👍

josmperez (approved):
Love it!

PR #507, opened 2023-10-30
Create Plugin: Replace 'master' with 'main' on  ref

joanlopez:

While starting to write a new plugin using the `@grafana/create-plugin` tool, I realized the boilerplate it generates includes a reference to the `master` branch as the source of the schema definition for the `plugin.json` file.

So, I decided to open this PR which updates the `$schema` reference in `plugin.json` templates to replace `master` with `main`, which is the correct one, I think. Otherwise, just feel free to dismiss it.

N/A

N/A


academo (approved):
Thanks for spotting this!

PR #533, opened 2023-11-09
Plugin E2E workspace prep

sunker:
The PR adds a new workspace for the `@grafana/plugin-e2e` package. It also adds a few config files. 

Part of https://github.com/grafana/grafana/issues/77484. This is the first PR of many that moves files from the [POC repo](https://github.com/sunker/plugin-e2e/pull/1) to plugin tools.


packages/plugin-e2e/docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
+version: "3.7"
```
jackw, line 1: I'm likely missing something here but I'm wondering if we should ship docker-compose files with this package. It might be to nicer to provide documentation explaining that to run e2e tests the dev needs to build their plugin, provision some dashboards/datasources and start the Grafana server (however they please).

packages/plugin-e2e/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "author": "Grafana",
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/",
+    "access": "public"
+  },
+  "scripts": {
+    "clean": "rm -rf ./dist ./compiled",
+    "build": "npm run clean && tsc -p ./tsconfig.json && rollup -c rollup.config.ts",
+    "build2": "tsc -p ./tsconfig.json && rollup -c rollup.config.ts",
```
jackw, line 25: Is this for testing something?

jackw (commented):
Awesome work splitting this out and getting the tooling in place in advance. Left a load of comments to try to bring it closer to what the other packages are using in this repo for building.

packages/plugin-e2e/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+FROM grafana/${grafana_image}:${grafana_version}
+
+# Make it as simple as possible to access the grafana instance for development purposes
+# Do NOT enable these settings in a public facing / production grafana instance
+ENV GF_AUTH_BASIC_ENABLED "false"
+# Set development mode so plugins can be loaded without the need to sign
+ENV GF_DEFAULT_APP_MODE "development"
+
+# Inject livereload script into grafana index.html
```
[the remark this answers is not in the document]
sunker: Good catch! Removed this file and updated docker-compose file.

packages/plugin-e2e/docker-compose.yaml
sunker: In the next PR, I'm going to add some playwright tests also to this repo. The tests (and the docker-compose file) won't be a part of the npm package, but they will be really useful while developing the package since you then don't have to `yarn link` a plugin. Sorry should have explained this in the PR description

packages/plugin-e2e/package.json
sunker: Oh this was just for debugging. Removed now.

packages/plugin-e2e/docker-compose.yaml
jackw: Ah ok, thanks for the explanation. Let's keep it as it sounds like it'll be very useful later on.

.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 # Build output
 packages/create-plugin/generated
 dist
+compiled
```
jackw, line 7: nit: don't think this is needed now.

packages/plugin-e2e/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@grafana/plugin-e2e",
+  "version": "0.0.1",
+  "private": true,
+  "main": "./dist/cjs/index.js",
+  "types": "./dist/cjs/index.d.ts",
```
jackw, line 6: I think these properties can now be

```suggestion
  "main": "./dist/index.js",
  "types": "./dist/index.d.ts",
```

jackw (approved):
LGTM 🚀 

Please could you double check the comment on the main and types paths in package.json before merging?

PR #537, opened 2023-11-13
Plugin E2E: Add E2E selectors

sunker:

So in the plugin e2e design doc, it was decided that the e2e selectors should live in a separate npm package that is (unlike today) detached from Grafana releases. This however, would be a big undertaking that requires a lot of coordination across many teams. For that reason, it may make sense to keep the selectors within the plugin-e2e package for some time, and then move them to its own package once we know plugin authors are starting to use plugin-e2e.

Most of the code here is copied from @grafana/e2e-selectors, and there's a bunch of things that can (and will!) be improved before plugin-e2e is released publicly the first time. 
* All selector that are not used by plugin-e2e (the ones that don't have a version associated with it) should be removed. 
* Better typing! Currently there's no type-safety between for example `Components` and `versionComponents`. This should be fairly easy to fix.  Also @mckn is looking into ways to improve this, so this will be revisited soon.
* Related to the previous - the resolver function has a bunch of `// @ts-ignore`. We'll come back to this. 
* There should be more unit tests

Why merge this when there's still things to improve? We need the selectors in all upcoming PRs. Hopefully this will make more sense then. 

Part of https://github.com/grafana/grafana/issues/78078

Fixes #

### DIFF
--- a/packages/plugin-e2e/package.json
+++ b/packages/plugin-e2e/package.json
@@ -26,6 +26,7 @@
     "lint:fix": "npm run lint -- --fix",
     "server": "docker-compose up --build",
     "typecheck": "tsc --emitDeclarationOnly false --noEmit",
+    "test": "jest",
     "playwright:test": "npx playwright test",
     "playwright:test:ui": "npx playwright test --ui"
   },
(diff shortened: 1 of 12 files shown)


mckn (approved):
LGTM! As you wrote in the description. There are a couple of things we can improve but lets to that in small steps in future PRs.

PR #539, opened 2023-11-14
Website: Update prism syntax highlighter theme

jackw:
The current syntax highlighting is a little lacking compared to the offerings that come with the prism-react-renderer lib. Rather than create and maintain our own I think it better if we repurpose an existing one, changing the background colour so it better fits in with our website. I've gone for a darker background colour here as the current one is the same colour as the info boxes.

**json**
 
Before:
Screenshot 2023-11-14 at 12 00 16 

After:
Screenshot 2023-11-14 at 12 00 26

---

**golang**

Before:
Screenshot 2023-11-14 at 12 04 12

After:

Screenshot 2023-11-14 at 12 04 24

---

**typescript**

Before:
Screenshot 2023-11-14 at 12 04 34

After:
Screenshot 2023-11-14 at 12 04 41

---

**tsx**

Before:
Screenshot 2023-11-14 at 12 05 23

After:

Screenshot 2023-11-14 at 12 05 29

Fixes #


leventebalogh (approved):
Nice 🚀

tolzhabayev:
@jackw merge?

PR #545, opened 2023-11-16
Mention new alerting extension point PR in docs

sd2k:
If/when https://github.com/grafana/grafana/pull/77900 is merged there will be a new extension point which plugin developers can register against. This PR adds the ID of the extension point to the `register-an-extension` docs.


sympatheticmoose (approved):
LGTM - thanks!

PR #546, opened 2023-11-16
Plugin E2E: Make it possible to test data source config page

sunker:

Fixes #


packages/plugin-e2e/.env.example
```diff
@@ -0,0 +1 @@
+GOOGLE_JWT_FILE=
```
sunker, line 1: secret added to the repo. can also be found in plugin-provisioning repo if you want to try this locally (or provide your own ofc).

.github/workflows/playwright.yml
```diff
@@ -10,11 +10,14 @@ jobs:
...
       fail-fast: false
       matrix:
-        GRAFANA_VERSION: ['latest', '9.5.5']
+        GRAFANA_VERSION: ['latest', '10.0.5', '9.5.5', '9.2.5']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
 
+      - name: Copy auth file
+        run: mkdir -p packages/plugin-e2e/playwright/.auth/ && cp .github/user.json packages/plugin-e2e/playwright/.auth/user.json
```
sunker, line 19: I wish playwright could create this file if it doesn't already exist, but doesn't seem like it's possible.

packages/plugin-e2e/src/fixtures/commands/createDataSource.ts
```diff
@@ -0,0 +1,50 @@
...
+  const text = await createDsReq.text();
+  const status = await createDsReq.status();
+  if (status === 200) {
+    console.log('Data source created: ', name);
+    return createDsReq.json().then((r) => r.datasource);
+  }
+
+  expect.soft(createDsReq.ok(), `Failed to create data source: ${text}`).toBeTruthy();
+  return existingDataSource.json();
+};
```
sunker, line 42: The code here is pretty rough and probably doesn't cover all scenarios...will probably come back to this in upcoming PRs.

packages/plugin-e2e/src/models/GrafanaPage.ts
```diff
@@ -0,0 +1,47 @@
...
+   * @param selector the data-testid or aria-label of the element
+   * @param root optional root locator to search within. If no locator is provided, the page will be used
+   */
+  getByTestIdOrAriaLabel(selector: string, root?: Locator): Locator {
+    if (selector.startsWith('data-testid')) {
+      return (root || this.ctx.page).getByTestId(selector);
+    }
+
+    return (root || this.ctx.page).locator(`[aria-label="${selector}"]`);
+  }
```
mckn, line 23: I do see the reason for adding this but it would be nice to push the user to use either test id or aria label. But given the structure we have in grafana core I guess it would be non-trivial to get in place.

mckn (approved):
LGTM!

packages/plugin-e2e/src/models/GrafanaPage.ts
sunker: Yes! However, lately steps have been taken to migrate all selectors from aria-label to data-testid. But even if we'd be able to get to that point reasonably soon in grafana core, this package would still need to function with much older versions of Grafana so we're gonna have to keep this abstraction for some time. =/

PR #549, opened 2023-11-18
Templates: import DataQuery from `@grafana/schema` instead of `@grafana/data`

### LINKED ISSUE #389: datasources use deprecated DataQuery from grafana/data
New plugins show deprecated warnings in `src/types.ts` with instruction to migrate to `@grafana/schema`

ludovicm67:

The import of `DataQuery` from `@grafana/data` is marked as deprecated.
This PR imports it from `@grafana/schema` instead, which is the suggested solution.
This removes a warning from the code editor.

Fixes #389 

### DIFF
--- a/packages/create-plugin/templates/datasource/src/types.ts
+++ b/packages/create-plugin/templates/datasource/src/types.ts
@@ -1,4 +1,5 @@
-import { DataQuery, DataSourceJsonData } from '@grafana/data';
+import { DataSourceJsonData } from '@grafana/data';
+import { DataQuery } from '@grafana/schema';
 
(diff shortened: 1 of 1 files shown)


academo (changes requested):
Hi @ludovicm67 and thanks for your contribution! the changes to the types file look correct but not all plugin types have `@grafana/schema` as a dependency. It is best if you define that package as a dependency for datasource types too.

ludovicm67:
Thank you @academo for the review!
I added a commit to make sure it gets included for each plugin type, which can be useful.

academo (approved):

PR #550, opened 2023-11-20
Update index.md

sympatheticmoose:
The guidance in the readme doesn't actually help much and we should look at improving. Given we already link to the readme further up in this page I would rather remove this sentence as it sets an expectation on content which isnt present

Fixes #


josmperez (approved):
LGTM.

PR #551, opened 2023-11-21
Plugin E2E: Make it possible to test query editor in panel edit page

sunker:
This PR adds fixtures, models and expect matchers that enable testing a data source query editor in the panel edit page. 

A few things to note: 
* Added two playwright tests files that tests the google sheets data source. One of them are doing full E2E tests, i.e requests are hitting downstream APIs. The other file is only testing the frontend by mocking the Grafana backend. 
* I've made the GH action matrix multi-dimensional so that tests are not only targeting multiple versions of Grafana, but also multiple versions of the Google Sheets data source.

Sorry for the large amount of files changed. Wanted the PR to include all the pieces that is needed to test a query editor so one can review this as a full feature. Let me know if you want me to try and split the PR. 

Part of  https://github.com/grafana/grafana/issues/78078

Fixes #


packages/plugin-e2e/playwright.config.ts
```diff
@@ -38,14 +38,23 @@ export default defineConfig<PluginOptions>({
       name: 'authenticate',
       testMatch: [/.*auth\.setup\.ts/],
     },
-    // 2. Run all tests in parallel with Chrome.
+    // 2. Create a datasource that can be used across tests (should use provsiioning DS instead, but currently not working in CI)
+    {
```
sunker, line 42: This is not ideal. I'd much rather have the data source being provisioned instead. For internal plugins, this will be straightforward. They can just clone the plugin-provisioning repo and mount it as a volume when starting up the grafana container. It's trickier for community plugins...me and Brain talked about adding a `provisioning` folder to the root of the plugin, and then mount this folder in the Grafana docker container. Possibly we could create a GH action that also expands secrets defined in the YAML. Anyway, this will be something to tackle in a future PR.

packages/plugin-e2e/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "outDir": "./dist",
-    "declaration": true
+    "declaration": true,
+    "stripInternal": true
```
sunker, line 5: Would like some types to be internal so added this option. Maybe this should go in the base config? Thoughts @jackw?

packages/plugin-e2e/src/models/TimeRange.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(ctx: PluginTestCtx, expect: Expect<any>) {
+    super(ctx, expect);
+  }
+
+  async set({ from, to, zone }: TimeRangeArgs) {
+    try {
+      await this.ctx.page
+        .getByLabel(this.ctx.selectors.components.PanelEditor.General.content)
+        .locator(`selector=${this.ctx.selectors.components.TimePicker.openButton}`)
+        .click({ force: true, timeout: 2000 });
```
leventebalogh, line 15: Why do we need the `2000ms` timeout here?

packages/plugin-e2e/src/models/TimeRange.ts
```diff
@@ -0,0 +1,47 @@
...
+    } catch (e) {
+      // seems like in older versions of Grafana the time picker markup is rendered twice
+      await this.ctx.page.locator('[aria-controls="TimePickerContent"]').last().click();
```
leventebalogh, line 18: Could it be that the error is caused by something else?

packages/plugin-e2e/tests/datasource/configEditor.spec.ts
```diff
@@ -11,6 +11,6 @@ test('valid credentials should return a 200 status code', async ({ createDataSou
   const configPage = await createDataSourceConfigPage({ type: 'grafana-googlesheets-datasource' });
   await page.getByText('Google JWT File', { exact: true }).click();
   await page.getByTestId('Paste JWT button').click();
-  await page.getByTestId('Configuration text area').fill(process.env.GOOGLE_JWT_FILE!);
+  await page.getByTestId('Configuration text area').fill(process.env.GOOGLE_JWT_FILE!.replace(/'/g, ''));
```
leventebalogh, line 14: Just out of curiosity, why do we need to remove the `'` single quotes here?

packages/plugin-e2e/tests/datasource/queryEditor.integration.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { expect, test } from '../../src';
+import { ProvisionFile } from '../../src/types';
+
+export type RedshiftDatasourceConfig = {
+  name: string;
+};
```
leventebalogh, line 6: I guess we can get rid of the `RedshiftDatasourceConfig`?

packages/plugin-e2e/tests/datasource/queryEditor.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  await page.keyboard.insertText('1h');
+  await page.keyboard.press('Enter');
+
+  const queryReq = panelEditPage.waitForQueryDataRequest((request) =>
+    (request.postData() ?? '').includes('"cacheDurationSeconds":3600')
+  );
+
+  await panelEditPage.refreshPanel();
+  await expect(await queryReq).toBeTruthy();
+});
```
leventebalogh, line 38: What's the "semantic" difference between the tests in `queryEditor.integration.spec.ts` and `queryEditor.spec.ts`? In other words, how do we suggest to organise tests?

packages/plugin-e2e/tsconfig.json
leventebalogh: Just out of curiosity - how do we decide which types should be private and which ones should be public? I sometimes find it useful during development to have access to as many types as possible when working with a package, while on the other hand it can tie our hands if we expose too much, as we can hardly change what a lot of people depend on.

leventebalogh (approved):
Nice work, I really like the explanatory comments, they helped a lot with reviewing. 👏  Generally looks good to me, left a few comments.

packages/plugin-e2e/src/api.ts
```diff
@@ -31,6 +56,15 @@ export type PluginFixture = {
    */
   createDataSourceConfigPage: (args: CreateDataSourcePageArgs) => Promise<DataSourceConfigPage>;
 
+  /**
+   * Fixture command that creates a data source via the Grafana API.
+   *
+   * If you have tests that depend on the the existance of a data source,
+   * you may use this command in a setup project. Read more about setup projects
+   * here: https://playwright.dev/docs/auth#basic-shared-account-in-all-tests
```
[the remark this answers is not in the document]
sunker: Yes 100%! Actually this comment was made before I knew that it's possible to use env variables in provisioning files. That kind of makes this comment obsolete. I'll change it. But we should definitely have examples of using provisioning with (and without) secrets in plugin-examples eventually.

packages/plugin-e2e/src/models/TimeRange.ts
sunker: Good question...this is mostly copied from plugin/e2e in grafana core. I'll remove it and we'll see. If timeout is not provided explicitly like in this case, timeout from playwright config will be used. It defaults to 5000ms if I recall correctly. Will changed this!

packages/plugin-e2e/src/models/TimeRange.ts
sunker: Unlikely, but maybe. Anyway, I don't like this so I'll add a `todo` for finding a better solution for this. Have a bunch of todos throughout the code...most of them involves changes in Grafana ui so would like to collect a few before I get to them.

packages/plugin-e2e/tests/datasource/configEditor.spec.ts
sunker: This variable is also used [here](https://github.com/grafana/plugin-tools/blob/query-editor/packages/plugin-e2e/provisioning/datasources/google-sheets-datasource-jwt.yaml#L15) in the provisioning file for google sheets. Seems like the jwt file needs to be wrapped in single quotes for provisioning to work, so that's how the GH secret is defined. That however doesn't work when creating a new sheets ds using the UI, so therefore I'm removing the single quotes.

packages/plugin-e2e/tests/datasource/queryEditor.spec.ts
sunker: It's a good question! 

`queryEditor.spec.ts` _just_ tests the query editor. I.e checking the right endpoints are being called when a certain button is clicked and so on. I want to test the query editor in isolation, so any backend endpoints are mocked. 

`queryEditor.integration.spec.ts` is calling the downstream sheets API and fetches data from a readonly spreadsheet. 

Ultimately, it will be up to the plugin author to decide what kind of tests they want to have. Both may be useful. Hitting real downstream APIs can be very useful, but it may come with security concerns and also cause flakiness.

packages/plugin-e2e/tsconfig.json
sunker: Good question. Think I just don't want to overload consumer with irrelevant types etc. But may be better to save this for later.

sunker:
Thanks for good feedback @leventebalogh!

PR #552, opened 2023-11-21
Run CI pipeline on pull requests too

academo:

Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
     branches:
       - main
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,53 @@
+name: Node CI
+on:
+  push:
+    branches:
+      - main
+  pull_request_target:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    needs: [test, generate-plugins]
+    if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci') && github.actor != 'dependabot[bot]'"
+    name: Release packages
+    env:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -1,13 +1,18 @@
 name: Node CI
```
academo, line 1: I removed all the mentions about tokens here.

andresmgot (approved):
Changes makes sense from the GH action perspective :+1: Maybe wait for @jackw to review the release content.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
+name: Node CI
+on:
+  push:
+    branches:
+      - main
+  pull_request_target:
```
jackw, line 6: Are we sure we want to allow npm releases from forks? Doesn't `pull_request_target` allow an external contributor to release a canary package with any code they like in it. 🤔

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci --no-audit
+
+      - name: Build
+        run: npm run build
```
jackw, line 46: If I fork this repo can I change this npm script in a way that will allow me to steal `process.env.NPM_TOKEN`?

.github/workflows/test.yml
jackw: `secrets.GRAFANA_ACCESS_POLICY_TOKEN`, and `secrets.GRAFANA_API_KEY` appear to still exist in this file. Is that an issue?

.github/workflows/release.yml
academo: Good point. I think it is just simpler to leave pull_request here

.github/workflows/test.yml
academo: I removed all the code that used those tokens. Can you point me where they are mentioned?

academo:
GRAFANA_API_KEY

PR #554, opened 2023-11-22
Typings for selectors

mckn:

Fixes #


packages/plugin-e2e/src/e2e-selectors/type-testing/types.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+export interface SelectorOptions {}
+
+// We could define this to be something else. Just continuing on the
+// idea I had before. Not important for the overall structure.
+export type SelectorFunction<Type extends SelectorOptions = SelectorOptions> = (options?: Type) => string;
+
+export type VersionedSelector = {
+  [Property: string]: SelectorFunction;
```
sunker, line 13: Should be `[Property: string]: SelectorFunction | string;` so we support string selectors too.

packages/plugin-e2e/src/e2e-selectors/type-testing/guards.ts
```diff
@@ -0,0 +1,14 @@
+import { VersionedSelector, VersionedSelectorGroup } from './types';
+import { keysOf } from './utils';
+
+export function isVersionedSelectors(value: unknown): value is VersionedSelector {
+  const notSelectorFunction = keysOf(value).find((key) => typeof value[key] !== 'function');
+  return !Boolean(notSelectorFunction);
+}
```
sunker, line 7: maybe something along those lines?
```typescript
return keysOf(value).every((key) => typeof value[key] === 'function')
```

sunker: but hmm..that doesn't take strings into account. 

return keysOf(value).every((key) => typeof value[key] === 'function') || keysOf(value).every((key) => typeof value[key] === 'string')

packages/plugin-e2e/src/e2e-selectors/type-testing/resolver.ts
```diff
@@ -0,0 +1,73 @@
...
+      return selectors;
+    }
+
+    if (isVersionedSelectorsGroup(value)) {
+      //@ts-ignore - not sure how to fix this
+      selectors[key] = toSelectors(value);
+      return selectors;
+    }
+
+    return selectors;
```
sunker, line 43: I guess this doesn't take nested selectors into consideration? For example 
```typescript
  datasource: {
    resource: {
      '9.4.4': '/api/datasources/uid/*/resources',
      [MIN_GRAFANA_VERSION]: '/api/datasources/*/resources',
    },
  },
```

sunker (commented):
This looks really promising. Great work here! Did not test it manually yet, but if I understand correctly we need to also add support for nested selectors and also it should be possible to define selectors of type string (now it's only functions). But shouldn't be a problem, right?

PR #557, opened 2023-11-25
Add new grafana/user/profile/tab plugin extension point to docs

joeyorlando:
See https://github.com/grafana/grafana/pull/77863


joeyorlando:
Not sure if we want to add an example?

ashharrison90 (approved):
given nothing else has an example, i think we're fine without? 🤔 

lgtm 👍

Issue #543, opened 2023-11-16
Move release github workflow code to a dedicated github action


academo:
this will make it possible to release updates to the build action. We don't currently distribute updates to the workflows and updating them via create-plugin might break flows because often people modify the workflows for their own needs.

academo:
Related PR in the plugin-actions repository https://github.com/grafana/plugin-actions/pull/8

Issue #541, opened 2023-11-15
Move release github action into a shared github grafana action


academo:
Instead of providing a full yaml file with the steps. we can provide a grafana action to use like this:

```
      - name: Build plugin
        uses: grafana/plugin-actions/build-plugin@v1
```

jackw:
@academo we can close this now, right?

academo:
yes. completed in https://github.com/grafana/plugin-actions/pull/8

PR #569, opened 2023-12-01
Create Plugin: Allow users to adopt CI/CD during update

asimpson:

Currently the GitHub workflow files are only generated during the `generate` command. Running `update` doesn't do anything with workflows and for good reason since users who run `update` _probably_ have CI/CD configured already. However, I ran into a situation recently where a datasource would have benefited from being able to "adopt" the workflow files.

#541 and #543 will potentially make this less destructive and we could explore just printing the necessary `yml` config to `stdout` at that point since the line count of the config will be much smaller.

With this change users can now run `npx @grafana/create-plugin@latest cicd` to explicitly adopt the Github Actions workflows.

### DIFF
--- a/docusaurus/docs/create-a-plugin/develop-a-plugin/set-up-github-workflows.md
+++ b/docusaurus/docs/create-a-plugin/develop-a-plugin/set-up-github-workflows.md
@@ -16,6 +16,13 @@ sidebar_position: 4
 
 Automate your development process to minimize errors and make it faster and more cost-efficient. The `create-plugin` tool helps you to configure your GitHub actions workflows to help automate your development process.
 
+:::tip
+
+Want to add the `create-plugin` Github workflows to your existing plugin? Run `npx @grafana/create-plugin@latest cicd` in an existing plugin to generate the `ci.yml` and `release.yml`
+workflows.
+
+:::
+
 ## The CI workflow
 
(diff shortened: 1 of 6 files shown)


academo (commented):
Code-wise this is correct but I'd like to hear input from @sympatheticmoose on this one. 

There's a risk that people would just press "Y" because we already have so many questions

Note that people will have a git project and reverting these changes might be trivial in case of mistake, so it is not a big problem to solve if you pressed the wrong key.

asimpson:
@academo @davemoore is correct, we settled on something like: `npx @grafana/create-plugin@latest cicd`. I can update this PR today hopefully with this change. 

I do think we could be more aggressive in the `update` command because of git like you said, but also because if someone is running `update` there's a built-in assumption (to me) that things will change :man_shrugging:

asimpson:
@sunker so sorry for the delay in getting back to this. I added docs and refactored things.

The new question since I created the original draft of this PR is should we be doing it this way given [the new action](https://github.com/grafana/plugin-actions/blob/main/build-plugin/README.md)? Will #742 impact how this works?

sunker (commented):
Sorry @asimpson I missed this. Changes looks good to me, but would like someone else from the team to have a look at it too.

sympatheticmoose:
@asimpson thanks again for this - theres some upcoming changes with generation which will cause issues with this - leave with us for now, but please reach out if this is blocking you

PR #578, opened 2023-12-11
Adding codeowners

tolzhabayev:
A lot of dependabot dependency updates do not get auto assigned to anyone and get stale. This adds codeowners file to avoid that.


sunker (approved):
LGTM

PR #579, opened 2023-12-12
Docs: Mention new command palette extension point and add example

sd2k:
This extension point was added in https://github.com/grafana/grafana/pull/78098.


sunker (commented):
Thanks for adding this @sd2k! 

Further down in this doc there's a few examples on how to configure different extensions. Would you be able to add an example also for this new extension point?

sunker (approved):
Awesome, thanks!

Issue #576, opened 2023-12-11
Plugin E2E: Tasklist


sunker:
### Which areas does this feature request relate to

- [ ] Create Plugin
- [ ] Sign Plugin
- [x] Plugin E2E
- [ ] Documentation

### Problem

This epic contains a list of all the tasks that need to be completed for the plugin e2e package MVP. 

For a detailed description of the problem, see https://github.com/grafana/grafana/issues/77484.  

```[tasklist]
### Tasks
- [x] [Add new workspace](https://github.com/grafana/plugin-tools/pull/533)
- [x] [Add e2e selectors](https://github.com/grafana/plugin-tools/pull/537)
- [x] [Add first fixtures](https://github.com/grafana/plugin-tools/pull/538)
- [x] [Enable testing Config editor page](https://github.com/grafana/plugin-tools/pull/546)
- [x] [Enable testing the Query editor](https://github.com/grafana/plugin-tools/pull/551)
- [x] [Add provisioning file](https://github.com/grafana/plugin-tools/pull/558)
- [x] [Enable testing the Variable editor](https://github.com/grafana/plugin-tools/pull/553)
- [x] [Enable testing the Annotation editor](https://github.com/grafana/plugin-tools/pull/577)
- [x] [Enable testing Explore](https://github.com/grafana/plugin-tools/pull/581)
- [x] [Add GH action that resolves target versions for Grafana](https://github.com/grafana/plugin-tools/pull/584)
- [x] [Replace magic strings with selectors](https://github.com/grafana/plugin-tools/pull/597) (https://github.com/grafana/plugin-tools/pull/581#discussion_r1425004723)
- [ ] https://github.com/grafana/plugin-tools/pull/605
- [x] [Harmonise model instantiations and navigation](https://github.com/grafana/plugin-tools/pull/603)
- [ ] https://github.com/grafana/plugin-tools/pull/651
- [ ] https://github.com/grafana/plugin-tools/issues/724
- [ ] https://github.com/grafana/plugin-tools/pull/610
- [ ] https://github.com/grafana/plugin-tools/pull/612
- [ ] https://github.com/grafana/plugin-tools/pull/616
- [ ] https://github.com/grafana/plugin-tools/pull/634
- [ ] https://github.com/grafana/plugin-tools/pull/645
- [x] Refine publish to GCS step (https://github.com/grafana/plugin-tools/pull/645#pullrequestreview-1838771761)
[...]

PR #581, opened 2023-12-12
Plugin E2E: Enable testing the query editor in Explore

sunker:
This PR adds fixture and model that enable testing a data source query editor and query runner in Explore. Also adding a few tests that showcases how Explore tests can be written using this feature.

Part of https://github.com/grafana/plugin-tools/issues/576

Fixes #

### DIFF
--- a/packages/plugin-e2e/src/models/ExplorePage.ts
+++ b/packages/plugin-e2e/src/models/ExplorePage.ts
@@ -28,12 +28,6 @@ export class ExplorePage extends GrafanaPage {
 
   async runQuery(options?: RequestOptions) {
     const components = this.ctx.selectors.components;
-    const runButtonTextContent = await this.getByTestIdOrAriaLabel(components.RefreshPicker.runButtonV2).textContent();
-    // if there is a query running, wait for it to finish before starting a new one
-    if (runButtonTextContent === 'Cancel') {
-      await this.ctx.page.waitForResponse((resp) => resp.url().includes('/api/ds/query'), options);
-    }
-
     const responsePromise = this.ctx.page.waitForResponse((resp) => resp.url().includes('/api/ds/query'), options);
     try {
       await this.getByTestIdOrAriaLabel(components.RefreshPicker.runButtonV2).click({
--- a/packages/plugin-e2e/src/models/ExplorePage.ts
+++ b/packages/plugin-e2e/src/models/ExplorePage.ts
@@ -1,0 +1,49 @@
+import { Expect, Locator } from '@playwright/test';
+import { NavigateOptions, PluginTestCtx, RequestOptions } from '../types';
+import { DataSourcePicker } from './DataSourcePicker';
+import { GrafanaPage } from './GrafanaPage';
+import { TimeRange } from './TimeRange';
+
+export class ExplorePage extends GrafanaPage {
+  datasource: DataSourcePicker;
+  timeRange: any;
+  constructor(ctx: PluginTestCtx, expect: Expect<any>) {
+    super(ctx, expect);
+    this.datasource = new DataSourcePicker(ctx, expect);
+    this.timeRange = new TimeRange(ctx, this.expect);
+  }
+
+  async goto(options?: NavigateOptions) {
+    await super.navigate(this.ctx.selectors.pages.Explore.url, options);
+  }
+
+  async getQueryEditorRow(refId: string): Promise<Locator> {
+    //TODO: add new selector and use it in grafana/ui
+    const locator = this.ctx.page.locator('[aria-label="Query editor row"]').filter({
+      has: this.ctx.page.locator(`[aria-label="Query editor row title ${refId}"]`),
+    });
+    await this.expect(locator).toBeVisible();
+    return locator;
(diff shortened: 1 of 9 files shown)


packages/plugin-e2e/src/models/GrafanaPage.ts
```diff
@@ -9,6 +9,16 @@ import { PluginTestCtx } from '../types';
 export abstract class GrafanaPage {
   constructor(public readonly ctx: PluginTestCtx, protected readonly expect: Expect<any>) {}
 
+  protected async navigate(url: string, options?: NavigateOptions) {
```
sunker, line 12: Planning to use this method across all models, but will do that in a separate PR.

packages/plugin-e2e/src/models/TimeRange.ts
```diff
@@ -27,18 +27,5 @@ export class TimeRange extends GrafanaPage {
     await toField.clear();
     await toField.fill(to);
     await this.getByTestIdOrAriaLabel(this.ctx.selectors.components.TimePicker.applyTimeRange).click();
-
-    await this.expect
```
sunker, line 31: Don't think we need these soft expects.

packages/plugin-e2e/src/models/ExplorePage.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+    await this.expect(locator).toBeVisible();
+    return locator;
+  }
+
+  async runQuery(options?: RequestOptions) {
+    const components = this.ctx.selectors.components;
+    const runButtonTextContent = await this.getByTestIdOrAriaLabel(components.RefreshPicker.runButtonV2).textContent();
+    // if there is a query running, wait for it to finish before starting a new one
+    if (runButtonTextContent === 'Cancel') {
```
academo, line 33: is there a way with playwright to check if there's a request to this url ongoing? or use the waitForResponse to return if no request is going?

packages/plugin-e2e/src/models/ExplorePage.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    const responsePromise = this.ctx.page.waitForResponse((resp) => resp.url().includes('/api/ds/query'), options);
+    try {
+      await this.getByTestIdOrAriaLabel(components.RefreshPicker.runButtonV2).click({
+        timeout: 1000,
+      });
+    } catch (_) {
+      // handle the case when the run button is hidden behind the "Show more items" button
+      await this.getByTestIdOrAriaLabel(components.PageToolbar.item('Show more items')).click();
```
academo, line 44: not sure how you are handling these magic strings so far, but would it be better to put all these in a file where we can re-use them and have more information related to it.

e.g. `PageToolbar.ShowMoreItemsButton`

This is because the string alone doesn't give context of where is this string supposed to be in the UI and less about being able to re-use or easily update.

packages/plugin-e2e/src/models/ExplorePage.ts
sunker: Yeah I agree this is not ideal. I'll look into ways of improving this tomorrow.

packages/plugin-e2e/src/models/ExplorePage.ts
sunker: Yeah this is a huge anti-pattern that I've come across a few times in the grafana/e2e-selectors package. The whole point of this package is to provide a safe way to access these selectors, but with these functions with magic strings we're kind of losing the whole purpose of it. 

I think rewriting all of these selectors in grafana/e2e-selectors will be a too big undertaking for now, but I'll make note in the tasklist to at least change the ones that we use in grafana-e2e (I know they're being used at least two more times).

academo: I am not suggesting we move them to e2e/selectors or a whole different package. something as simple as defining the variable at the beginning of the file to avoid this to turn into a  [magic string](https://en.wikipedia.org/wiki/Magic_string)

packages/plugin-e2e/src/models/ExplorePage.ts
sunker: Networking is an area where Playwright falls a bit short IMO. To achieve something like this, we'd need to setup one event listener that keeps track of new requests, and a second one that keeps track of responses. Then you'd need your own mechanism that correlates these two. 

It seems like the sheets data source is triggering a `onRunQuery` after any change in the query editor. For that reason, there may already be an ongoing request when run query method is executed in the test. This is rather unconventional, so after closer consideration I don't think this should be handled in the lib, but in the data source instead. I'm removing this code.

packages/plugin-e2e/src/models/ExplorePage.ts
sunker: Done

packages/plugin-e2e/src/models/ExplorePage.ts
academo: sounds good! the current approach will have to work I guess :)

academo (approved):

PR #584, opened 2023-12-13
Plugin-E2E: Add version resolver action

sunker:

This PR adds an Github Action action that has the purpose of resolving what Grafana versions to target in the E2E tests. The output of the action is a json array of Grafana minor versions that can be used to specify a version matrix in the next workflow job. See [this](https://github.com/grafana/plugin-tools/blob/plugin-e2e/version-resolver-script/.github/workflows/playwright.yml#L8-L26) example. 

Currently, the actions support two modes:

**plugin-grafana-dependency (default)**
The will return all minor versions of Grafana since the version that was specified as grafanaDependency in the plugin.json. (does not work for this repo since it's not a plugin). 

**version-support-policy**
This will resolve versions according to the min support policy defined in [this](https://docs.google.com/document/d/17xrR-uEooyZsTE7HoJZnImTC6bCkLqRLiFw8tA-DhO4/edit#heading=h.5xmdrfuzlmt9) doc. TLDR; all minors for the current major version and the last minor of the previous major version.

If a plugin author requires more flexibility, they can ofc define their own matrix. 

At some point (when it's more stable) we probably want to move this action to its own repo and tag a release. Maybe also rewrite using Typescript. 

Fixes #
Part of https://github.com/grafana/plugin-tools/issues/576

For now I'm just creating a bundle using the following command and committing it. 
```node
npx ncc build ./index.js -o output
```
When the action is moved to its own repo, we can look at better alternatives for this.


.github/workflows/playwright.yml
```diff
@@ -1,18 +1,31 @@
 name: E2E tests - Playwright
 on:
   pull_request:
-    paths:
-      - 'packages/plugin-e2e/**'
+    # paths:
+    #   - 'packages/plugin-e2e/**'
```
sunker, line 5: Need to uncomment before merging

jackw:
>At some point (when it's more stable) we probably want to move this action to its own repo and tag a release. Maybe also rewrite using Typescript.

I was going to suggest this could be moved to the [plugin-actions](https://github.com/grafana/plugin-actions) repo.

sunker:
> > At some point (when it's more stable) we probably want to move this action to its own repo and tag a release. Maybe also rewrite using Typescript.
> 
> I was going to suggest this could be moved to the [plugin-actions](https://github.com/grafana/plugin-actions) repo.

Ah sweet, didn't know we had that repo. In that case I'm closing this PR. Will open a new in plugin-actions shortly.

PR #618, opened 2023-12-19
Revert "Chore: Bump Auto npm packages to ^11.x.x"

jackw:
Reverts grafana/plugin-tools#596

We're seeing some oddness in the tagging and changelogs for plugin-e2e. One possible reason is this PR that bumped all the auto packages so reverting to see if it has any effect.


sunker (approved):
LGTM

.github/dependabot.yml
```diff
@@ -3,32 +3,26 @@
 version: 2
 updates:
   # Check for updates to GitHub Actions every month
-  - package-ecosystem: 'github-actions'
-    directory: '/'
+  - package-ecosystem: "github-actions"
```
tolzhabayev, line 6: https://github.com/grafana/plugin-tools/blob/main/.prettierrc#L7 singleQuote or double quote?

jackw: It should really be single quotes. I'm hoping to revert this revert soon enough. 😄

PR #623, opened 2023-12-20
Plugin-E2E: Allow provisioningRootDir to be overridden

sunker:

When using the `readProvision` fixture, provisioned files are read from a rootdir. For external plugins, this will most likely always be `<projectRoot>/provisioning`. However in some projects, you may want to specify an alternative directory for your provisioning files. For example, in core Grafana, you may want to use `devenv`. 

This PR makes it possible to override the default directory for provisioning files. It can be done globally or on a project level in the playwright config, or in an individual test. 

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/plugin-e2e@0.3.0-canary.623.aedff75.0
  # or 
  yarn add @grafana/plugin-e2e@0.3.0-canary.623.aedff75.0
  ```
</details>


packages/plugin-e2e/src/api.ts
```diff
@@ -15,7 +15,19 @@ import { grafanaE2ESelectorEngine } from './selectorEngine';
 import { ExplorePage } from './models/ExplorePage';
 
 export type PluginOptions = {
-  selectorRegistration: void;
```
academo, line 18: is removing this from the type intentional?

sunker: Good catch - should have added a GH comment about this. Yes, it was intentionally removed. It's not an actual fixture option, so it doesn't belong there.

academo (approved):

PR #629, opened 2023-12-29
Docs: Add new streaming data source tutorial

### LINKED ISSUE #428: Epic: Update and enhance streaming data source tutorial
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/grafana/issues/71766 
- [ ] https://github.com/grafana/grafana/issues/67566
- [ ] https://github.com/grafana/grafana/issues/68741
```

josmperez:
Add new streaming data source tutorial. Draws on source content by Hugo at https://dev.to/oshirohugo/grafana-streaming-backend-datasource-plugin-1ob7

Fixes: https://github.com/grafana/plugin-tools/issues/428


docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,286 @@ keywords:
...
-   ```ts
-   const observables = options.targets.map((target) => {
-     return new Observable<DataQueryResponse>((subscriber) => {
-       // ...
-     });
-   });
+```
+  query(request: DataQueryRequest<MyQuery>): Observable<DataQueryResponse> {
+    const observables = request.targets.map((target, index) => {
+      const query = defaults(target, DEFAULT_QUERY);
```
sunker, line 102: I think we'd want devs to use the `getDefaultQuery` method rather than setting defaults like this, so you can remove this line.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,286 @@ keywords:
...
+
+1. Add the data source. Since we are running in our `docker-compose` environment, we don't need to install it, and it will be directly available for usage. Go to **Connections > Data sources**, using the left menu as shown in the following image:
+
+1. A new page opens, and then click **Add data source**. Grafana opens another page where you can search for the data source name that we've just created.
+
+1. Click on the data source's card, and then click **Save & test** on the next page.
+
+![Data sources - Grafana UI.](/img/data-sources-menu.png)
+
+1. If a randomized error occurs, just click **Save & test** until a success message appears:
```
sunker, line 287: What is this about?

oshirohugo: This comes from the code provided by the scaffold created by plugin tools
https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/backend/pkg/plugin/datasource.go#L96-L109

But maybe we can add an extra step in the tutorial to change it to

```go
func (d *Datasource) CheckHealth(_ context.Context, req *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
	return &backend.CheckHealthResult{
		Status:  backend.HealthStatusOk,
		Message: "Data source is working",
	}, nil
}
```
What do you think about that @josmperez ?
Or you can make it as a tip as you did with the `QueryData`

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,286 @@ keywords:
...
-     clearInterval(intervalId);
-   };
-   ```
+```
+var (
+    _ backend.CheckHealthHandler    = (*Datasource)(nil)
+    _ instancemgmt.InstanceDisposer = (*Datasource)(nil)
+    _ backend.StreamHandler         = (*Datasource)(nil)
+)
+```
```
oshirohugo, line 147: we could also use syntax highlight in the backend part:
```go
var (
    _ backend.CheckHealthHandler    = (*Datasource)(nil)
    _ instancemgmt.InstanceDisposer = (*Datasource)(nil)
    _ backend.StreamHandler         = (*Datasource)(nil)
)
```

the same for all the others piece of code for the backend

oshirohugo:
Awesome work @josmperez !

Reviewing your tutorial I've notice that I forgot to mention the changes in `src/plugin.json`. We need to add the `"streaming": true` to it. Then it should be something like:

```json
{
  ...
  "id": "grafana-streamingbackendexample-datasource",
  "metrics": true,
  "backend": true,
  ...
  "streaming": true
}
```

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: @sunker - To be clear, are you saying that the only change needed here is to remove line 102, or something else?

josmperez:
Hugo, regarding this:

https://github.com/grafana/plugin-tools/pull/629#discussion_r1443048552

I want to add this as a tip, but please clarify where the code you provided should be added.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: Interesting. With regard to syntax highlight, I have two concerns. First, do you feel that the tutorial's text is referring to specific parts of the code block which need to be called out? Second, I want to ensure consistency with other Grafana documentation, so I am researching this further.

josmperez: done.

josmperez:
I can't find any part of the below code block in this tutorial. Could you please clarify where this line should be added?

> ```json
> {
>   ...
>   "id": "grafana-streamingbackendexample-datasource",
>   "metrics": true,
>   "backend": true,
>   ...
>   "streaming": true
> }
> ```

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
sunker: yes - but also change line 101 from 
`const observables = request.targets.map((target, index) => {`
to
`const observables = request.targets.map((query, index) => {`

oshirohugo:
> Hugo, regarding this:
> 
> [#629 (comment)](https://github.com/grafana/plugin-tools/pull/629#discussion_r1443048552)
> 
> I want to add this as a tip, but please clarify where the code you provided should be added.

This code should be added to `pkg/plugin/datasource.go`. It should replace the `CheckHealth` method that already exists there and it was created by the scaffold.

oshirohugo:
> I can't find any part of the below code block in this tutorial. Could you please clarify where this line should be added?

Your totally right, so far there are no mentions to `src/plugin.json` in the tutorial. It's a file that should also be present when the plugin is create by `create-plugin` tool. For example:
https://github.com/grafana/grafana-plugin-examples/blob/main/examples/panel-flot/src/plugin.json

And in this tutorial the full `src/plugin.json` will be something like:
https://github.com/oshirohugo/datasource-streaming-backend/blob/main/src/plugin.json

where we added the property `"streaming": true`.

We need to add a new part to the tutorial mentioning that we need of this modification. Maybe it can be a new step before the `Step 4: Build the plugin`.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: done

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: @oshirohugo - I have added this code, but you don't state where the code should be added. Can you check the change and suggest the point of addition?

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,286 @@ keywords:
...
+
+## Step 1: Scaffold the plugin
+
+It is possible to create all the plugin code from scratch, as long as you follow the Grafana plugins specification and implement all the necessary interfaces. However, the easiest way to create a plugin is using the `@grafana/create-plugin` tool that we provide.
+
+1. Go to the location where you want your plugin directory to be created and run:
+
+```
+npx @grafana/create-plugin@latest
+```
```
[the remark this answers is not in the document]
josmperez: I have added syntax highlight elsewhere, but Docusauurus documentation points to this: https://github.com/FormidableLabs/prism-react-renderer/blob/master/packages/generate-prism-languages/index.ts as the list of supported languages for syntax highlighting and bash or shell aren't on the list. I'm not sure that we can add highlight here.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,286 @@ keywords:
...
-   ```ts
-   query(options: DataQueryRequest<MyQuery>): Observable<DataQueryResponse> {
-     // ...
-   }
-   ```
+1. Replace the `QueryEditor` function in `/src/components/QueryEditor.tsx` with the following code:
 
-1. Create an `Observable` instance for each query, and then combine them all using the `merge` function from the `rxjs` package.
+```
+export function QueryEditor({ query, onChange, onRunQuery }: Props) {
```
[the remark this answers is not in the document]
josmperez: done, except for CLI commands

josmperez:
> Awesome work @josmperez !
> 
> Reviewing your tutorial I've notice that I forgot to mention the changes in `src/plugin.json`. We need to add the `"streaming": true` to it. Then it should be something like:
> 
> ```json
> {
>   ...
>   "id": "grafana-streamingbackendexample-datasource",
>   "metrics": true,
>   "backend": true,
>   ...
>   "streaming": true
> }
> ```

Where should this code be added?

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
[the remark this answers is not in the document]
oshirohugo: Thank you for checking this Joe. I didn't know shell or bash where not supported. From the list that you referenced, it seems like there is no equivalent. So I think it's ok to let it as it was before, without any language specification :+1:

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
oshirohugo: I think it's better to add this part of the code to the end of `Step 3: Set up the plugin backend`, since it should also be included also in `pkg/plugin/datasource.go`.

We can say something like "to prevent a randomize error, change the `CheckHealth` method provided by default for the following code..."

We should also state that this part has nothing to do with the streaming plugin itself. It's just to avoid the error that will be present by method provided by default.

oshirohugo:
> Where should this code be added?

I think we can create a new step after `Step 3: Set up the plugin backend` and before `Step 4: Build the plugin`, where we say something like:

Grafana needs to recognize that the new plugin is a streaming plugin when it's first loaded, to accomplish this, simply add `"streaming":true` property to your `src/plugin.json`

```json
{
  ...
  "id": "grafana-streamingbackendexample-datasource",
  "metrics": true,
  "backend": true,
  "streaming": true
  ...
}
```

It can be a new step called "Changing `plugin.json`` or `Set up the plugin as streaming` or something like that.

josmperez:
@oshirohugo - I have added the two new parts to this tutorial. Please re-review.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,321 @@ keywords:
...
-      const frame = new CircularDataFrame({
-        append: 'tail',
-        capacity: 1000,
-      });
+That means our `DataSource` will implement `backend.CheckHealthHandler`, `instancemgmt.InstanceDisposer`, and `backend.StreamHandler`. The methods necessary for the first two are already provided in the scaffold; therefore, we need to implement only the methods necessary for `backend.StreamHandler`.
 
-      frame.refId = query.refId;
-      frame.addField({ name: 'time', type: FieldType.time });
-      frame.addField({ name: 'value', type: FieldType.number });
+2. This code will be called when the user tries to subscribe to a channel. You can implement permissions checking here, but in our case we just want the user to connect successfully in each attempt; therefore, we just return a backend: `SubscribeStreamStatusOK`.
```
oshirohugo, line 150: From the text content, it seems like the `SubscribeStream` code should be added before this item

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,321 @@ keywords:
...
-  return merge(...streams);
+4. To prevent a random error, change the `CheckHealth` method provided by default for the following code:
+
+```tsx
+func (d *Datasource) SubscribeStream(context.Context, *backend.SubscribeStreamRequest) (*backend.SubscribeStreamResponse, error) {
+    return &backend.SubscribeStreamResponse{
+        Status: backend.SubscribeStreamStatusOK,
+    }, nil
+}
+```
```
oshirohugo, line 172: I think this part here is not correct. We are saying that the `CheckHealth` should be replaced, but then we are showing the code for `SubscribeStream`. I think the correct core should be the `CheckHealth`

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
```diff
@@ -15,143 +14,321 @@ keywords:
...
+      Message: "Data source is working",
+    }, nil
+  }
+```
+
+:::note
+
+If a randomized error occurs, just click **Save & test** until a success message appears:
+
+![Datasource dialog box - Grafana UI.](/img/datasource-dialog.png)
```
oshirohugo, line 318: Since we are adding the `CheckHealth` code in `Step 3: Set up the plugin backend`, we need to remove this.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: sorry, I misunderstood earlier. I've moved the new step to the bottom (#7) of step 3.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: fixed.

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: removed

docusaurus/docs/tutorials/build-a-streaming-data-source-plugin.md
josmperez: I'm not quite sure how to implement your suggestion. It sounds like you are saying perhaps that we need to insert a new step between steps 1 and 2 on "Step 3", one which would look like this:

2. Implement code that returns `SubscribeStreamStatusOK` when the user attempts to connect to the backend:

 ```tsx
func (d *Datasource) SubscribeStream(context.Context, *backend.SubscribeStreamRequest) (*backend.SubscribeStreamResponse, error) {
    return &backend.SubscribeStreamResponse{
        Status: backend.SubscribeStreamStatusOK,
    }, nil
}
```

Is that right?

oshirohugo (approved):

PR #636, opened 2024-01-03
CI: Conditionally build docs

marefr:
Uses https://github.com/dorny/paths-filter to be able to conditionally build docs and mark the check as successful if there are no docs changes and by that making it possible to make Test docs build a required check before merge.

Fixes #

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -13,14 +13,6 @@ jobs:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
       NX_BRANCH: ${{ github.event.number || github.ref_name }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build.yml
```diff
@@ -19,16 +13,30 @@ jobs:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
       NX_BRANCH: ${{ github.event.number || github.ref_name }}
     steps:
+      - name: Check for docs file changes
```
tolzhabayev, line 16: Should we maybe split the jobs? The check and then the deploy that `needs` the check

marefr: Nice one :+1:

tolzhabayev (approved):
🚀

PR #655, opened 2024-01-10
Plugin E2E: Change signature of custom matcher

sunker:

Playwright has a [`toBeOK`](https://playwright.dev/docs/api/class-apiresponseassertions#api-response-assertions-to-be-ok) matcher. Unfortunately, it only applies to the [APIResponse](https://playwright.dev/docs/api/class-apiresponse) type, so I have defined a custom matcher with the same name that applies also to `Response` type (which is useful when executing grafana queries). I'd like our custom matchers to align with Playwright matchers if possible, so this PR changes the signature of the custom `toBeOK` matcher to accept a `Response` instead of `Promise<Response>`. 

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/plugin-e2e@0.9.0-canary.655.ce61438.0
  # or 
  yarn add @grafana/plugin-e2e@0.9.0-canary.655.ce61438.0
  ```
</details>

### DIFF
--- a/packages/plugin-e2e/src/index.ts
+++ b/packages/plugin-e2e/src/index.ts
@@ -18,7 +18,7 @@ declare global {
       /**
        * Await the response of a Playwright request and asserts the response was successful (status in the range 200-299).
        */
-      toBeOK(this: Matchers<unknown, Promise<Response>>): R;
+      toBeOK(this: Matchers<unknown, Response>): R;
 
       /**
        * Asserts that preview text elements are displayed on the Variable Edit Page. You should make sure any variable queries are completed before calling this matcher.
--- a/packages/plugin-e2e/src/index.ts
+++ b/packages/plugin-e2e/src/index.ts
@@ -16,7 +16,7 @@ declare global {
       [t]: T;
 
       /**
-       * Await the response of a Playwright request and asserts the response was successful (status in the range 200-299).
+       * Asserts that a Playwright response was successful (status in the range 200-299).
        */
       toBeOK(this: Matchers<unknown, Response>): R;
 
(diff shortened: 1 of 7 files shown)


academo (commented):
Correct me if I am wrong because I might be speaking out of my lack of knowledge about playwright, but, do we want to start customizing the playwright API out of convenience ?

If that's the case it may lead us to a path where things get harder and harder to migrate when newer versions come out because we started to create wrappers around things as we did with cypress and the e2e package.

Do we really need this change for something more than a bit of ease of typing or is it really contributing to make plugins more "testable"?

sunker:
> Correct me if I am wrong because I might be speaking out of my lack of knowledge about playwright, but, do we want to start customizing the playwright API out of convenience ?

In this case we're not customizing the Playwright API - we're just adding a new plugin-e2e matcher for a type that doesn't already have a corresponding matcher. 

> If that's the case it may lead us to a path where things get harder and harder to migrate when newer versions come out because we started to create wrappers around things as we did with cypress and the e2e package.

> Do we really need this change for something more than a bit of ease of typing or is it really contributing to make plugins more "testable"?

This is a very valid concern, but again we're not extending Playwright apis here. I think it's important that we only add custom matchers to plugin-e2e when there's an obvious need for it. In this case, it's a little bit about "ease of typing" but mostly for readability. 

Using our custom matcher: 
```typescript
await expect(await panelEditPage.refreshPanel()).toBeOK();
```

Using Playwright matcher: 
```typescript
const response = await panelEditPage.refreshPanel();
await expect(response.ok()).toBeTruthy();
```

I find the first one more expressive. One also have to keep in mind that this type of assertion (verifying that the response after a certain user interaction is valid or not) is going to be used in basically every single test we write.

sunker:
Have raised a [feature request](https://github.com/microsoft/playwright/issues/29231) in the Playwright repo. Closing this pr for now.

Issue #654, opened 2024-01-10
Feat: Migrate packages to ESM


jackw:
### Which areas does this feature request relate to

- [X] Create Plugin
- [X] Sign Plugin
- [ ] Documentation

### Problem

We have dependencies in our projects in this repo which we cannot update to because they are ESM only packages and current build tooling doesn't support them.

### Solution

Migrate sign-plugin and create-plugin to build as ESM packages. We only support node LTS and as they are cli scripts that are run via NPX I don't consider this will be a breaking change if we drop CommonJS support.

```[tasklist]
### Tasks
- [x] Migrate create-plugin to ESM only package https://github.com/grafana/plugin-tools/pull/658
- [x] Migrate sign-plugin to ESM only package https://github.com/grafana/plugin-tools/pull/662
- [x] Clean up mono-repo post ESM only package changes https://github.com/grafana/plugin-tools/pull/694
```

### Alternatives

Don't update dependencies?

### Are you interested in contributing the solution?

- [X] Yes
- [ ] No

jackw:
Closing as all related work is now merged to main.

PR #658, opened 2024-01-11
Build: Migrate create-plugin to ESM

jackw:
This PR updates the create-plugin package to build only as an es module. The following changes have been made with reasoning on approaches:

- Updates Typescript config to output ESM. These changes are in line with [tsconfig/bases/node20](https://github.com/tsconfig/bases/blob/main/bases/node20.json) suggested config. Having found these configs I'm wondering if we should update the entire repo to use them over writing our own.
- Updates all import paths inline with Typescripts requirements by adding `.js` extension to imports. Note that auto barrel file resolution (`from './commands/'` resolved to `from './commands/index.js'`) no longer work with ESM. You have to reference a file. VsCode autocompletion seems to work out the box. I attempted to use a bundler ESbuild to prevent the need to change all these imports but [hit a wall](https://github.com/evanw/esbuild/issues/622#issuecomment-752915822).
- Update all imports to use `import` over `require`. I'm not sure why `utils.console.ts` used require but changing them led to complications with TS types and needing to patch the marked-terminal package and set overrides for `marked` nested dependency.
- Replaces Jest with Vitest. Turns out Jest support for ESM is quite poor. Mocking in particular is complicated due to [esm static import analysis](https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm) and needing to use `unstable_mockModule` which I didn't particularly like the sound of. Vitest migration was quite painless, only the mocks needed some attention.
- Updates _some_ dependencies to their ESM compatible versions. Other dependency updates will be done in future PRs.

Related #654

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@3.0.0-canary.658.52bb056.0
  # or 
  yarn add @grafana/create-plugin@3.0.0-canary.658.52bb056.0
  ```
</details>


packages/sign-plugin/tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "compilerOptions": {
+    "lib": ["es2023", "dom"],
+    "module": "node16",
+    "target": "es2022",
+    "moduleResolution": "node16",
```
jackw, line 6: Until setting `"type": "module"` in sign-plugin package.json these changes will continue to output CJS. They are necessary to prevent failures in typechecking.

mckn (approved):
LGTM! Interesting since I tried to do the same thing with Levitate and manage to get everything working except the jest tests. But then the solution might be to migrate over to vitest.

jackw:
>But then the solution might be to migrate over to vitest.

Probably will make life much easier. Admittedly this is a small project but the Vitest migration was surprisingly straightforward. It was only mocks that needed any attention. Jest and ESM just don't very work well together. 😞

PR #660, opened 2024-01-11
Use `custom.ini` file provided to configure Grafana docker container

samjewell:
We provide a `custom.ini` file here:
https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/scenes-app/provisioning/custom.ini But then we completely ignore it! So it's currently a red-herring.

This change means we no longer ignore it, so app developers have an obvious place they can go to eg. enable feature-flags.

Refer to [Grafana configuration docs](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#config-file-locations) and [Docker configuration docs](https://grafana.com/docs/grafana/latest/setup-grafana/configure-docker/)

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@2.12.0-canary.660.dd63773.0
  # or 
  yarn add @grafana/create-plugin@2.12.0-canary.660.dd63773.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/app/provisioning/custom.ini
+++ b/packages/create-plugin/templates/app/provisioning/custom.ini
@@ -1,0 +1,5 @@
+app_mode = development
+
+[feature_toggles]
+
+
--- a/packages/create-plugin/templates/app/provisioning/custom.ini
+++ b/packages/create-plugin/templates/app/provisioning/custom.ini
@@ -1,5 +1,0 @@
-app_mode = development
-
-[feature_toggles]
-
-
--- a/packages/create-plugin/templates/common/.config/Dockerfile
+++ b/packages/create-plugin/templates/common/.config/Dockerfile
@@ -10,8 +10,6 @@ ENV GF_AUTH_ANONYMOUS_ENABLED "true"
 ENV GF_AUTH_BASIC_ENABLED "false"
 # Set development mode so plugins can be loaded without the need to sign
 ENV GF_DEFAULT_APP_MODE "development"
-# Use the config file in this repo
-ENV GF_PATHS_CONFIG "/etc/grafana/provisioning/custom.ini"
 
 # Inject livereload script into grafana index.html
 USER root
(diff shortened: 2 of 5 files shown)


mckn (commented):
Thanks for the PR! I'm not sure we should scaffold a `custom.ini` at all. Or, I'm a bit split about the whole thing. Until now we have tried to get developers to manage configurations via environment variables in the docker-compose file. All values that can be set in the `custom.ini` can also be set via environment variables.

The benefit is that we keep the environment configuration at the same place as where we defined the rest of the environment in docker-compose. If we scaffold custom.ini for plugin developers we basically have two ways of managing the environment which makes things harder to debug.

I think this is fine for plugin developers to add on their own. But I think it makes more sense to have one recommended way that we scaffold.

What does @jackw @academo @sunker @leventebalogh think?

jackw:
I'm not sure I see the need for these `config.ini` files either. As you've explained in the PR description @samjewell they are a red herring and create confusion. Like @mckn I too think it would be easier to manage if we don't have these config.ini files and use the docker-compose as the source of truth. Maybe we need to update our [dev env documentation](https://grafana.com/developers/plugin-tools/get-started/set-up-development-environment) explaining that docker-compose.yml is the place to set config via `GF_<config_var_name>` and maybe refer to https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#override-configuration-with-environment-variables?

samjewell:
Thanks for your thoughts @mckn and @jackw 

> I too think it would be easier to manage if we don't have these config.ini files and use the docker-compose as the source of truth

Fair enough. I'm happy to abandon this PR, or update it to remove the red-herring `config.ini`. I don't want to do more than that, as I have my own projects to work on - I was just making a drive-by PR because I hit this myself, and thought it should be fixed. WDYT? Updating the docs sounds like something worth doing, but I don't personally have time.

Thanks for explaining that these can be configured through environment-variables - TIL.

academo:
I'm with @mckn and @jackw in their opinion that it is better not to generate this file for the reasons they pointed and I'll like to add the risk of users modifying this file with special configurations for their plugins that later can not be easily replicated in other environments thinking this file will be distributed with their plugins.

samjewell:
Thanks. I've reversed the change here, and removed the `custom.ini` file from the scenes-app template instead. Could you take another look please?

leventebalogh (approved):
I agree with the points above, thanks for the change @samjewell! 👏 
LGTM 👍

academo (approved):

PR #661, opened 2024-01-11
Docs: Add doc for building plugin locally

### LINKED ISSUE #628: Research missing "develop locally" topic
Staff reports "develop Grafana locally" topic can't be found. Last source is here:

https://github.com/grafana/grafana/blob/a9b9b96c61700f072040b981c701b70a1c581588/docs/sources/developers/plugins/development-with-local-grafana.md#run-your-backend-plugin-with-a-debugger

This topic was not moved in the re-org and may need to be re-added under "Get started". Investigate whether content is still relevant.

josmperez:
Restores a file that did not get moved during the transition to the Developer Portal. Fixes https://github.com/grafana/plugin-tools/issues/628

### DIFF
--- a/docusaurus/docs/get-started/develop-locally.md
+++ b/docusaurus/docs/get-started/develop-locally.md
@@ -42,9 +42,11 @@ To clone and run Grafana locally:
 
 Another option is to run Grafana in a container with `docker compose` (or `docker-compose`, which is a version of this command deprecated by Docker). To do so, create the `docker compose` file in your plugin directory:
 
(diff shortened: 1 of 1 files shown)


sympatheticmoose (changes requested):
This was intentionally removed as docker is the preferred method. Will add more comments when possible, but for now please don't merge

josmperez:
Ah, thanks. That's the information I needed. I couldn't remember why we didn't move it over so I wanted to use this PR to discuss.

PR #662, opened 2024-01-12
Build: Migrate sign-plugin to ESM

jackw:

This PR updates the sign-plugin package to build only as an es module. The following changes have been made with reasoning on approaches:

- Updates Typescript config to output ESM. These changes are in line with [tsconfig/bases/node20](https://github.com/tsconfig/bases/blob/main/bases/node20.json) suggested config. Having found these configs I'm wondering if we should update the entire repo to use them over writing our own.
- Updates all import paths inline with Typescripts requirements by adding `.js` extension to imports. Note that auto barrel file resolution (`from './commands/'` resolved to `from './commands/index.js'`) no longer work with ESM. You have to reference a file. VsCode autocompletion seems to work out the box. I attempted to use a bundler ESbuild to prevent the need to change all these imports but [hit a wall](https://github.com/evanw/esbuild/issues/622#issuecomment-752915822). 
- Replaces Jest with Vitest. Turns out Jest support for ESM is quite poor. Mocking in particular is complicated due to [esm static import analysis](https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm) and needing to use `unstable_mockModule` which I didn't particularly like the sound of. Vitest migration was quite painless, only the mocks needed some attention.

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/sign-plugin@3.0.0-canary.662.8ab45d5.0
  # or 
  yarn add @grafana/sign-plugin@3.0.0-canary.662.8ab45d5.0
  ```
</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28549,7 +28549,8 @@
         "sign-plugin": "dist/bin/run.js"
       },
       "devDependencies": {
-        "@types/minimist": "^1.2.2"
+        "@types/minimist": "^1.2.2",
+        "vitest": "^1.1.3"
       },
       "engines": {
         "node": ">=20"
--- a/packages/sign-plugin/jest.config.js
+++ b/packages/sign-plugin/jest.config.js
@@ -1,3 +1,0 @@
-const sharedConfig = require('../../jest.config.base');
-
-module.exports = sharedConfig;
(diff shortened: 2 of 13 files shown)


leventebalogh (approved):
Nice work, LGTM 🚀

academo:
> I'm wondering if we should update the entire repo to use them over writing our own.

let's use the standard rules unless there's something we really don't want from them. Less work to maintain them.

academo (approved):

PR #669, opened 2024-01-15
Create Plugin: run prettier on .config after update

### LINKED ISSUE #656: Run prettier on `.config` after update.
Currently some of our templates generate code that is not correctly formatted after is generated. 

We should:

* detect if the project still uses prettier
* attempt to run prettier in the updated `.config` folder files.

after the update command is called.

oshirohugo:

To correctly format the code added after we update `.config` content we need to run prettier after it.

Fixes https://github.com/grafana/plugin-tools/issues/656

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@3.6.0-canary.669.68b0597.0
  # or 
  yarn add @grafana/create-plugin@3.6.0-canary.669.68b0597.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/commands/generate.command.ts
+++ b/packages/create-plugin/src/commands/generate.command.ts
@@ -45,7 +45,7 @@ export const generate = async (argv: minimist.ParsedArgs) => {
   if (answers.hasBackend) {
     await execPostScaffoldFunction(updateGoSdkAndModules, exportPath);
   }
-  await execPostScaffoldFunction(prettifyFiles, exportPath);
+  await execPostScaffoldFunction(prettifyFiles, { targetPath: exportPath });
 
   printGenerateSuccessMessage(answers);
 };
--- a/packages/create-plugin/src/commands/generate.command.ts
+++ b/packages/create-plugin/src/commands/generate.command.ts
@@ -12,7 +12,7 @@ import { getPackageManagerFromUserAgent, getPackageManagerInstallCmd } from '../
 import { getExportPath } from '../utils/utils.path.js';
 import { renderTemplateFromFile } from '../utils/utils.templates.js';
 import { getVersion } from '../utils/utils.version.js';
-import { prettifyFiles } from './generate/prettify-files.js';
+import { prettifyFiles } from '../utils/utils.prettifyFiles.js';
 import { printGenerateSuccessMessage } from './generate/print-success-message.js';
 import { promptUser } from './generate/prompt-user.js';
 import { updateGoSdkAndModules } from './generate/update-go-sdk-and-packages.js';
--- a/packages/create-plugin/src/commands/generate/prettify-files.ts
+++ b/packages/create-plugin/src/commands/generate/prettify-files.ts
@@ -4,13 +4,13 @@ import fs from 'node:fs';
 
 const exec = promisify(nodeExec);
 
-export async function prettifyFiles(exportPath: string) {
+export async function prettifyFiles(exportPath: string, packageManagerExecute = 'npx') {
   if (!fs.existsSync(exportPath)) {
     return '';
   }
 
   try {
-    let command = 'npx -y prettier@2 . --write';
+    let command = `${packageManagerExecute} -y prettier@2 . --write`;
     await exec(command, { cwd: exportPath });
   } catch (error) {
     throw new Error(
--- a/packages/create-plugin/src/commands/update.command.ts
+++ b/packages/create-plugin/src/commands/update.command.ts
@@ -10,11 +10,11 @@ import {
   updateNpmScripts,
   writePackageManagerInPackageJson,
   projectUsesPrettier,
-  runPrettierOnConfigDir,
 } from '../utils/utils.npm.js';
 import { isGitDirectory, isGitDirectoryClean } from '../utils/utils.git.js';
 import { isPluginDirectory } from '../utils/utils.plugin.js';
 import { getPackageManagerExecute, getPackageManagerWithFallback } from '../utils/utils.packageManager.js';
+import { prettifyFiles } from './generate/prettify-files.js';
 
 export const update = async (argv: minimist.ParsedArgs) => {
   try {
@@ -63,7 +63,7 @@ In case you want to proceed as is please use the ${chalk.bold('--force')} flag.)
 
       if (projectUsesPrettier()) {
         const packageManagerExecute = getPackageManagerExecute();
-        runPrettierOnConfigDir(packageManagerExecute);
+        prettifyFiles('.config', packageManagerExecute);
       }
 
       printSuccessMessage(TEXT.overrideFilesSuccess);
--- a/packages/create-plugin/src/commands/update.command.ts
+++ b/packages/create-plugin/src/commands/update.command.ts
@@ -13,7 +13,7 @@ import {
 import { isGitDirectory, isGitDirectoryClean } from '../utils/utils.git.js';
 import { isPluginDirectory } from '../utils/utils.plugin.js';
 import { getPackageManagerWithFallback } from '../utils/utils.packageManager.js';
-import { isPrettierUsed, prettifyFiles } from '../utils/utils.prettifyFiles.js';
+import { prettifyFiles } from '../utils/utils.prettifyFiles.js';
 
 export const update = async (argv: minimist.ParsedArgs) => {
   try {
@@ -60,11 +60,10 @@ In case you want to proceed as is please use the ${chalk.bold('--force')} flag.)
     if (await confirmPrompt(TEXT.updateConfigPrompt)) {
       compileTemplateFiles(UDPATE_CONFIG.filesToOverride, getTemplateData());
 
-      if (isPrettierUsed()) {
-        prettifyFiles('.config', '.');
-      }
-
       printSuccessMessage(TEXT.overrideFilesSuccess);
+
+      const prettifyMsg = await prettifyFiles({ targetPath: '.config', projectRoot: '.' });
+      printMessage(prettifyMsg);
     } else {
       printMessage(TEXT.overrideFilesAborted);
       process.exit(0);
@@ -115,6 +114,9 @@ In case you want to proceed as is please use the ${chalk.bold('--force')} flag.)
     // Guarantee that the package manager property is set in the package.json file if it is missing
     const packageManager = getPackageManagerWithFallback();
     writePackageManagerInPackageJson(packageManager);
+
+    const prettifyMsg = await prettifyFiles({ targetPath: '.config', projectRoot: '.' });
+    printMessage(prettifyMsg);
 
     // 4. Summary
     // -------------
--- a/packages/create-plugin/src/commands/update.command.ts
+++ b/packages/create-plugin/src/commands/update.command.ts
@@ -9,12 +9,11 @@ import {
   getPackageJsonUpdatesAsText,
   updateNpmScripts,
   writePackageManagerInPackageJson,
-  getPrettierVersion,
 } from '../utils/utils.npm.js';
 import { isGitDirectory, isGitDirectoryClean } from '../utils/utils.git.js';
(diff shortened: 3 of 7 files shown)


packages/create-plugin/src/utils/utils.npm.ts
```diff
@@ -236,3 +237,17 @@ function sortKeysAlphabetically(obj: Record<string, string>) {
       return acc;
     }, {});
 }
+
+export function projectUsesPrettier() {
+  const packageJson = getPackageJson();
+  return Boolean(packageJson.devDependencies?.prettier);
+}
+
+export function runPrettierOnConfigDir(packageManagerName: string) {
```
academo, line 246: We already have this https://github.com/grafana/plugin-tools/blob/2012aa1d94e7db8864c831833d40a149338b479b/packages/create-plugin/src/commands/generate/prettify-files.ts#L7

Perhaps you can use that and modify it so it uses the logic for the package manager execution?

Additionally it'd be best if you validate that we are in the correct place. i.e. `.config` actually exists.

packages/create-plugin/src/commands/generate/prettify-files.ts
```diff
@@ -4,13 +4,13 @@ import fs from 'node:fs';
```
academo, line 4: nit: move this file out of the `generate` folder since it is now used in more than the generate command.

packages/create-plugin/src/commands/generate/prettify-files.ts
```diff
@@ -4,13 +4,13 @@ import fs from 'node:fs';
 
 const exec = promisify(nodeExec);
 
-export async function prettifyFiles(exportPath: string) {
+export async function prettifyFiles(exportPath: string, packageManagerExecute = 'npx') {
```
academo, line 7: instead of accepting the packageManagerExecute, can you use the new `getPackageManagerExecute` here?

packages/create-plugin/src/commands/generate/prettify-files.ts
```diff
@@ -4,13 +4,13 @@ import fs from 'node:fs';
...
   if (!fs.existsSync(exportPath)) {
     return '';
   }
 
   try {
-    let command = 'npx -y prettier@2 . --write';
+    let command = `${packageManagerExecute} -y prettier@2 . --write`;
```
academo, line 13: the `-y` option will cause problem with other package managers than are not `npx`. They most likely fail because the option doesn't exist.

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -107,3 +107,18 @@ function getPackageManagerFromPackageJson(): PackageManager | undefined {
 
   return undefined;
 }
+
+export function getPackageManagerExecute() {
+  const packageManager = getPackageManagerWithFallback();
+
+  switch (packageManager.packageManagerName) {
+    case 'yarn':
+      return 'yarn dlx';
```
jackw, line 116: I'm afraid there is no npx equivalent for yarn classic which (to the best of my knowledge) is the only version of yarn create-plugin supports right now.

```
$ yarn dlx prettier@2 . --write
yarn run v1.22.19
error Command "dlx" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

As yarn classic puts packages on the filesystem in a very similar manner to npm you might be able to get away with using the default case (`npx`) for yarn here too.

packages/create-plugin/src/commands/generate/prettify-files.ts
jackw, line 13: Do we need to specify the package manager to use here? `npx` will always be available regardless of the packageManager the project uses and with `-y` will silently grab the version from npmjs if it's not available as a local binary already. If we want to improve this, maybe it's worth checking the package.json to understand which version should be used instead of assuming it's `2`. Maybe projectUsesPrettier could be used here to return the version else it bails if not found?

packages/create-plugin/src/utils/utils.packageManager.ts
```diff
@@ -107,3 +107,18 @@ function getPackageManagerFromPackageJson(): PackageManager | undefined {
 
   return undefined;
 }
+
+export function getPackageManagerExecute() {
```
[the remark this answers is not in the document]
oshirohugo: as you mention that `npx` will always be available with every package manager, let's remove this function and just go with `npx` in all cases.

packages/create-plugin/src/commands/generate/prettify-files.ts
jackw: >On the other hand, the same function prettifyFiles is used during create-plugin generate. In this phase, we don't have a package.json to parse and check the prettier version, so version 2 should be used.

I don't think this is correct. We should have a package.json on the filesystem before we call `prettifyFiles`. When generate is run it awaits `generateFiles` which awaits `writeFile` which writes all the files to the disk. Only after that happens do we call `prettifyFiles`.

packages/create-plugin/src/utils/utils.packageManager.ts
oshirohugo: since we are going to always use `npx` I totally removed this function

academo (approved):

packages/create-plugin/src/commands/generate.command.ts
```diff
@@ -12,7 +12,7 @@ import { getPackageManagerFromUserAgent, getPackageManagerInstallCmd } from '../
 import { getExportPath } from '../utils/utils.path.js';
 import { renderTemplateFromFile } from '../utils/utils.templates.js';
 import { getVersion } from '../utils/utils.version.js';
-import { prettifyFiles } from './generate/prettify-files.js';
+import { prettifyFiles } from '../utils/utils.prettifyFiles.js';
```
jackw, line 15: 👏

packages/create-plugin/src/utils/utils.prettifyFiles.ts
```diff
@@ -1,16 +1,19 @@
...
 
 const exec = promisify(nodeExec);
 
-export async function prettifyFiles(exportPath: string) {
+export async function prettifyFiles(exportPath: string, projectRoot = exportPath) {
   if (!fs.existsSync(exportPath)) {
     return '';
   }
 
+  const prettierVersion = getPrettierVersion() || '2';
```
jackw, line 13: nit: if we're only ever calling this when prettier is used I'm not sure there's a need for the fallback. 🤔

packages/create-plugin/src/utils/utils.prettifyFiles.ts
```diff
@@ -19,3 +22,12 @@ export async function prettifyFiles(exportPath: string) {
...
   return 'Successfully ran prettier against new plugin.';
 }
+
+export function isPrettierUsed() {
+  return Boolean(getPrettierVersion());
+}
+
+function getPrettierVersion() {
+  const packageJson = getPackageJson();
+  return packageJson.devDependencies?.prettier || packageJson.dependencies?.prettier;
```
jackw, line 32: I was slightly concerned that using a semver range with npx (e.g. `npx prettier@^2.8.7 . --write`) would cause it to fail but it turns out it works! 🥳

packages/create-plugin/src/utils/utils.packagejson.ts
```diff
@@ -10,8 +10,8 @@ export type PackageJson = {
   devDependencies: Record<string, string>;
   packageManager?: string;
 } & Record<string, any>;
-export function getPackageJson(): PackageJson {
-  return readJsonFile(path.join(process.cwd(), 'package.json'));
+export function getPackageJson(rootPath = process.cwd()): PackageJson {
```
jackw, line 13: Unless I'm blind I cannot see any usage of this additional parameter. 🤔

packages/create-plugin/src/utils/utils.prettifyFiles.ts
```diff
@@ -1,16 +1,19 @@
 import { exec as nodeExec } from 'node:child_process';
 import { promisify } from 'node:util';
 import fs from 'node:fs';
+import { getPackageJson } from './utils.packagejson.js';
 
 const exec = promisify(nodeExec);
 
-export async function prettifyFiles(exportPath: string) {
+export async function prettifyFiles(exportPath: string, projectRoot = exportPath) {
```
jackw, line 8: Where is `projectRoot` used within this function?

packages/create-plugin/src/utils/utils.packagejson.ts
oshirohugo: This is necessary, because now `getPackageJson` is being used in an additional context: the `create-plugin generate`. Therefore I need to indicate where the project root is, so the `getPackageJson` can run there. In all the other contexts, `getPackageJson` run in the same directory as the command, therefore I kept the `process.cwd` as default.

It wasn't possible to see that in the last PR because I did a mistake and forgot to pass a different `rootPath` to `getPackgeJson`. Now this is fixed. Please check the other comments.

packages/create-plugin/src/utils/utils.prettifyFiles.ts
oshirohugo: You were right, the last version was wrong and I forgot to use the `projectRoot`.

This was fixed and `projectRoot` is used in line 13

```ts
 const prettierVersion = getPrettierVersion(projectRoot);
```

This is necessary because `prettifyFiles` is called in two different ways, in plugin generate, when `exportPath` is the same as `projectRoot` and in plugin update when exportPath is `.config` and `projectRoot` is the current command location.

packages/create-plugin/src/commands/update.command.ts
```diff
@@ -20,6 +21,11 @@ export const update = async () => {
     // -----------------------------------
     if (await confirmPrompt(TEXT.updateConfigPrompt)) {
       compileTemplateFiles(UDPATE_CONFIG.filesToOverride, getTemplateData());
+
+      if (isPrettierUsed()) {
+        prettifyFiles('.config', '.');
+      }
```
leventebalogh, line 27: **Nit:** we could make the call-site simpler by moving the `if (isPrettierUsed()) { ...` check into the `prettifyFiles()` function.

leventebalogh: I think it would also make sense to use named parameters for `prettifyFiles()`, it would make it more clear by looking at it what those strings / paths are, e.g.:

```typescript
type PrettifyFilesArgs = {
  // The path where we want to prettify files, defaults to the CWD
  targetPath?: string;
  
  // In case the `targetPath` would be set to a subdirectory, e.g. `.config/`. Defaults to `targetPath`.
  projectRoot?: string;
}

export async function prettifyFiles(options: PrettifyFilesArgs) {
  const targetPath = options.targetPath ?? process.cwd();
  const projectRoot = options.projectRoot ?? targetPath;
   
```

packages/create-plugin/src/utils/utils.prettifyFiles.ts
```diff
@@ -0,0 +1,33 @@
...
+    throw new Error(
+      'There was a problem running prettier on the plugin files. Please run `npx -y prettier@2 . --write` manually in your plugin directory.'
+    );
+  }
+  return 'Successfully ran prettier against new plugin.';
+}
+
+export function isPrettierUsed() {
+  return Boolean(getPrettierVersion());
+}
```
leventebalogh, line 28: I guess `isPrettierUsed()` needs to support the `projectRoot` setting as well?

```javascript
export function isPrettierUsed(projectRoot?: string) {
  return Boolean(getPrettierVersion(projectRoot));
```

packages/create-plugin/src/utils/utils.packagejson.ts
jackw: Ok nice. Thanks for the explanation @oshirohugo ! 👍

oshirohugo:
Closing in favor of https://github.com/grafana/plugin-tools/pull/722

PR #678, opened 2024-01-19
CI: Enable NPM publishing

jackw:
NPM publishing was disabled whilst we debugged issues with strange changelogs and incorrect package versions being released. We now believe the main cause of this was due to a gh outage which prevented the latest release from being set post publish of a package. This resulted in Auto using already released PRs to calculate new versions of packages. The exact reason why it published a 1.0.0 is still a bit of a mystery but looking at [the most recent action run](https://github.com/grafana/plugin-tools/actions/runs/7572224011/job/20621710121) it looks like we're back where we should be. We're leaving the verbose flag in place for future debugging.

Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "docs:build": "nx build website",
     "lint": "nx run-many --target=lint --parallel",
     "lint:fix": "nx run-many --target=lint:fix --parallel",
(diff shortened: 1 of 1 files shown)


leventebalogh (approved):
LGTM 👍

sunker (approved):

PR #681, opened 2024-01-22
Chore: Rollback Auto to 11.0.2 due to potential publishing bug

jackw:

This PR rolls back the version of Auto to 11.0.2.

After all the debugging and thinking our issues around publishing were caused by a GH outage I noticed that the GH release for the latest published version end of last week still lacked the latest tag.  We assumed it was caused by a GH outage but I've now discovered that [Auto 11.0.3 introduced a new flag](https://github.com/intuit/auto/pull/2391) to the GH publish command but the NPM plugin doesn't set this and the flag is false by default. I'm now pretty sure [this is](https://github.com/intuit/auto/issues/2422) the cause of the issues we've had recently. 

Fixes #


leventebalogh (approved):
🚢🇮🇹

academo (approved):

PR #664, opened 2024-01-12
angular to react configuration migration guide

briangann:
**What this PR does / why we need it**:

Adds how-to guide for angular plugins migrating to react

Fixes grafana/grafana#79200

PR #682, opened 2024-01-22
create-plugin: adds archive option for CI workflow

briangann:

Adds archiving of build so it can be downloaded/tested locally or through other automation methods.

The build will be signed if the key is present, otherwise it is just packaged as-is.

A "working" example can be found here:

https://github.com/briangann/grafana-gauge-panel/actions/runs/7617166961

image

doc referenced in #664

### DIFF
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      GRAFANA_ACCESS_POLICY_TOKEN: \${{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
+      GRAFANA_ACCESS_POLICY_TOKEN: $\{{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
     steps:
       - uses: actions/checkout@v3
 {{#if_eq packageManagerName "pnpm"}}
@@ -99,7 +99,7 @@ jobs:
 
       - name: Sign plugin
         run: {{ packageManagerName }} run sign
-        if: \${{ env.GRAFANA_ACCESS_POLICY_TOKEN != '' }}
+        if: $\{{ env.GRAFANA_ACCESS_POLICY_TOKEN != '' }}
 
       - name: Get plugin metadata
         id: metadata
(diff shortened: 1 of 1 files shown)


packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
```diff
@@ -94,3 +96,33 @@ jobs:
...
+        if: \${{ env.GRAFANA_ACCESS_POLICY_TOKEN != '' }}
+
+      - name: Get plugin metadata
+        id: metadata
+        run: |
+          sudo apt-get install jq
+
+          export GRAFANA_PLUGIN_ID=$(cat dist/plugin.json | jq -r .id)
+          export GRAFANA_PLUGIN_VERSION=$(cat dist/plugin.json | jq -r .info.version)
+          export GRAFANA_PLUGIN_ARTIFACT=${GRAFANA_PLUGIN_ID}-${GRAFANA_PLUGIN_VERSION}.zip
```
leventebalogh, line 111: Nit: maybe we don't need to export these variables?

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
```diff
@@ -94,3 +96,33 @@ jobs:
...
+        id: package-plugin
+        run: |
+          mv dist \${{ steps.metadata.outputs.plugin-id }}
+          zip \${{ steps.metadata.outputs.archive }} \${{ steps.metadata.outputs.plugin-id }} -r
+
+      - name: Archive Build
+        uses: actions/upload-artifact@v3
+        with:
+          name: \${{ steps.metadata.outputs.plugin-id }}-\${{ steps.metadata.outputs.plugin-version }}
+          path: \${{ steps.metadata.outputs.plugin-id }}
```
leventebalogh, line 127: Nit: is the escaping of the dollar signs necessary here?

leventebalogh (approved):
LGTM 👍

academo (approved):

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
briangann: double checked this and fixed the escapes - they needed to be after the $ not before, the generated file now matches what is expected

PR #688, opened 2024-01-23
Plugin E2E: Enable assertion on panel data 

sunker:

Plugin E2E currently provides two ways to assert that a data source query works as expected: 
* using the `toBeOK` matcher to check that the data query response was ok
 e.g `await expect(explorePage.runQuery()).toBeOK();`
* using the `hasPanelError` matcher to checks weather the panel has an error icon

As @yesoreyeram pointed out, we need a way to check what data is being displayed in the panel too. That way, one can assert that the entire data flow of a plugin (data source plugin in particular) works as expected. 

Asserting on the data that is being displayed in a panel is tricky as the markup is a moving target. For that reason, it's important that we somehow incapsulate panel data extraction within plugin-e2e so we have a chance to fix the parsing whenever the panel markup changes. I was initially considering writing a plugin-e2e custom matcher to cater for this. However, that may lead to a scenario where we need to provide a whole bunch of different matchers to allow users to find assert on single value, multiple values, substring searches etc. It would be better if we could find a way to use the built-in Playwright matchers. 

In this PR, the `Panel` model exposes two methods - `getFieldNames` and `getData`. getFieldNames returns a locator that resolves to element(s) that point to data frame field name, and getData returns a locator that resolves to element(s) that point to data frame values. Since they both return a Playwright locator, it can be used together with any of built-in [playwright matchers](https://playwright.dev/docs/test-assertions#auto-retrying-assertions). Currently, this only works for the table panel. That is likely the only panel you'd want to use in an E2E test anyway, but if another panel is being used one can always toggle `View table` (in panel edit), and then assertion should work.  

e.g

```ts
test('table panel data assertions', async ({ page, selectors, grafanaVersion, request, readProvision }) => {
    const dashboard = await readProvision<Dashboard>({ filePath: 'dashboards/redshift.json' });
    const panelEditPage = new PanelEditPage({ page, selectors, grafanaVersion, request }, { dashboard, id: '3' });
    await panelEditPage.goto();
    await panelEditPage.setVisualization('Table');
    await expect(panelEditPage.panel.getData()).toContainText(['staging']);
    await expect(panelEditPage.panel.getFieldNames()).toContainText(['time', 'temperature']);
});
```

You'll find more examples of how data assertions can be done in provisioned dashboard, explore page etc in the `dataAssertion.spec.ts` file. 

<details>
  <br />

  ```bash
  npm install @grafana/plugin-e2e@0.13.0-canary.688.5df6275.0
  # or 
  yarn add @grafana/plugin-e2e@0.13.0-canary.688.5df6275.0
  ```
</details>

### DIFF
--- a/packages/plugin-e2e/src/models/ExplorePage.ts
+++ b/packages/plugin-e2e/src/models/ExplorePage.ts
@@ -17,32 +17,50 @@ export class ExplorePage extends GrafanaPage {
   datasource: DataSourcePicker;
   timeRange: any;
   timeSeriesPanel: Panel;
-  logsPanel: Panel;
   tablePanel: Panel;
 
   constructor(ctx: PluginTestCtx) {
     super(ctx);
     this.datasource = new DataSourcePicker(ctx);
     this.timeRange = new TimeRange(ctx);
-    this.timeSeriesPanel = new Panel(this.ctx, () =>
-      this.getPanelLocators(TIME_SERIES_PANEL_SELECTOR_SUFFIX, TIME_SERIES_PANEL_TEXT)
-    );
-    this.tablePanel = new Panel(this.ctx, () => this.getPanelLocators(TABLE_PANEL_SELECTOR_SUFFIX, TABLE_PANEL_TEXT));
-    this.logsPanel = new Panel(this.ctx, () => this.getPanelLocators(LOGS_PANEL_SELECTOR_SUFFIX, LOGS_PANEL_TEXT));
+    this.timeSeriesPanel = new Panel(this.ctx, () => this.getTimeSeriesPanelLocator());
+    this.tablePanel = new Panel(this.ctx, () => this.getTablePanelLocator());
   }
 
-  private getPanelLocators(suffix: string, text: string) {
-    const page = this.ctx.page;
-    let locator = this.getByTestIdOrAriaLabel(this.ctx.selectors.components.Panels.Panel.title(suffix), {
-      startsWith: true,
-    });
+  private getTablePanelLocator() {
+    let locator = this.ctx.page.getByTestId('Panel');
+    if (semver.lt(this.ctx.grafanaVersion, '9.3.0')) {
+      // seems there is no way to get the locator for the panels in Grafana versions < 9.3.0
+      throw new Error(`Can't locate panels on Explore page for Grafana versions < 9.3.0. Please skip this test.`);
+    } else if (semver.lt(this.ctx.grafanaVersion, '10.0.0')) {
+      // having to use these selectors is unfortunate, but the Explore page did not use data-testid on the panels before Grafana 10.
+      locator = this.ctx.page.getByRole('button', { name: TABLE_PANEL_TEXT }).locator('..');
+    } else if (semver.lt(this.ctx.grafanaVersion, '10.4.0')) {
+      locator = this.getByTestIdOrAriaLabel(
+        this.ctx.selectors.components.Panels.Panel.title(TABLE_PANEL_SELECTOR_SUFFIX),
+        {
+          startsWith: true,
+        }
+      );
+    }
+
+    return locator;
+  }
+
+  private getTimeSeriesPanelLocator() {
+    let locator = this.getByTestIdOrAriaLabel(
+      this.ctx.selectors.components.Panels.Panel.title(TIME_SERIES_PANEL_SELECTOR_SUFFIX),
+      {
+        startsWith: true,
+      }
+    );
 
     if (semver.lt(this.ctx.grafanaVersion, '9.3.0')) {
       // seems there is no way to get the locator for the panels in Grafana versions < 9.3.0
       throw new Error(`Can't locate panels on Explore page for Grafana versions < 9.3.0. Please skip this test.`);
     } else if (semver.lt(this.ctx.grafanaVersion, '10.0.0')) {
       // having to use these selectors is unfortunate, but the Explore page did not use data-testid on the panels before Grafana 10.
-      locator = page.getByRole('button', { name: text }).locator('..');
+      locator = this.ctx.page.getByRole('button', { name: TIME_SERIES_PANEL_TEXT }).locator('..');
     }
 
     return locator;
--- a/packages/plugin-e2e/src/models/ExplorePage.ts
+++ b/packages/plugin-e2e/src/models/ExplorePage.ts
@@ -23,10 +23,11 @@ export class ExplorePage extends GrafanaPage {
     super(ctx);
     this.datasource = new DataSourcePicker(ctx);
     this.timeRange = new TimeRange(ctx);
-    this.timeSeriesPanel = new Panel(this.ctx, () =>
+    this.timeSeriesPanel = new Panel(
+      this.ctx,
       this.getPanelLocators(TIME_SERIES_PANEL_SELECTOR_SUFFIX, TIME_SERIES_PANEL_TEXT)
     );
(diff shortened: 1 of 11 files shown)


packages/plugin-e2e/src/models/panels/visualizations.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+export class Table extends VisualizationDataExtractor {
+  constructor() {
+    super('Table');
+  }
+
+  getLabels(): Locator {
+    return this.panel.locator('[role="columnheader"]');
```
sunker, line 50: Of course we'd need to add proper e2e selectors to all visualizations in Grafana an use them instead of week magic string like this (the same goes for all following extractors).

yesoreyeram (commented):
Feels like a good start. We also need to consider multiple cases and make it more generic. In general, instead of implementing individual methods for panels such as `getLabels`/`getValues`, I would like to see `getData` kind of abstraction which should return array of data frames kind of data used by the viz to render the data. We could use `toggle table view` and extract data to make it more consistent for the viz. 

If I am datasource developer, I will least bother about how different visualisation render my data. Only care about if the visualization getting correct/consistent data from the query.

sunker:
Thanks for feedback @yesoreyeram! 

Yes I like the idea of toggling to table view for all panels. Will make it way less complicated. 

> I would like to see getData kind of abstraction which should return array of data frames kind of data used by the viz to render the data. 

How would we be able to get the data frames passed to a panel from within an E2E test? Checking the Query inspector panel data would be slow and brittle. 

I'm not sure you would want to assert on content of the data frames in an E2E test. Even though the format is to be considered stable, it's still moving target so you'd need to change your tests whenever a change is introduced. I think we're better of looking at what's being rendered in the table panel. _If_ the rendering of table data is changed, we would catch that in an E2E test in grafana core and provide a fix in the plugin-e2e parser. 

What about if we skip the whole vizualization concept, and instead expose two methods from the panel model - `getData` and `getFieldNames` (or getColumnNames). These methods would first toggle to Table view and then return a Playwright locator which will give plugin authors access to the variety  of Playwright matchers that already exist (there's a lot of them). Then if there's a need to assert on the shape of the data frames, they can use the json in the response returned by for example `panelEditPage.refreshPanel()`. It won't cover the scenario where data is being further processed in the frontend, but probably good enough in most cases.

packages/plugin-e2e/src/models/panels/visualizations.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  init(ctx: PluginTestCtx, panel: Locator) {
+    this.ctx = ctx;
+    this.panel = panel;
+  }
+
+  getValues(): Locator {
+    throw new Error(`getValues locator method was not implemented for the ${this.visualizationType} visualization`);
+  }
+  getLabels(): Locator {
```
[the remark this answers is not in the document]
sunker: good point!

packages/plugin-e2e/src/models/panels/visualizations.ts
```diff
@@ -0,0 +1,96 @@
...
+export class Table extends VisualizationDataExtractor {
+  constructor() {
+    super('Table');
+  }
+
+  getLabels(): Locator {
+    return this.panel.locator('[role="columnheader"]');
+  }
+
+  getValues(): Locator {
```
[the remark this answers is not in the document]
sunker: hmm that's the wide format right? doesn't the table panel give you the option to change series from a dropdown then? maybe plugin-e2e could provide apis for changing that.

packages/plugin-e2e/src/models/panels/visualizations.ts
```diff
@@ -0,0 +1,96 @@
...
+      return new Gauge();
+    case 'Bar gauge':
+      return new BarGauge();
+    default:
+      console.log(`No visualization data extractor found for visualization type: ${visualizationType}`);
+      return new VisualizationDataExtractor(visualizationType);
+  }
+};
+
+export class VisualizationDataExtractor {
```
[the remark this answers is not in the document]
sunker: Yeah you may be right. I think I mostly wanted to support different visualizations because it would partly enable E2E testing that a provisioned databoard is working as expect. 

```ts
const dashboard = await readProvision<Dashboard>({ filePath: 'dashboards/redshift.json' });
 const dashboardPage = new DashboardPage({ page, selectors, grafanaVersion, request }, { uid: dashboard.uid });
 await dashboardPage.goto();
 // this is a Gauge panel
 const panel = await dashboardPage.getPanelByTitle('Redshift');
 expect(await panel.getValues()). toHaveValue('test')
```

But ofc this would work also by switching to table view.

sunker:
Awesome @yesoreyeram - I take it we're on the same page and have a path forward. I'll try to update the PR tomorrow. 👍

packages/plugin-e2e/src/models/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { Page } from '@playwright/test';
+
+export const radioButtonSetChecked = async (
```
sunker, line 3: Planning to expose some kind of UI helpers through the base page, but will do that in a different [pr](https://github.com/grafana/plugin-tools/pull/703).

packages/plugin-e2e/src/models/Panel.ts
```diff
@@ -10,6 +10,26 @@ export class Panel extends GrafanaPage {
...
+  /**
+   * Returns a locator that resolves element(s) that contain the value(s) that are currently displayed in the panel.
+   *
+   * Can be used to assert the values displayed in the panel visualization. e.g
+   * await expect(panelEditPage.panel.getData()).toContainText(['1', '4', '14']);
+   */
+  getData(): Locator {
+    return this.locator().locator('[role="cell"]');
+  }
+
```
sunker, line 32: FYI I'm working on adding proper e2e selectors for the table panel in grafana core. 

Even though people will probably want to use the Table panel in most cases, it won't cover the scenario when you want to assert on panel (that is not table panel) data in a dashboard, as you can't toggle to `Table view` from there.  Not high priority, but we could easily add parsing functionality to some of the most common panel types. E.g

```ts
getData(): Locator {
    switch (this.visualizationType) {
      case 'table':
        return this.locator().locator('[role="cell"]');
      case 'logs':
        return this.locator().locator(selectors.components.Panels.Panel.Logs.logRow);
      case 'gauge':
        return this.locator().locator(selectors.components.Panels.Panel.Gauge.value);
      default:
        throw new Error(`No data parser implemented for this visualization type: ${this.visualizationType}`);
    }
  }
```

packages/plugin-e2e/src/models/ExplorePage.ts
```diff
@@ -17,32 +17,50 @@ export class ExplorePage extends GrafanaPage {
...
+    this.timeSeriesPanel = new Panel(this.ctx, () => this.getTimeSeriesPanelLocator());
+    this.tablePanel = new Panel(this.ctx, () => this.getTablePanelLocator());
   }
 
-  private getPanelLocators(suffix: string, text: string) {
-    const page = this.ctx.page;
-    let locator = this.getByTestIdOrAriaLabel(this.ctx.selectors.components.Panels.Panel.title(suffix), {
-      startsWith: true,
-    });
+  private getTablePanelLocator() {
```
sunker, line 30: locating table panel and time series panel in explore is a mess. for some reason, selectors seem to change in every version of Grafana. need to look into this at some point and investigate how this can be improved.

yesoreyeram (approved):
LGTM. Let's hope we will find more usecases and improve this in later iterations.

PR #694, opened 2024-01-24
Build: Migrate plugin-e2e to Vitest and use tsconfigs/base configs

jackw:

After migrating both create-plugin and sign-plugin to only build as ESM packages and use Vitest it's left this repo in a state of multiple test runners (Jest and Vitest) and copied over tsconfig settings from the [tsconfigs/bases](https://github.com/tsconfig/bases) project. This PR removes Jest from the repo and migrates plugin-e2e to use vitest.

It replaces our tsconfigs with the tsconfigs/bases "recommended" and "node20" configs. This change required fixing a load of TS errors.

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@3.1.2-canary.694.e630b31.0
  npm install @grafana/plugin-e2e@0.10.3-canary.694.e630b31.0
  npm install @grafana/sign-plugin@3.0.1-canary.694.e630b31.0
  # or 
  yarn add @grafana/create-plugin@3.1.2-canary.694.e630b31.0
  yarn add @grafana/plugin-e2e@0.10.3-canary.694.e630b31.0
  yarn add @grafana/sign-plugin@3.0.1-canary.694.e630b31.0
  ```
</details>


packages/create-plugin/src/commands/generate.command.ts
```diff
@@ -175,9 +184,15 @@ async function generateFiles({ actions }: { actions: any[] }) {
         path: action.path,
       });
     } catch (error) {
+      let message;
+      if (error instanceof Error) {
+        message = error.message;
+      } else {
+        message = String(error);
+      }
```
leventebalogh, line 192: Did you receive some errors due to the previous implementation?

packages/create-plugin/src/utils/utils.config.ts
```diff
@@ -23,10 +23,10 @@ export function getConfig(): CreatePluginConfig {
     return {
       ...rootConfig,
       ...userConfig,
-      version: rootConfig.version,
+      version: rootConfig!.version,
       features: createFeatureFlags({
-        ...rootConfig.features,
-        ...userConfig.features,
+        ...rootConfig!.features,
+        ...userConfig!.features,
```
leventebalogh, line 29: Aaa, these are necessary because of the `tsconfig` updates!

packages/plugin-e2e/src/api.ts
```diff
@@ -176,7 +177,7 @@ export type PluginFixture = {
 };
 
 // extend Playwright with Grafana plugin specific fixtures
-export const test = base.extend<PluginFixture, PluginOptions>(fixtures);
+export const test = base.extend<PluginFixture, PluginOptions>({ ...fixtures, ...options });
```
leventebalogh, line 180: I am probably missing something, but why did we need this change?

leventebalogh: I think I understand it now b reading further, it was mostly a refactoring.

packages/plugin-e2e/src/options/index.ts
```diff
@@ -0,0 +1,10 @@
+import path = require('path');
+import { Fixtures } from '@playwright/test';
+import { PluginOptions } from '../api';
+
+const options: Fixtures<{}, PluginOptions> = {
+  featureToggles: [{}, { option: true, scope: 'worker' }],
+  provisioningRootDir: [path.join(process.cwd(), 'provisioning'), { option: true, scope: 'worker' }],
+};
+
+export default options;
```
leventebalogh, line 10: **Nit:** is there a reason to have `options/index.ts` instead of just a simple `options.ts`? Do we imagine having more files under `options/`?

leventebalogh (approved):
LGTM, nice update of the tsconfig! 🚀

packages/create-plugin/src/commands/generate.command.ts
jackw: Yeah, I think enabling `strict` via the tsconfig changes caused this to error. Error is typed as `unknown` so `error.message || error.toString()` were causing complaints. Dunno if that helps explain it better?

packages/plugin-e2e/src/options/index.ts
jackw: I think it would be wise if we don't create any nested barrel files within any packages and always use named exports when re-exporting. Using nested barrel files makes it difficult to understand what is being exported and easy to export something for a test or use elsewhere that then becomes part of the public api.

**naughty**
```
// index.ts
export * from './directory';

// ./directory/index.ts
export * from './file';

```
**nice**
```
// index.ts
export {moduleA, moduleB, typeA} from './directory/file'
```

packages/create-plugin/src/commands/generate.command.ts
leventebalogh: Yea, absolutely, thanks for the explanation 👍

packages/create-plugin/src/utils/utils.config.ts
jackw: Yep! We're kind of abusing the try/catch here to figure out what to return so I thought it was _okay_ to do?

leventebalogh: I think it is 👍

sunker (approved):
Nice work on this @jackw!

PR #695, opened 2024-01-24
CI: Test scaffolds against built NPM packages

jackw:

Since removing plop we've been able to scaffold plugins via the cli by passing args however we're still testing them in CI via the "generate-<pluginType>" scripts. These scripts were originally created for development and don't mimic a user journey all that well. This PR aims to address this by doing the following:

- `jobs.test` now npm packs and uploads the npm packages as tarballs
- `jobs.generate-plugins` no longer clones the repo. Instead it downloads the previously built tarballs and installs them globally using `npm -g install <name_of_package>.tgz`.
- `jobs.generate-plugins` matrix has been updated to run each command using the globally installed package (`create-plugin` or `sign-plugin`) and passes args to answer any prompts.
- `jobs.generate-plugins` now additionally runs `typecheck` on the scaffolded plugins

Fixes #

Artifacts are only retained for a maximum of 1 day. I'm happy to change this but given how often this workflow is run I see no need to clutter the storage

From the current builds in this PR it appears these changes speed up the job by ~1.5mins.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
         with:
           node-version: '20'
           registry-url: 'https://registry.npmjs.org'
-          cache: 'npm'
+          # cache: 'npm'
 
       - name: Download packed artifacts
         uses: actions/download-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,16 @@ jobs:
         run: npm exec nx affected -- --target test
       - name: Build all packages
         run: npm run build
+      - name: Pack packages for testing
+        run: |
+          mkdir ./packed-artifacts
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -42,86 +52,87 @@ jobs:
...
+        working-directory: ./${{ matrix.workingDir }}
 
       - name: Lint plugin frontend
         run: npm run lint
-        working-directory: ./packages/create-plugin/generated
+        working-directory: ./${{ matrix.workingDir }}
+
+      - name: Typecheck plugin frontend
+        run: npm run typecheck
+        working-directory: ./${{ matrix.workingDir }}
```
leventebalogh, line 103: 👏

leventebalogh (approved):
Nice one! 🥇

PR #696, opened 2024-01-25
Fix broken livereload injection when using Grafana 10.2.3

samjewell:
I needed this version of Grafana to use the latest features in my
scenesapp plugin. But bumping to this version of Grafana broke the
livereload injection.
I haven't tracked down exactly when/how/why the old injection command
no longer works, but I have found this fix. If you're happy with it,
maybe we can just merge this change, without bothering to find the
cause of the issue? Or maybe you can help by tracking down the cause
(or help me to do so)?

Thanks

~~Also bumping the Grafana version here, so you can see this in action. But happy to revert the Grafana version and just keep the livereload change, if we need a smaller scope to get this merged.~~

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@3.1.2-canary.696.20ca811.0
  # or 
  yarn add @grafana/create-plugin@3.1.2-canary.696.20ca811.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/constants.ts
+++ b/packages/create-plugin/src/constants.ts
@@ -43,7 +43,7 @@ export enum PLUGIN_TYPES {
 // and will be available to use in the templates.
 // Example: "@grafana/ui": "{{ grafanaVersion }}"
 export const EXTRA_TEMPLATE_VARIABLES = {
-  grafanaVersion: '10.0.3',
+  grafanaVersion: '10.2.3',
   grafanaImage: 'grafana-enterprise',
 };
 
--- a/packages/create-plugin/src/constants.ts
+++ b/packages/create-plugin/src/constants.ts
@@ -43,7 +43,7 @@ export enum PLUGIN_TYPES {
 // and will be available to use in the templates.
 // Example: "@grafana/ui": "{{ grafanaVersion }}"
 export const EXTRA_TEMPLATE_VARIABLES = {
-  grafanaVersion: '10.2.3',
+  grafanaVersion: '10.0.3',
   grafanaImage: 'grafana-enterprise',
 };
 
--- a/packages/create-plugin/templates/common/.config/Dockerfile
+++ b/packages/create-plugin/templates/common/.config/Dockerfile
@@ -13,4 +13,4 @@ ENV GF_DEFAULT_APP_MODE "development"
 
 # Inject livereload script into grafana index.html
 USER root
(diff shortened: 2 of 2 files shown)


samjewell:
I guessed at the `patch` label on this PR. Please change if you'd like.

packages/create-plugin/src/constants.ts
```diff
@@ -43,7 +43,7 @@ export enum PLUGIN_TYPES {
 // and will be available to use in the templates.
 // Example: "@grafana/ui": "{{ grafanaVersion }}"
 export const EXTRA_TEMPLATE_VARIABLES = {
-  grafanaVersion: '10.0.3',
+  grafanaVersion: '10.2.3',
```
jackw, line 46: This isn't a straight-forward change I'm afraid. You'll notice the CI is now failing due to peer dependency issues. As such can you please revert this?

packages/create-plugin/templates/common/.config/Dockerfile
```diff
@@ -13,4 +13,4 @@ ENV GF_DEFAULT_APP_MODE "development"
 
 # Inject livereload script into grafana index.html
 USER root
-RUN sed -i 's/<\/body><\/html>/<script src=\"http:\/\/localhost:35729\/livereload.js\"><\/script><\/body><\/html>/g' /usr/share/grafana/public/views/index.html
\ No newline at end of file
+RUN sed -i 's/<\/body>/<script src=\"http:\/\/localhost:35729\/livereload.js\"><\/script><\/body>/g' /usr/share/grafana/public/views/index.html
```
jackw, line 16: I'm pretty sure what you have here will cover both 10.0.3 and 10.2.3. Did you test both cases locally? The only thing I might suggest is using `|` instead of `/` as the delimiter to make it easier to differentiate between them and the slashes in the URL.

```suggestion
RUN sed -i 's|</body>|<script src="http:\/\/localhost:35729\/livereload.js"><\/script>&|' /usr/share/grafana/public/views/index.html
```

jackw (commented):
@samjewell thanks so much for taking the time to put a PR together to fix this. 🏅 

> I haven't tracked down exactly when/how/why the old injection command
no longer works, but I have found this fix.

Great spot! The reason is due to changes in the Grafana codebase which now doesn't minify the html template in the same way which prevents the regex in the sed command from matching.

samjewell:
@jackw final review, so we can merge this? 🙏

jackw:
> @jackw final review, so we can merge this? 🙏

@samjewell apologies, I'll get to it when I can. I need to test it across Grafana versions to make sure it doesn't cause it to break in earlier versions of Grafana.

jackw (approved):
LGTM! 🚀 

Tested across 10.0.x, 10.1.x, and 10.2.x

PR #710, opened 2024-01-31
Create-plugin: Use react-router v6 by default

leventebalogh:
### What changed?
The goal of this PR is to by default scaffold app plugins with the v6 version of react-router. The scenes apps are still scaffolded with react-router@v5, as bumping this will require changes in the `@grafana/scenes` package, as soon as this is done we are going to scaffold those with v6 as well.

**Updating plugins**
This PR introduces a new feature flag called `useReactRouterV6`. This can later be used during plugin updates to migrate to the v6 version of react-router. (Currently not possible to do this with the create-plugin tool, as it is not touching dependencies).

- [x] Add a feature-toggle to use react-router v6 (used for updates, so we don't introduce a breaking change)
- [x] Move the scaffolded _app_ to by default use react-router v6
- [x] Make sure the scaffolded scenes app is still on react-router v5
- [x] Check if any tests need to be updated

@sympatheticmoose for visibility.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@3.3.0-canary.710.17aff14.0
  # or 
  yarn add @grafana/create-plugin@3.3.0-canary.710.17aff14.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/commands/generate.command.ts
+++ b/packages/create-plugin/src/commands/generate.command.ts
@@ -69,6 +69,7 @@ function getTemplateData(answers: CliArgs) {
     isNPM: packageManagerName === 'npm',
     version: currentVersion,
     bundleGrafanaUI: features.bundleGrafanaUI,
+    reactRouterV6: features.reactRouterV6,
   };
 
   return templateData;
--- a/packages/create-plugin/src/commands/generate.command.ts
+++ b/packages/create-plugin/src/commands/generate.command.ts
@@ -69,9 +69,8 @@ function getTemplateData(answers: CliArgs) {
     isNPM: packageManagerName === 'npm',
     version: currentVersion,
     bundleGrafanaUI: features.bundleGrafanaUI,
-    useReactRouterV6: features.useReactRouterV6,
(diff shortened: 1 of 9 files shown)


packages/create-plugin/src/commands/generate.command.ts
```diff
@@ -69,6 +69,8 @@ function getTemplateData(answers: CliArgs) {
     isNPM: packageManagerName === 'npm',
     version: currentVersion,
     bundleGrafanaUI: features.bundleGrafanaUI,
+    useReactRouterV6: features.useReactRouterV6 && pluginType === PLUGIN_TYPES.app,
+    reactRouterVersion: features.useReactRouterV6 && pluginType === PLUGIN_TYPES.app ? '6.22.0' : '5.2.0',
```
academo, line 73: can we make this work with only the boolean parameter? as they are right now it seems this second one is just a calculated field from the previous one.

academo (approved):
I tested this locally and it works as expected. nice work!

packages/create-plugin/src/commands/generate.command.ts
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/710/commits/17aff14e5abbfc595ce038235fc0d8fa57a3aa0d

PR #730, opened 2024-02-09
Plugin E2E: Fix feature toggles race condition

sunker:

Plugin E2E allows users to [override feature toggles](https://github.com/grafana/plugin-tools/pull/651) during test sessions. The way it works is that it simply wait until `window.grafanaBootData` is rendered, and then sets the toggles that the user defined in the playwright config. While working with other things, I've noticed there's some intermittent flakiness to this. I turns out it's because even though the `window.grafanaBootData` is defined, it does not necessarily mean that `window.grafanaBootData.settings.featureToggles` is defined yet. That meant the toggles that were set by plugin-e2e (sometimes) became overwritten by the ones coming from Grafana. 

This PR simply checks that Grafana have added its feature toggles to the boot data before plugin-e2e adds its overrides. 

Fixes #

### DIFF
--- a/packages/plugin-e2e/src/fixtures/scripts/overrideFeatureToggles.js
+++ b/packages/plugin-e2e/src/fixtures/scripts/overrideFeatureToggles.js
@@ -1,9 +1,8 @@
 // this script is evaluated in the browser context, so we cannot use typescript
 export const overrideFeatureToggles = (featureToggles) => {
-  const timeout = 5;
-
+  const timeout = 1;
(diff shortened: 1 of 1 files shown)


leventebalogh (approved):
LGTM 👍 

(Did decreasing the timeout to 1ms completely remove the flakiness?)

sunker:
> LGTM 👍
> 
> (Did decreasing the timeout to 1ms completely remove the flakiness?)

Yes it seems so! :)

PR #741, opened 2024-02-12
Plugin E2E: Fix typing in config page

sunker:

Fix a badly typed method in the config page. 

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/plugin-e2e@0.13.1-canary.741.4e6318e.0
  # or 
  yarn add @grafana/plugin-e2e@0.13.1-canary.741.4e6318e.0
  ```
</details>

### DIFF
--- a/packages/plugin-e2e/src/models/DataSourceConfigPage.ts
+++ b/packages/plugin-e2e/src/models/DataSourceConfigPage.ts
@@ -1,3 +1,4 @@
+import { Response } from '@playwright/test';
 import { DataSource, NavigateOptions, PluginTestCtx, TriggerQueryOptions } from '../types';
 import { GrafanaPage } from './GrafanaPage';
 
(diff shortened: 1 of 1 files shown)


packages/plugin-e2e/src/models/DataSourceConfigPage.ts
```diff
@@ -28,9 +29,10 @@ export class DataSourceConfigPage extends GrafanaPage {
    *
    * Optionally, you can skip waiting for the response by passing in { skipWaitForResponse: true } as the options parameter
    */
-  async saveAndTest(options?: TriggerQueryOptions) {
+  async saveAndTest(options?: TriggerQueryOptions): Promise<Response> {
```
academo, line 32: There's more than a  type fixing here. It is also adding an additional waitForResponse that was not there before. is this intended?

sunker: Yes this will resolve a playwright response. Looks a bit weird, but with an empty string it resolves immediately. 

The alternative with be to introduce a new method - `saveAndTestWithoutWaitingForResponse`, but I think the approach in this PR is better.

academo: Good. I was just making sure since your PR title and description only talked about types and this is actually changing business logic.

academo (approved):

Issue #742, opened 2024-02-12
Create Plugin: Update github workflows to use shared actions


sympatheticmoose:
This is to address inconsistency in our CI and Release workflows so we just have to maintain build-action as well as some security improvements (getting flagged by tooling)
```[tasklist]
### Tasks
- [x] Update [release.yaml](https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/github/workflows/release.yml) to use https://github.com/grafana/plugin-actions/blob/main/build-plugin/action.yml
- [x] Update [build-action](https://github.com/grafana/plugin-actions/blob/main/build-plugin/action.yml) versions i.e. setup-node -> v4
- [x] Add top level `permissions: read-all` and job level `permissions: contents: write` as appropriate
- [ ] Update ci.yaml to use plugin-actions
```

sympatheticmoose:
Created some quick PRs which cover everything except consolidation.

sympatheticmoose:
@jackw https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/github/workflows/ci.yml still seems to not use it?

jackw:
Have added it to the task list. 👍 

Not sure how best to handle adding plugin-actions to ci.yml. Probably needs splitting up cos it has dependencies on [other scaffolded workflows](https://github.com/grafana/plugin-tools/pull/1357/files#diff-97f3aa13af615c55d3d9151ff85d191d5e228ba2a99a070ad0a6f316b54a85c7).

PR #743, opened 2024-02-12
chore: update permissions

sympatheticmoose:

Introduces top level read-only permissions in the CI workflow

Moves write permissions to within the single job (kind of a vanity change but stops certain security tooling from complaining)

part of https://github.com/grafana/plugin-tools/issues/742

Tested with a custom plugin and CI/Release still appeared to execute as expected. 
CI workflow may require more changes as per the issue

### DIFF
--- a/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
@@ -8,11 +8,12 @@ on:
     tags:
       - 'v*' # Run workflow on version tags, e.g. v1.0.0.
 
-permissions:
-  contents: write
+permissions: read-only
 
 jobs:
   release:
+    permissions:
+      contents: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/release.yml
@@ -8,7 +8,7 @@ on:
     tags:
       - 'v*' # Run workflow on version tags, e.g. v1.0.0.
 
-permissions: read-only
+permissions: read-all
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


leventebalogh (commented):
Hmm, maybe I am doing something wrong, but this didn't work for me ([failing workflow](https://github.com/grafana/plugin-workflow-tester/actions/runs/7883330266/workflow)). Are you sure `read-only` is a valid permission value? As far as I can see [here](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes) to define access for all the available scopes we can either use `read-all` or `write-all`.

leventebalogh:
(Maybe it would be useful to automatically open a PR in the https://github.com/grafana/plugin-workflow-tester/ repo if any of the workflow template files change to see if they pass. However as we don't edit these too often maybe it doesn't worth the effort.)

sunker:
> (Maybe it would be useful to automatically open a PR in the https://github.com/grafana/plugin-workflow-tester/ repo if any of the workflow template files change to see if they pass. However as we don't edit these too often maybe it doesn't worth the effort.)

Nice, did not know we had this repo.

sympatheticmoose:
> Hmm, maybe I am doing something wrong, but this didn't work for me ([failing workflow](https://github.com/grafana/plugin-workflow-tester/actions/runs/7883330266/workflow)). Are you sure `read-only` is a valid permission value? As far as I can see [here](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes) to define access for all the available scopes we can either use `read-all` or `write-all`.

🤦

packages/create-plugin/templates/github/ci/.github/workflows/release.yml
```diff
@@ -8,11 +8,12 @@ on:
     tags:
       - 'v*' # Run workflow on version tags, e.g. v1.0.0.
 
-permissions:
-  contents: write
+permissions: read-all
```
leventebalogh, line 11: As we only have a single job in this workflow, what's the reason for this change at this point?

leventebalogh (approved):

PR #755, opened 2024-02-19
Plugin-e2e: Add yaml as a dependency

jackw:

Currently when importing plugin-e2e an error appears complaining that the `yaml` package isn't available. This is due to it not being a direct dependency of the project.

image

The PR adds it as a dependency.

Fixes #


sunker (approved):
Good catch!

leventebalogh (approved):

PR #756, opened 2024-02-20
Docs: Add the react-router migration docs to `v10.x-v11.x` as well

leventebalogh:
### What changed?

- [x] Extract the react-router migration docs as a partial
- [x] Add the react-router migration docs to the `v10.x-v11.x` guide as well


josmperez (approved):
This LGTM and I see no blocking issues. My only concern is that the new shared file isn't located under the docs/docusaurus/shared path, so this might cause a bit of confusion down the road.

leventebalogh:
Thanks, that's a good point @josmperez, I have addressed it in https://github.com/grafana/plugin-tools/pull/756/commits/66e434883e9abbb40c071112c8d6faa85fe999a8. 👍

PR #775, opened 2024-02-23
Chore: avoid race conditions in label checks

tolzhabayev:
Avoid unnecessary runs of label checks when multiple actions performed (label, unlabel) in a short amount of time.

### DIFF
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -9,7 +9,7 @@ on:
       - unlabeled
       - edited
 concurrency:
-  group: check-labels-${{ github.event.number }}
+  group: ${{ github.workflow }}-${{ github.event.number }}
   cancel-in-progress: true
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-labels.yml
```diff
@@ -8,6 +8,9 @@ on:
       - labeled
       - unlabeled
       - edited
+concurrency:
+  group: check-labels-${{ github.event.number }}
```
leventebalogh, line 12: Could we use `${{ github.workflow }}-${{ github.ref }}` as well to cancel the in-progress runs of the current workflow? ([docs](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow))

leventebalogh: Aaa, I see, thanks for the clarification 👍

leventebalogh (approved):
LGTM 👍

PR #776, opened 2024-02-26
Docs: Use correct panel context type in extension example

svennergr:

Since `PanelContext` is already a type in `@grafana/ui` specifying the correct type for the extension context can be useful.


mckn (approved):
LGTM! Great feedback!

PR #800, opened 2024-03-03
Add zip command for plugins

SandersAaronD:
This PR adds an npx utility for zipping plugins, analogous to the one for signing plugins.

Unsure where issues are open, does concern ongoing migration to Github Actions where this workflow is currently being used ad-hoc as a bash script written in the Github Actions yaml. It did not seem tenable to extend this to, specifically, the case of zipping plugins with backends that have OS and architecture-dependent binaries, so I moved it over to this.


jackw (changes requested):
Hey @SandersAaronD, I'm sorry but I'm going to block this for now as we have an [open issue](https://github.com/grafana/plugin-tools/issues/411) with [on going discussions](https://github.com/grafana/grafana-com/issues/8027) and we have already [closed one PR](https://github.com/grafana/plugin-tools/pull/447) related to this. Right now we aren't sure how we would like to tackle this as multiple architectures for the community makes things a lot more complicated for submission, review etc. Plus when there's only a single zip we have the additional maintenance of an npm package for doing:

```
mv dist/ myorg-simple-panel
zip myorg-simple-panel-1.0.0.zip myorg-simple-panel -r
```

I think it would probably work for all if we invest our time in the existing [build-plugin action](https://github.com/grafana/plugin-actions/blob/main/build-plugin/README.md) with a `with.multiArch` config flag to support multi-archtecture if that is required. But I would ask you to join those discussions first to help us better understand your needs and we can find a way forward.

SandersAaronD:
@jackw This defaults to zipping the default 'dist' folder if no arguments are provided, mirroring the sign-plugin command. So it is fine for single-binary builds, and I can add more documentation to that effect.

I agree that this should ideally be run in a github action. I don't think that means it shouldn't be in this repository; the github action can call the script with npx to run it, and that is what I am doing right now [here](https://github.com/grafana/grafana-llm-app/blob/4d0eaf4f8a0d5fb4865d3fec13561c56b87979bb/.github/workflows/release.yml#L119). This is what I'm currently doing because it solves my current build problem. If the code were to be built directly in a github action we would have to commit the compiled code and, basically, use Git as a package manager for Node. That seems like a bad practice, especially when we already have very good scripts here that do their jobs extremely well and that can be package managed through NPM normally.

If this isn't a good fit for this repo that's fine, but that gives me no clear path to upstreaming my changes and leads me to want to just fork this repository instead.

PR #805, opened 2024-03-06
Plugin Meta Extractor: Initial version

mckn:
## What is this?
The **Plugin Meta Extractor (`@grafana/plugin-meta-extractor`)** is a new package that can be used to extract meta-information from the plugin's source code. The goal is to be able to automatically generate static information about the plugin without needing the plugin-dev to care about these (e.g. extension point information).
[View README](https://github.com/grafana/plugin-tools/blob/mckn/pme-tool/packages/plugin-meta-extractor/README.md)
[PR for adding the plugin to our webpack config](https://github.com/grafana/plugin-tools/pull/871)

### What kind of meta-data?
Currently we only extract basic information about the extensions that the plugin registers.

### Supported plugin format
The tool currently only recognises extensions that are registered in the following format:
```ts
// src/module.ts
export const plugin = new AppPlugin<{}>()
  .setRootPage(App)
  .configureExtensionLink<PluginExtensionPanelContext>({ 
    // ...
  })
  .configureExtensionLink<PluginExtensionPanelContext>({ 
    // ...
  });
```

### How does it work?
It uses the typescript AST to find usages of the functions that register extensions. The package also exposes a Webpack plugin that can be used to generate a meta-info file during the build of the plugin.

### How to use it?

**CLI**
1. Check out the branch
2. Install dependencies - `npm install`
3. Go to the package - `cd packages/plugin-meta-extractor`
4. Build the package - `npm run build`
5. Run the binary - `./dist/bin/run.js ~/my-grafana-plugin/src/module.ts` _(prints to the console)
6. Use `npm link` to depend on the local version of the package while developing

**Code**
```ts
import { extractPluginMeta } from '@grafana/plugin-meta-extractor';

const entryPoint = `${PLUGIN_ROOT}/src/module.ts`;
const pluginMeta = extractPluginMeta(entryPoint); // see the format of the data below
```

**Webpack**
```ts
import { GrafanaPluginMetaExtractor } from '@grafana/plugin-meta-extractor';

export default {
 // ...
 plugins: [
  // updates the `generated` property of the plugin.json in the dist directory
  new GrafanaPluginMetaExtractor(),
  // ...
  ]
};
```

### What data does it give back?
It always returns with a valid JSON in the following format:
```json
{
  "extensions": [
    {
      "type": "link",
      "extensionPointId": "grafana/dashboard/panel/menu",
      "title": "Open from time series or pie charts (path)",
      "description": "This link will only be visible on time series and pie charts"
    },
    {
      "type": "link",
      "extensionPointId": "grafana/dashboard/panel/menu",
      "title": "Open from time series or pie charts (onClick)",
      "description": "This link will only be visible on time series and pie charts"
    }
  ]
}
```

### Todo
- [x] Use the webpack config entry parameter and handle the different scenarios
  - [x] A single string: `entry: "./module.ts"`
  - [x] An array of strings: `entry: ["./module.ts", "./datasource/module.ts"]`
  - [x] An object of entries: `entry: { module: "./module.ts" }`
  - [x] A promise resolving any of the above
- [x] Test it with all plugin versions (app, panel, datasource)
- [x] Only update the plugin.json if necessary

<details>
  <br />

  ```bash
  npm install @grafana/plugin-meta-extractor@0.0.1-canary.805.f998cc3.0
  # or 
  yarn add @grafana/plugin-meta-extractor@0.0.1-canary.805.f998cc3.0
  ```
</details>


packages/plugin-meta-extractor/src/webpack-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+export type GrafanaPluginMetaExtractorOptions = {
+  entryFile?: string;
+  outputAssetName?: string;
+};
+
+export class GrafanaPluginMetaExtractor {
+  options: GrafanaPluginMetaExtractorOptions;
+
+  constructor(options: GrafanaPluginMetaExtractorOptions) {
+    this.options = options;
```
jackw, line 16: It might be worth using [schema-utils](https://github.com/webpack/schema-utils) to validate options being passed to the plugin.

packages/plugin-meta-extractor/src/webpack-plugin.ts
```diff
@@ -0,0 +1,35 @@
+import path from 'node:path';
+import * as webpack from 'webpack';
+import { extractPluginMeta } from './meta-extractor.js';
+
+const DEFAULT_GENERATED_FILENAME = 'plugin.generated.json';
+
+export type GrafanaPluginMetaExtractorOptions = {
+  entryFile?: string;
+  outputAssetName?: string;
```
jackw, line 9: Are we sure we want to expose this? To the best of my knowledge the name of the outputted file will be a contract of sorts with the Grafana server.

packages/plugin-meta-extractor/src/webpack-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class GrafanaPluginMetaExtractor {
+  options: GrafanaPluginMetaExtractorOptions;
+
+  constructor(options: GrafanaPluginMetaExtractorOptions) {
+    this.options = options;
+  }
+
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.emit.tapAsync('GrafanaPluginMetaExtractor', (compilation, callback) => {
```
jackw, line 20: I think `emit` is called before writing assets to the output dir. I'm thinking it's highly unlikely but... is it possible this plugin could clobber a file with the same output name?

jackw (commented):
Great work on this @mckn and @leventebalogh ! 👏 

I've left a few comments for consideration.

packages/pme/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "directory": "packages/pme",
+    "url": "https://github.com/grafana/plugin-tools"
+  },
+  "keywords": [
+    "grafana",
+    "plugin",
+    "tool"
+  ],
+  "author": "Grafana",
+  "license": "Apache-2.0",
```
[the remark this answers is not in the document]
leventebalogh: Good catch, will do 👍

packages/plugin-meta-extractor/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "types": ["node", "jest"],
+    "outDir": "./dist",
+    "stripInternal": true,
+    "removeComments": false,
+    "declaration": true,
+    "allowJs": true
+  },
+  "exclude": ["node_modules", "*.test.ts"],
```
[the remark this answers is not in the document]
leventebalogh: Nah, I think we can remove this.

packages/plugin-meta-extractor/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "lib": ["ESNext"],
```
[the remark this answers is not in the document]
leventebalogh: Good catch 👍

packages/plugin-meta-extractor/src/webpack-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+      const { webpack } = compiler;
+      const { RawSource } = webpack.sources;
+      const entryFile = this.options.entryFile || 'index.js';
+      const entryFilePath = path.resolve(compiler.context, entryFile);
+      const pluginMeta = extractPluginMeta(entryFilePath);
+
+      const newAssetName = this.options.outputAssetName || DEFAULT_GENERATED_FILENAME;
+      const newAssetContent = JSON.stringify(pluginMeta, null, 2);
+
+      compilation.assets[newAssetName] = new RawSource(newAssetContent);
```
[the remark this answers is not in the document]
leventebalogh: I am not 100% sure that it can be used in this hook. But will check 👍

packages/plugin-meta-extractor/src/webpack-plugin.ts
leventebalogh: Yepp, that's a good suggestion! (Although we might don't need this in case we use the entries from the webpack config.)

packages/plugin-meta-extractor/src/webpack-plugin.ts
leventebalogh: I am not sure yet to be honest, will double-check, thanks!

sunker (commented):
Think this looks really cool. Awesome work on this! 
Some high level questions from me. 

In real world scenarios today, can it be the case that an app plugin conditionally configures an extension? For example only registering an extension link in case a certain feature toggle is enabled. In that case, will it be problematic that extensions are extracted at build time rather than runtime?

If I understand correctly, we can traverse the AST also in the case the entry file is a javascript file. So in theory we could support extraction also for javascript plugins? It seems we use the typechecker in a few places in the code. Are we targeting typescript plugins only? 

I understand this is still in experimental phase, but I'm curious to know how this fits into the story around registering extensions. Do you think using this tool to extract extensions built time could become a requirement, or would we support extracting them runtime too? How does this relate to the reactive extensions registry feature (that is not merged yet)?

packages/plugin-meta-extractor/README.md
```diff
@@ -0,0 +1,92 @@
+# Grafana / Plugin Meta Extractor
+
+`@grafana/plugin-meta-extractor` is a small package that can be used to extract meta-information from the source code of a Grafana plugin. The main use case is to be able to automatically generate static information for a plugin without the need for manual intervention.
```
tolzhabayev, line 3: I think what is missing in the readme is:
- why this is needed. Yes we say "generate static information" but probably need more explanation to why
- How often and when this should run and where (locally, ci, etc.)

marefr:
Are you imagine the generated content to go into the plugin.json or how are you imagine this to be consumed within grafana?

packages/pme/package.json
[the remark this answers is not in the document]
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/a1e7b8965e65006d31a746a46625f906a2b32251.

packages/plugin-meta-extractor/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "types": ["node", "jest"],
```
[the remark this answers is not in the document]
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/36ed164b6b39087d8ed706d9f32de5045d6729d8.

packages/plugin-meta-extractor/tsconfig.json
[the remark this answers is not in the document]
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/c8918f797518fef6ad1e22b1538dd4e314a6290d.

packages/plugin-meta-extractor/src/webpack-plugin.ts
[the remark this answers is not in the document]
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/eeb37605c8a122e12f059a912ad86817c4fd1e26.

packages/plugin-meta-extractor/tsconfig.json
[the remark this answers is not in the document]
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/61f2c9cf566d5d86c93b61e51a2595849b95e452.

packages/plugin-meta-extractor/src/webpack-plugin.ts
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/d8b0c6cfe13859e9d67520e369ce02d2ca818950.

packages/plugin-meta-extractor/README.md
```diff
@@ -0,0 +1,92 @@
...
+// webpack.config.ts
+// -----------------
+
+import { GrafanaPluginMetaExtractor } from '@grafana/plugin-meta-extractor';
+
+export default {
+  // ...
+  plugins: [
+    // Creates a "plugin.generated.json" file in the output directory
+    // `entryFile` - an absolute path pointing to the module.(ts|tsx) file of your plugin
```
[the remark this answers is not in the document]
leventebalogh: After having a look I think it would be tricky to completely rely on the webpack config inside the plugin.

`compiler.options.entry` would give us the actual entry points, however based on it's types it can also be a function that returns a promise, or an object with different entry points (which is based on how the user writes the webpack config). We could say that we create a contract between our webpack config and this plugin, however I am not sure if that's the best thing to do. I would probably rather keep the argument to be more explicit and reusable.

packages/plugin-meta-extractor/src/webpack-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  constructor(options: GrafanaPluginMetaExtractorOptions) {
+    this.options = options;
+  }
+
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.emit.tapAsync('GrafanaPluginMetaExtractor', (compilation, callback) => {
+      const { webpack } = compiler;
+      const { RawSource } = webpack.sources;
+      const entryFile = this.options.entryFile || 'index.js';
```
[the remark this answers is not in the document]
leventebalogh: Resolving as it's discussed in https://github.com/grafana/plugin-tools/pull/805#discussion_r1559543055.

leventebalogh:
@sunker Thanks for the questions!

>In real world scenarios today, can it be the case that an app plugin conditionally configures an extension? For example only registering an extension link in case a certain feature toggle is enabled. In that case, will it be problematic that extensions are extracted at build time rather than runtime?

Unfortunately we don't yet support every dynamic pattern with this tool for registering extensions from a plugin (we only support basic chained method calls on the AppPlugin instance at this point) - however we are planning to do so in the future.
The `plugin.generated.json` file that we generate will only give information to Grafana core about when a plugin should be preloaded. In case a plugin uses some runtime conditions to decide if it registers an extension, the tool should record the static meta-data for the extension regardless, which means that the plugin will always be preloaded for that extension-point, and it can decide during runtime if it registers the extension.

(Something like this can also be done in the chained-method-call pattern: returning `undefined` from the `configure()` function of the extension will hide the extension.)

>If I understand correctly, we can traverse the AST also in the case the entry file is a javascript file. So in theory we could support extraction also for javascript plugins? It seems we use the typechecker in a few places in the code. Are we targeting typescript plugins only?

I am afraid this will only work properly for plugins written in typescript at this point (`module.ts` / `module.tsx`).

>I understand this is still in experimental phase, but I'm curious to know how this fits into the story around registering extensions. Do you think using this tool to extract extensions built time could become a requirement, or would we support extracting them runtime too? How does this relate to the reactive extensions registry feature (that is not merged yet)?
This tool is planned to be used seamlessly in our plugin build configuration that is provided by create-plugin, however the meta-info it generates will only be used to decide when a plugin should be preloaded. (Plugins will still be able to dynamically register an extension point, for these scenarios - at least for now - we will always need to preload the plugin.)

packages/plugin-meta-extractor/README.md
leventebalogh: I updated the intro a bit in https://github.com/grafana/plugin-tools/pull/805/commits/4162173898580b4ce30f73c6bc3ce261cb8d08a4 - is this something similar to what you were thinking about?

packages/plugin-meta-extractor/src/webpack-plugin.ts
mckn: So I guess the question is if we want plugin developers to be able to override it? If it lives in the build config it can be overriden by the plugin author.

mckn (commented):
I think this looks great! A question regarding the webpack plugin. Will it update the plugin.json in the `/dist` folder or the one in `/src`?

leventebalogh:
> I think this looks great! A question regarding the webpack plugin. Will it update the plugin.json in the /dist folder or the one in /src?

That's a good question. It only updates the one in the `/dist` folder and leaves the one in `/src` intact. (When running `npm run dev` it updates the one in the dist folder on every code change.)

packages/plugin-meta-extractor/src/webpack-plugin.ts
leventebalogh: Fixed in https://github.com/grafana/plugin-tools/pull/805/commits/918a35cf4e548fedb6177a81265a4f61ee67f28b.

packages/plugin-meta-extractor/src/webpack-plugin.ts
leventebalogh: (Resolving this as in the end we are updating the `plugin.json` file.)

packages/plugin-meta-extractor/README.md
```diff
@@ -0,0 +1,92 @@
+# Grafana / Plugin Meta Extractor
+
+`@grafana/plugin-meta-extractor` is a small package that can be used to extract meta-information from the source code of a Grafana plugin. The main use case is to be able to automatically generate static information for a plugin without the need for manual intervention, which can help Grafana core to understand when a plugin that registers extensions should be preloaded without running the plugin code.
+
+(If you are developing a plugin there is nothing that you _have to_ do - only make sure that your plugin build workflow is up-to-date by running `npx @grafana/create-plugin@latest update`, and the build config will run this tool in the background.)
+
+## Install
+
+```
+npm install @grafana/plugin-meta-extractor
```
jackw, line 10:
```suggestion
npm install @grafana/plugin-meta-extractor --save-dev
```

packages/plugin-meta-extractor/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+## Usage
+
+#### With Webpack
+
+The package exposes a webpack plugin that can be used to generate plugin meta-information on every build based on the source code. The plugin is adding the information to the `plugin.json` file by overriding the `generated` property in the JSON.
```
jackw, line 17: nit:

```suggestion
The package exposes a webpack plugin that can be used to generate plugin meta-information on every build based on the source code. The plugin adds additional information to the `plugin.json` file by overriding the `generated` property in the JSON.
```

packages/plugin-meta-extractor/src/bin/run.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// Check if the file exists
+if (!fs.existsSync(entryPath)) {
+  console.error(`File not found: ${entryPath}`);
+  process.exit(1);
+}
+
+// Check if it is a module file
+const fileName = entryPath.split('/').pop();
+if (!fileName?.match(/^(module\.ts|module\.tsx)$/)) {
```
jackw, line 22: nit: I've not checked this but the following should be able to match without need for duplicating the filename

```suggestion
if (!fileName?.match(/^(module\.tsx?)$/)) {
```

packages/plugin-meta-extractor/README.md
```diff
@@ -0,0 +1,92 @@
...
+    extensionPointId: string,
+    title: string,
+    description: string
+  }>
+}
+```
+
+## Plugin support
+
+At this point the tool requires the `module.(ts|tsx)` to be in a certain format to be able to parse it (we are planning to improve on this in the future). The method calls for registering extensions have to be in the `module.(ts|tsx)`, and they need to be called on the `AppPlugin` instance in a "chained" manner:
```
jackw, line 74: nit:

```suggestion
This package requires the `module.(ts|tsx)` to be in a certain format to be able to parse it (we are planning to improve on this in the future). The method calls for registering extensions have to be in the `module.(ts|tsx)`, and they need to be called on the `AppPlugin` instance in a "chained" manner:
```

jackw (approved):
Awesome work on this @mckn and @leventebalogh . 🚀

packages/plugin-meta-extractor/src/bin/run.ts
leventebalogh: Nice, thanks! 👍

Issue #735, opened 2024-02-09
feat: add support for remote debugging in docker development environment


sympatheticmoose:
To support a better DX using our recommended workflows, we need to remove the docker incompatibility of our initial support for remote debugging of backend applications. Possibly with use of `dlv attach $(pgrep whatever-is-in-plugin-json.executable)`

As part of this effort we need to reinstate the documentation for using the debugger as this was bundled with local development guidance https://github.com/grafana/plugin-tools/pull/661/files

oshirohugo:
Closed by https://github.com/grafana/grafana-plugin-sdk-go/pull/926

PR #809, opened 2024-03-12
Create Plugin: add support for remote debugging in docker dev env

oshirohugo:

This PRs changes our development docker environment to allow remote debugging for the backend.
The work was based on the draft PR https://github.com/grafana/plugin-tools/pull/773 and aims to fulfill it's pending items:

- [x] Move backend plugin build inside docker container and supervise a mage build:watch
  - [x] docker-compose would need to map the git repo to a volume in order to build/rebuild with mage
- [x] Modify delve to always kill process on disconnect
- [x] Get the gpx_* binary name from src/plugin.json vs having to set env var
- [x] Maybe detect from src/plugin.json if backend is supported (and do not try to spawn delve), this can still be useful for frontend debug tools
- [x] Another option - write a small watcher service to respawn dlv with the new PID when process is killed (mage build:watch might be sufficient)
- [x] tested on alpine

The `mage watch` target is changed in this PR: https://github.com/grafana/grafana-plugin-sdk-go/pull/926

And the necessary base Docker image will be added by the PR: https://github.com/grafana/grafana/pull/83321

This PRs is part of:
https://github.com/grafana/plugin-tools/issues/735

This PR covers only backend debugging.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@4.3.0-canary.809.5b2f9cc.0
  npm install @grafana/plugin-e2e@0.20.0-canary.809.5b2f9cc.0
  # or 
  yarn add @grafana/create-plugin@4.3.0-canary.809.5b2f9cc.0
  yarn add @grafana/plugin-e2e@0.20.0-canary.809.5b2f9cc.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/.config/Dockerfile
+++ b/packages/create-plugin/templates/common/.config/Dockerfile
@@ -26,11 +26,18 @@ USER root
 
 
 # Installing supervisor and inotify-tools
-RUN apt-get update && \
-    apt-get install -y supervisor && \
-    apt-get install -y  inotify-tools && \
-    rm -rf /var/lib/apt/lists/*
+RUN if grep -i -q alpine /etc/issue; then \
+      apk add supervisor inotify-tools git; \
+    elif grep -i -q ubuntu /etc/issue; then \
+      DEBIAN_FRONTEND=noninteractive && \
+      apt-get update && \
+      apt-get install -y supervisor inotify-tools git && \
+      rm -rf /var/lib/apt/lists/*; \
+    else \
+      echo 'ERROR: Unsupported base image' && /bin/false; \
+    fi
 
+COPY supervisord/supervisord.conf /etc/supervisor.d/supervisord.ini
 COPY supervisord/supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 
 
@@ -56,4 +63,6 @@ RUN git clone https://github.com/magefile/mage; \
 RUN sed -i 's|</body>|<script src="http://localhost:35729/livereload.js"></script></body>|g' /usr/share/grafana/public/views/index.html
 
 
-ENTRYPOINT ["/usr/bin/supervisord", "-c", "/etc/supervisor/supervisord.conf"]
+COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
--- a/packages/create-plugin/templates/common/.config/Dockerfile
+++ b/packages/create-plugin/templates/common/.config/Dockerfile
@@ -1,7 +1,12 @@
-ARG grafana_version=latest
-ARG grafana_image=grafana-enterprise
+ARG ARCH=amd64
+ARG BASE_IMAGE=grafana/grafana-oss-${ARCH}:development-ubuntu
 
-FROM grafana/${grafana_image}:${grafana_version}
+FROM ${BASE_IMAGE} as plugindev
+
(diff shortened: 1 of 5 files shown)


packages/create-plugin/templates/common/.config/Dockerfile
```diff
@@ -1,7 +1,12 @@
-ARG grafana_version=latest
-ARG grafana_image=grafana-enterprise
+ARG ARCH=amd64
+ARG BASE_IMAGE=grafana/grafana-oss-${ARCH}:development-ubuntu
```
marefr, line 2: This changes https://github.com/grafana/plugin-tools/pull/197. @mckn we still would like enterprise as the default?

packages/create-plugin/templates/common/.config/Dockerfile
```diff
@@ -1,7 +1,12 @@
-ARG grafana_version=latest
```
sunker, line 1: Think we still need a way to specify version. https://grafana.com/developers/plugin-tools/get-started/set-up-development-environment#configure-the-grafana-version

packages/create-plugin/templates/common/.config/Dockerfile
oshirohugo: It's possible to maintain the base image as it was before with a few adjustments, we would have to select alpine or ubuntu to install the necessary packages (dlv, supervisord, etc.). The differences that I can see is that the development image suggested by Brian, provides node running in development mode and the possibility to build for an specific architecture. (maybe I'm missing other points)
Regarding debugging, capabilities, it's the same.

@briangann Do you think that it would be a significant loss to go with grafana-enterprise as a base image?

oshirohugo: Another idea would be to provide two totally separated environments, like staging and dev. The dev one we do based on the dev images and the staging we do based on grafana enterprise.

oshirohugo: @marefr we decided to use the non dev base images as it was before, therefore the Dockerfiles will still have the grafana enterprise as default.

packages/create-plugin/templates/common/.config/Dockerfile
oshirohugo: @sunker thank you for your review, we decided to use the non dev base images the same way as it was before, therefore we have again the ability to set grafana version. I've updated the PR to do that. This should also solve the issue with the e2e tests.

packages/create-plugin/templates/common/.config/Dockerfile
marefr: 👍

packages/create-plugin/templates/common/docker-compose.yaml
```diff
@@ -11,6 +13,23 @@ services:
...
     ports:
       - 3000:3000/tcp
+{{#if hasBackend}}
+      - 2345:2345/tcp # delve
+{{/if}}
+    security_opt:
+      - "apparmor:unconfined"
+      - "seccomp:unconfined"
+    cap_add:
+      - SYS_PTRACE
```
marefr, line 23: Nit. If this is needed for debug maybe only render if `hasBackend`?

marefr (approved):
LGTM this looks great. Haven't tested it though

oshirohugo:
@sunker could check if everything is ok now?

sunker (approved):
Not tested this either, but looks good to me. This will be a big improvement in DX. Nice work!

PR #810, opened 2024-03-12
Plugin E2E: Add QueryField container selector

kevinwcyu:
Adds a selector for the [`QueryField` component](https://github.com/grafana/grafana/blame/265200799d9f142622a99f3561e1864f84216ecd/packages/grafana-e2e-selectors/src/selectors/components.ts#L433)

Fixes #

Tested with Grafana version 10.2.5 and 10.3.0. The selector was originally added some time in v7 of Grafana from what I see from the git history, and it was originally `Query Field`.


sunker (approved):
Great, thanks for contributing @kevinwcyu!

PR #831, opened 2024-03-19
Docs: Clarify some points in documentation and typedocs

adamyeats:
This PR clarifies some very minor things that I tripped up on while setting up the E2E tests in `grafana/surrealdb-datasource`. Please consider this a first pass as I intend on taking another pass at this in future.

(@sunker Again, great job on this, feels like a huge step forward from where we were Cypress! 🚀)


sunker (approved):
Thanks for raising this PR and many thanks for feedback @adamyeats! It's very appreciated. 

The auth state mechanism is also described in the [Authentication](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/use-authentication) article, so I pushed a commit that adds the same comment to the snippets in this file.

PR #82, opened 2022-10-19
Feature: Add signPlugin.version to manifest to sign-plugin

jackw:
This PR adds `signPlugin.version` to manifest calls to gcom. Previously this field was called toolkit. This PR will cause signing attempts to fail until gcom is updated to cater for this new field. ✌️

PR #834, opened 2024-03-21
Create Plugins: Fix plugin id mismatch

oshirohugo:
**What this PR does / why we need it**:

There was some mismatching in the pluginId created in the templates, this PR fix it.

**Which issue(s) this PR fixes**:

Fixes https://github.com/grafana/plugin-tools/issues/758

**Special notes for your reviewer**:

To test this, I create projects using the old version and the new version for all plugin types: app, datasource, panel and scenesapp (with all the options on). I compared the results, and they were the same for app, datasource and panel. For scenesapp the fixes made all the pluginIds to be the same and the export path new has the `-app` sufix instead of `-scenesapp`

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/create-plugin@4.2.6-canary.834.966fae0.0
  # or 
  yarn add @grafana/create-plugin@4.2.6-canary.834.966fae0.0
  ```
</details>

PR #837, opened 2024-03-22
Create-plugin: Refactor getTemplateData to be used in generate command and update commands

### LINKED ISSUE #758: Bug: scenesapp backend created with different plugin-id

- [X] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E

### Package versions

```shell
@grafana/create-plugin@4.1.0
```

### What happened?

generating a scenes app with backend results in a main.go with a different name as the plugin-id

the plugin-id should probably match src/plugin.json

1. In this environment...
  Node 20
2. With this config...
  npx
3. Run '...'
```SCRIPT
❯ npx @grafana/create-plugin@latest
Need to install the following packages:
@grafana/create-plugin@4.1.0
Ok to proceed? (y) y
✔ What is going to be the name of your plugin? · pleasework
✔ What is the organization name of your plugin? · briangann
✔ How would you describe your plugin? · something
✔ What type of plugin would you like? · scenesapp
✔ Do you want a backend part of your plugin? (y/N) · true
✔ Do you want to add Github CI and Release workflows? (y/N) · true
✔ Do you want to add a Github workflow for automatically checking "Grafana API compatibility" on PRs? (y/N) · true
```
5. See error...

```SCRIPT
❯ grep -r "briangann-pleasework-scenesapp"
./docker-compose.yaml:    container_name: 'briangann-pleasework-scenesapp'
./docker-compose.yaml:      - ./dist:/var/lib/grafana/plugins/briangann-pleasework-scenesapp
./pkg/main.go:	if err := app.Manage("briangann-pleasework-scenesapp", plugin.NewApp, app.ManageOpts{}); err != nil {
```

```SCRIPT
❯ grep -r "briangann-pleasework-app" *
provisioning/plugins/app.yaml:  - type: 'briangann-pleasework-app'
src/plugin.json:  "id": "briangann-pleasework-app",
```

This doesn't appear to cause any issues running the plugin, but there is a disconnect in naming, and also the container_name.

The validator (and gcom) doesn't understand `-scenesapp` as a suffix, we should probably just use `-app` here.

### Environment

```shell
System:
    OS: macOS 14.3
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i9-12900K
    Memory: 20.53 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.volta/tools/image/node/20.11.0/bin/node
    Yarn: 4.1.0 - ~/.volta/tools/image/yarn/4.1.0/bin/yarn
    npm: 10.2.4 - ~/.volta/tools/image/node/20.11.0/bin/npm
    pnpm: 8.15.1 - ~/.volta/bin/pnpm
  Browsers:
    Brave Browser: 121.1.62.165
    Chrome: 121.0.6167.184
    Safari: 17.3
```

### Additional context

_No response_

Ukochka:
After removing Plop another way to getting template data was introduced and it is a duplication of getTemplate data we already had. We need to unify getting template data for all the commands: generate, migrate, update, provisioning

After refactoring: getTemplateData returns TemplateData object. It takes data from User prompt for generate command and from package.json for other commands

Fixes #82

I merged changes from this [PR](https://github.com/grafana/plugin-tools/pull/834), which fixes https://github.com/grafana/plugin-tools/issues/758

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@4.9.1-canary.837.e9165e7.0
  npm install @grafana/plugin-e2e@1.2.1-canary.837.e9165e7.0
  # or 
  yarn add @grafana/create-plugin@4.9.1-canary.837.e9165e7.0
  yarn add @grafana/plugin-e2e@1.2.1-canary.837.e9165e7.0
  ```
</details>


packages/create-plugin/src/commands/generate.command.ts
```diff
@@ -19,9 +23,8 @@ import { updateGoSdkAndModules } from './generate/update-go-sdk-and-packages.js'
 import { CliArgs, TemplateData } from '../types.js';
 
 export const generate = async (argv: minimist.ParsedArgs) => {
-  const answers = await promptUser(argv);
-  const templateData = getTemplateData(answers);
-  const exportPath = getExportPath(answers.pluginName, answers.orgName, answers.pluginType);
+  const templateData = getTemplateData(buildProxyFromUserPrompt(await promptUser(argv)));
```
jackw, line 26: nit: it might just be me but I find we lose legibility by nesting multiple function calls inside each other. I have to read "from the inside out" rather than top to bottom which I think is more natural.

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -4,7 +4,7 @@ import fs from 'node:fs';
 import { mkdirp } from 'mkdirp';
 import createDebug from 'debug';
 import { filterOutCommonFiles, isFile, isFileStartingWith } from './utils.files.js';
-import { renderHandlebarsTemplate } from './utils.handlebars.js';
+import { renderHandlebarsTemplate, normalizeId } from './utils.handlebars.js';
```
jackw, line 7: I think we should move normalizeId out of this file and delete the normalize_id helper entirely to prevent future changes reintroducing it to the templates.

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -88,23 +92,70 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
   return renderHandlebarsTemplate(fs.readFileSync(templateFile).toString(), data);
 }
 
-export function getTemplateData(): TemplateData {
+export function buildProxyFromJson() {
   const pluginJson = getPluginJson();
+  return new Proxy(pluginJson, {
```
leventebalogh, line 97: What's the reason for using a `Proxy` here? (Instead of using a simple object with overrides?)

jackw, line 97: May I ask what the benefits are of using a proxy here are?

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -88,23 +92,70 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
...
+      } else if (prop === 'packageManager') {
+        return getPackageManagerWithFallback();
+      } else {
+        throw new Error(`Property ${String(prop)} not found in package.json.`);
+      }
+    },
+  });
+}
+export function buildProxyFromUserPrompt(target: CliArgs) {
+  return new Proxy(target, {
```
jackw, line 124: As above, I'm not sure what the benefits are of using a proxy here.

jackw (commented):
Nice @Ukochka. This is looking good. I've left a nit and a couple of questions around the proxy usage.

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -88,23 +92,70 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
...
+      } else if (prop === 'hasBackend') {
+        return target.hasOwnProperty('hasBackend') && target.hasBackend;
+      } else {
+        throw new Error(`Property ${String(prop)} not found in user's prompt.`);
+      }
+    },
+  });
+}
+
+export function getTemplateData(proxy: any): TemplateData {
```
leventebalogh, line 141: I think it would be nice to also type the input parameter of `getTemplateData()` (instead of using `any`). I think what would be the most straightforward is if the function would receive a typed object that explicitly defines what input information we need to generate the template data.

packages/create-plugin/src/types.ts
```diff
@@ -12,6 +12,13 @@ export type CliArgs = {
 
 export type TemplateData = {
   pluginId: string;
+  pluginName: string;
+  pluginDescription: string;
+  hasBackend: boolean;
+  orgName: string;
+  pluginType: PLUGIN_TYPES;
+  hasGithubWorkflows?: boolean;
+  hasGithubLevitateWorkflow?: boolean;
```
leventebalogh, line 21: I think we probably shouldn't have optional properties in the TemplateData type. Maybe I am wrong, but I think looking at it from the template's perspective it's better if the available data properties always have a value - and they are set to a sane default value by `getTemplateData()` in case they are not set by the user prompt.

packages/create-plugin/src/utils/utils.templates.ts
Ukochka: I have used Proxy, because we have two different types of objects (CliArgs and json) which we want to use as an argument of getTemplateData. And each object has different names for the same property, for example at CliArgs we have `pluginId` and in json we have just an `id`. We can update objects to have the same naming although the feature scope would explode
In case of overrides the original object will be modified and it will be harder to understand what is going on in the future. The mapping will not be explicit. In case of Proxy we can see which properties are there
If we decide to unify naming at some point, we will do so, delete Proxy and rest of the code will stay the same

leventebalogh: I think we could just static-type the input of `getTemplateData()` and the callsite could take care of passing in data in the right format (and doing the data-mapping necessary). I would personally find that simpler to scan, and also the "knowledge" of call-site specific naming (CliArgs and json) would stay closer to where they are created.

jackw: I think I agree with Levi on this one. The proxy usage seems a bit much to me compared to a function that maybe takes in CliArgs else gets the pluginJson file then does the work to "evolve" the information we need into the object thats expected to be passed into `getTemplateData`.

packages/create-plugin/src/utils/utils.templates.ts
jackw: Let's not do this in this PR. It's been tackled [in this PR](https://github.com/grafana/plugin-tools/pull/834).

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -13,11 +13,16 @@ import {
...
+import { CliArgs, TemplateData } from '../types.js';
+import {
+  getPackageManagerInstallCmd,
+  getPackageManagerWithFallback,
+  getPackageManagerFromUserAgent,
+} from './utils.packageManager.js';
 import { getExportFileName } from '../utils/utils.files.js';
 import { getVersion } from './utils.version.js';
 import { FeatureFlags, getConfig } from './utils.config.js';
+import { get } from 'node:http';
```
leventebalogh, line 25: Are we using this import somewhere?

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -88,40 +93,66 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
...
+      isAppType: pluginJson.type === PLUGIN_TYPES.app || pluginJson.type === PLUGIN_TYPES.scenes,
+      isNPM: packageManagerName === 'npm',
+      version: currentVersion,
+      bundleGrafanaUI: features.bundleGrafanaUI ?? DEFAULT_FEATURE_FLAGS.bundleGrafanaUI,
+      useReactRouterV6,
+      reactRouterVersion: useReactRouterV6 ? '6.22.0' : '5.2.0',
+      usePlaywright,
+      e2eTestCmd,
+    };
+  }
```
leventebalogh, line 154: Looks like there are is some redundant logic in the two branches of the if-else, that we could unify (e.g. `reactRouterVersion`, etc.).

What if we would only set a few properties of the template data using the arguments? Instead of the if-else structure here I would probably build a single object and rely on the data object argument for each specific property.

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -96,10 +96,10 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
 export function getTemplateData(data?: Partial<TemplateData>): TemplateData {
   const { features } = getConfig();
   const currentVersion = getVersion();
-  const pluginJson = getPluginJson();
-  const pluginType = data?.pluginType ?? pluginJson.type;
-  const packageManagerInfo = getPackageManagerWithFallback();
-  const packageManagerName = data?.packageManagerName ?? packageManagerInfo.packageManagerName;
+  const pluginJson = data ? undefined : getPluginJson();
```
leventebalogh, line 98: Hmm, I am not sure that we should do this to be honest - like this we remove the possibility to only specify (override) certain template variables using the `data` param (now in case the call-site only wants to specify the `pluginName`, the other properties will fall back to `pluginJson?.`, which will always be undefined).

packages/create-plugin/src/utils/utils.templates.ts
```diff
@@ -96,10 +96,10 @@ export function renderTemplateFromFile(templateFile: string, data?: any) {
...
+  const pluginType = data?.pluginType ?? pluginJson?.type;
+  const packageManagerInfo = data ? getPackageManagerFromUserAgent() : getPackageManagerWithFallback();
```
leventebalogh, line 100: I think it was fine to leave the logic of `getPackageManagerFromUserAgent()` at the generate command - there can be cases when the `data` is provided yet it's not a generate command (e.g. in tests).

packages/create-plugin/src/utils/utils.templates.ts
Ukochka: When it is just `const pluginJson = getPluginJson();` and we create a new plugin it is failing because it is executing getPluginJson function anyway and can not find it. And I see the error "Make sure you run this command from a plugins root directory."

packages/create-plugin/src/utils/utils.templates.ts
Ukochka: Same is here - when it is just `const packageManagerInfo =  getPackageManagerWithFallback();` it is failing with the message "Make sure you run this command from a plugins root directory."

packages/create-plugin/src/utils/utils.templates.ts
leventebalogh: In that case we should try fetching the plugin json here in a resilient way (with a fallback). Although these changes indirectly fix this problem of the not existing plugin json, they now make it necessary for the call-site to specify all the attributes in the data (as if they override a single one, the pluginJson will be set to undefined and they cannot rely on it as a default anymore).

packages/create-plugin/src/utils/utils.templates.ts
leventebalogh: (Resolving this, as the proxy got removed.)

packages/create-plugin/src/utils/utils.templates.ts
leventebalogh: (Resolving this, as the proxy got removed.)

packages/create-plugin/src/utils/utils.templates.ts
leventebalogh: After thinking about this for a few more times, I now actually think that I was wrong suggesting this function to be more generic, and accept a partial of the `TemplateData` types. (Making it generic in a sense that it can accept an override of even a single property made it quite complicated, mostly due to the fact that the plugin.json and package.json files are not always available as a fallback.)

Summing it up please ignore my comments above, and I will resolve this thread now.

packages/create-plugin/src/utils/utils.templates.ts
leventebalogh: Yeah, you are actually right, handling these errors differently (in case it's not run from a plugin directory) would make the code quite complicated, and it's probably also unnecessary at this point.

leventebalogh (approved):
LGTM, nice work on this @Ukochka 👏

PR #855, opened 2024-04-03
Create plugin: Add e2e specs

sunker:
This PR adds plugin type specific specs to the scaffolding. There's specific specs for...
* panels
* backend datasources
* non-backend datasources (different approach to e2e is necessary when testing for example health checks)
* apps (I believe specs would be the same for app with and without a backend). 

Fixes #


packages/create-plugin/templates/backend/pkg/plugin/datasource.go
```diff
@@ -94,16 +94,23 @@ func (d *Datasource) query(_ context.Context, pCtx backend.PluginContext, query
...
 // a datasource is working as expected.
 func (d *Datasource) CheckHealth(_ context.Context, req *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
-	var status = backend.HealthStatusOk
-	var message = "Data source is working"
+	res := &backend.CheckHealthResult{}
+	config, err := models.LoadPluginSettings(*req.PluginContext.DataSourceInstanceSettings)
 
-	if rand.Int()%2 == 0 {
-		status = backend.HealthStatusError
-		message = "randomized error"
```
sunker, line 102: Replacing this randomized error with something that is consistent so that e2e tests can be added for the health check.

packages/create-plugin/templates/app/src/components/AppConfig/AppConfig.tsx
```diff
@@ -50,6 +50,7 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
         <Field label="API Key" description="A secret key for authenticating to our custom API">
           <SecretInput
             width={60}
+            id="config-api-key"
```
sunker, line 53: ..so that input field becomes associated with the label field.

packages/create-plugin/templates/datasource/src/datasource.ts
```diff
@@ -1,3 +1,4 @@
+import { getBackendSrv, isFetchError } from '@grafana/runtime';
```
sunker, line 1: More of less copied from datasource-http example

packages/create-plugin/templates/app/src/components/AppConfig/AppConfig.tsx
jackw: For my own understanding - is this required for playwright to find the input field?

packages/create-plugin/templates/datasource/src/datasource.ts
```diff
@@ -43,11 +49,46 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
...
+        };
+      } else {
+        return {
+          status: 'error',
+          message: response.statusText ? response.statusText : defaultErrorMessage,
+        };
+      }
+    } catch (err) {
+      let message = '';
+      if (_.isString(err)) {
```
jackw, line 80: Hmm... lodash feels a bit overkill for checking primitives. Is this how we do it in examples repo? frontend platform has been trying to remove/discourage usage of lodash in core for quite some time now.

packages/create-plugin/templates/common/_package.json
```diff
@@ -22,7 +22,7 @@
     "@grafana/e2e": "{{ grafanaVersion }}",
     "@grafana/e2e-selectors": "{{ grafanaVersion }}",{{/unless}}
     "@grafana/eslint-config": "^7.0.0",{{#if usePlaywright}}
-    "@grafana/plugin-e2e": "^0.24.0",{{/if}}
+    "@grafana/plugin-e2e": "^0.27.0",{{/if}}
```
jackw, line 25: I'm under the impression that we're not gonna release these changes until 1.0.1 is out the door?

jackw (commented):
Looks like a lot of changes to what is scaffolded though I think they're for the best as it feels like neater approaches to things like health checks.

packages/create-plugin/templates/app/src/components/AppConfig/AppConfig.tsx
sunker: You can find the input field using test id too, but it's preferred to test user-visible behavior if possible. So `await page.getByRole('textbox', { name: 'API Key' })` is preferred over `await page.getByTestId('config-api-key').`

Ofc this is not always possible. When accessing grafana ui elements outside of the plugin code, plugin-e2e users should use the e2e selectors. e.g `panelEditPage.getByGrafanaSelector(selectors.components.CodeEditor.container)`. There's more info regarding this in the [Select UI elements](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/selecting-elements) guide in the e2e docs.

packages/create-plugin/templates/datasource/src/datasource.ts
sunker: Copied from the examples repo. Yeah I don't like this either. Will try to fix here and update examples repo too

packages/create-plugin/templates/common/_package.json
sunker: No you're right! It's just that one of the specs depend on changes released in 0.27.0, so needed to bump to be able to test and verify what was generated. We should point to 1.0.1 before merging to main. 👍

PR #863, opened 2024-04-08
feat: check compatibility with min version

sympatheticmoose:
This is just an idea and not verified whether it actually works or not, especially substituting the version within the levitate call.

Its very brittle based on how people _could_ use this field, but may serve as a way to check against whether recent changes break older compatibility.


sympatheticmoose:
note to self: missing  `jq -r '.dependencies.grafanaDependency'`

Also copying over from slack w @leventebalogh 
>I think instead of bash maybe we could use javascript in the step's script, and then we could probably more easily parse the semver string (if it's a semver string) to get the min and max versions?
That way we could we could tell if the plugin is compatible with the provided grafana dependency range.
(Latest is maybe outside of that range - however it can still be a useful information if it would break with latest)

PR #877, opened 2024-04-12
Docs: Update the README with instructions for redirects

josmperez:
Update the README in docusaurus/website directory to include detailed instructions for creating and testing client-side redirects.

### DIFF
--- a/docusaurus/website/README.md
+++ b/docusaurus/website/README.md
@@ -32,3 +32,11 @@ Search functionality is handled by a "local search" index built using (docusauru
 $ npm run build
 $ npm run serve
 ```
+
+### Redirects
+
+When moving a file, create a client-side redirect so as not to break links from the old location. It is a best practice to also manually update old links to the new location whenever possible.
+
+Insert the client-side redirect here: /plugin-tools/docusaurus/website/docusaurus.config.dev.js
+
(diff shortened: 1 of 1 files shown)


docusaurus/website/README.md
```diff
@@ -32,3 +32,11 @@ Search functionality is handled by a "local search" index built using (docusauru
...
 $ npm run serve
 ```
+
+### Redirects
+
+When moving a file, create a client-side redirect so as not to break links from the old location. It is a best practice to also manually update old links to the new location whenever possible.
+
+Insert the client-side redirect here: /plugin-tools/docusaurus/website/docusaurus.config.dev.js
+
+To test that the redirect works prior to publishing the PR, do this blah blah blah...
```
tolzhabayev, line 42:
```suggestion
To test that the redirect works prior to publishing the PR deploy your branch to DEV Stage of developer portal. You can do so by [running this action](https://github.com/grafana/plugin-tools/actions/workflows/deploy-to-developer-portal-dev.yml) called `Deploy to Developer Portal DEV Bucket`.

Hit the `run workflow` button and choose following parameters:
- Use workflow from: Branch `main` (keep it as it is by default)
- Which branch to use? Here you should enter your branch name.
```

tolzhabayev (approved):
@josmperez I think this can be merged?

PR #881, opened 2024-04-17
Docs: Update the docs for UI-Extensions

leventebalogh:
### What changed?
With [making the extensions registry reactive](https://github.com/grafana/grafana/pull/83085), we need to update the docs as well to give information about the React hooks and what other implications the change might causes.

[View it on dev &rarr;](https://grafana-dev.com/developers/plugin-tools/ui-extensions/create-an-extension-point)

### DIFF
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -39,9 +39,201 @@ When you create an extension point in a plugin, you create a public interface fo
 
 You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
 
+### `usePluginExtensions()`
+
+:::note
+
+**Performance tip:** in case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
+
+:::
+
+The `usePluginExtensions` react hooks returns the list of extensions that are registered for a certain extension point ID. The hook will dynamically update its return value when the list of extensions changes - this usually happens when extensions are registered during runtime due to dynamic plugin loading.
+
+#### Syntax
+
+```tsx
+usePluginExtensions(options);
+usePluginLinkExtensions(options); // Only returns extensions that have type `type="link"`
+usePluginComponentExtensions(options); // Only returns extensions that have type `type="component"`
+```
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`, for example: `plugins/myorg-super-app`.
+
+##### `options?.context` - _object (Optional)_
+
+An arbitrary object, that contains information related to your extension point which you would like to share with the extensions, for example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, for those you can pass contextual information using the component props. **Note:** the provided context object always gets frozen (turned immutable) before being shared with the extensions.
+
+##### `options?.limitPerPlugin` - _number (Optional)_
+
+It can be used to limit maximum how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
+
+#### Return value
+
+The hooks return an object in the following format:
+
+```tsx
+// usePluginExtensions()
+{
+  isLoading: boolean;
+  extensions: Array<PluginExtensionLink | PluginExtensionComponent>
+}
+
+// usePluginLinkExtensions()
+{
+  isLoading: boolean;
+  extensions: PluginExtensionLink[]
+}
+
+// usePluginComponentExtensions()
+{
+  isLoading: boolean;
+  extensions: PluginExtensionComponent[]
+}
+```
+
+(For more information check the type definitions of [`PluginExtensionLink`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L27) and [`PluginExtensionComponent`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L35C13-L35C37).)
+
+#### Example - rendering link extensions (static context)
+
+In the following example we render a link component all link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
+
+```tsx
+import { usePluginLinkExtensions } from '@grafana/runtime';
+
+// We are defining the `context` outside of the React component for performance reasons.
+// (In case we would declare it inside the component below, then it would result in a new object on every render,
+// which would unnecessarily trigger the `usePluginLinkExtensions()` hook.)
+const context = {
+  referenceId: '12345',
+  timeZone: 'UTC',
+};
+
+function AppMenuExtensionPoint() {
+  // This only returns type="link" extensions
+  const { extensions } = usePluginLinkExtensions({
+    extensionPointId: 'plugins/another-app-plugin/menu',
+    context,
+  });
+
+  if (extensions.length === 0) {
+    return null;
+  }
+
+  return (
+    <div>
+      {extensions.map((extension) => {
+        return (
+          <a href={extension.path} onClick={extension.onClick} title={extension.description} key={extension.key}>
+            {extension.title}
+          </a>
+        );
+      })}
+    </div>
+  );
+}
+```
+
+#### Example - rendering link extensions (dynamic context)
+
+In this example we are creating the context object dynamically. Although this is a common practice, we have to be aware, that the `usePluginLinkExtensions()` hook will rerender in the following scenarios:
+
+- IF the `context` object changes _(so the extensions can react to the context changes)_
+- IF the extension-registry changes
+
+Because of these we have to make sure, that we only change the `context` object if its content changes, otherwise we will cause unnecessary rerenders. The following example shows how to approach these scenarios in a safe way:
+
+```tsx
+import { useMemo } from 'react';
+import { usePluginLinkExtensions } from '@grafana/runtime';
+
+function AppMenuExtensionPoint({ referenceId }) {
+  // Instead of just defining the object here (which would result in a new object on every render),
+  // we use `useMemo()` to only update the context object when its "dynamic" dependencies change.
+  const context = useMemo(
+    () => ({
+      referenceId,
+      timeZone: 'UTC',
+    }),
+    [referenceId]
+  );
+  const { extensions } = usePluginLinkExtensions({
+    extensionPointId: 'plugins/another-app-plugin/menu',
+    context,
+  });
+
+  if (extensions.length === 0) {
+    return null;
+  }
+
+  return (
+    <div>
+      {extensions.map((extension) => {
+        return (
+          <a href={extension.path} onClick={extension.onClick} title={extension.description} key={extension.key}>
+            {extension.title}
+          </a>
+        );
+      })}
+    </div>
+  );
+}
+```
+
+#### Example - rendering component extensions
+
+:::note
+
+**Available in Grafana >=10.1.0** <br /> (_Component type extensions are only available in Grafana 10.1.0 and above._)
+
+:::
+
+Component type extensions are simple React components, which gives much more freedom on what they are able to do. You can pass contextual information to the extension components using props.
+
+```tsx title="src/components/Toolbar.tsx"
+import { usePluginComponentExtensions } from '@grafana/runtime';
+
+export const Toolbar = () => {
+  // This only returns type="component" extensions
+  // Heads up! We don't specify a context object below, we pass in the contextual information as a prop to the component later.
+  const { extensions } = usePluginComponentExtensions({ extensionPointId: '<extension-point-id>' });
+
+  return (
+    <div>
+      <div className="title">Title</div>
+      <div className="extensions">
+        {/* Loop through the available extensions */}
+        {extensions.map((extension) => {
+          const Component = extension.component as React.ComponentType<{
+            version: string;
+          }>;
+
+          // Render extension component and pass contextual information (version)
+          return (
+            <div key={extension.id}>
+              <Component version="1.0.0" />
+            </div>
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+```
+
 ### `getPluginExtensions()` - _deprecated_
 
 The `getPluginExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+
+:::note
+
+This function only returns the state of the extensions registry (the extensions registered by plugins) at a given time. If there are extensions registered by plugins after that point in time, you won't receive them. <br />
+**We strongly suggest to use the reactive [`usePluginExtensions()`](#usepluginextensions) hook instead wherever possible.**
+
+:::
 
 #### Syntax
 
@@ -138,13 +330,13 @@ export const Toolbar = () => {
         {/* Loop through the available extensions */}
         {extensions.map((extension) => {
           const Component = extension.component as React.ComponentType<{
-            context: { version: string };
+            version: string;
           }>;
 
           // Render extension component and pass contextual information (version)
           return (
             <div key={extension.id}>
-              <Component context={{ version }} />
+              <Component version="1.0.0" />
             </div>
           );
         })}
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -71,7 +71,7 @@ The unique identifier of your extension point. It must begin with `plugins/<PLUG
 
 ##### `options?.context` - _object (Optional)_
 
-As an arbitrary object, it contains information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions because you can pass contextual information using the component props. 
+As an arbitrary object, it contains information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions because you can pass contextual information using the component props.
 
 :::note
 
@@ -199,11 +199,11 @@ function AppMenuExtensionPoint({ referenceId }) {
 
 :::note
 
-**This feature is only available in Grafana versions 10.1.0 and above.
+\*\*This feature is only available in Grafana versions 10.1.0 and above.
 
 :::
 
-Component type extensions are simple React components, which gives much more freedom on what they are able to do. You can pass contextual information to the extension components using props.
+Component type extensions are simple React components, which gives you much more freedom in what you can make them do. You can pass contextual information to the extension components using props.
 
 ```tsx title="src/components/Toolbar.tsx"
 import { usePluginComponentExtensions } from '@grafana/runtime';
@@ -263,9 +263,9 @@ The unique identifier of your extension point. It must begin with `plugins/<PLUG
 
 ##### `options?.context` - _object (Optional)_
 
-As an arbitrary object, it contains information related to your extension point which you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. 
+As an arbitrary object, it contains information related to your extension point which you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`.
 
-This parameter is not available for component extensions; for those, you can pass contextual information using the component props. 
+This parameter is not available for component extensions; for those, you can pass contextual information using the component props.
 
 :::note
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -15,7 +15,7 @@ keywords:
   - apps
 ---
 
-In this guide you will learn how to provide an extension point so that other app plugins can add their extensions to your plugin.
+In this guide you will learn how to provide an extension point so that app plugins can add their extensions to your plugin.
 
 ## What is an extension point?
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -163,7 +163,7 @@ import { useMemo } from 'react';
 import { usePluginLinkExtensions } from '@grafana/runtime';
 
 function AppMenuExtensionPoint({ referenceId }) {
-  // Instead of just defining the object here (which would result in a new object on every render),
+  // Instead of defining the object here (which would result in a new object on every render),
   // we use `useMemo()` to only update the context object when its "dynamic" dependencies change.
   const context = useMemo(
     () => ({
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -15,7 +15,6 @@ keywords:
   - apps
 ---
 
-
 In this guide you will learn how to provide an extension point so that other app plugins can add their extensions to your plugin.
 
 ## What is an extension point?
@@ -52,7 +51,7 @@ You can easily create an extension point using the following functions (they liv
 
 :::note
 
-In case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
+In case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#example---rendering-link-extensions-dynamic-context)
 
 :::
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -1,6 +1,6 @@
 ---
 id: create-an-extension-point
-title: Create an extension point in your app plugin
+title: Create an extension point
 sidebar_label: Create an extension point
 sidebar_position: 2
 description: Learn how to provide an extension point so that other applications can contribute their extensions.
@@ -15,19 +15,25 @@ keywords:
   - apps
 ---
 
-Providing an extension point within your UI enables other plugins to contribute supplementary capabilities, and leverage the context within the current view to guide users to a potential next step. Here's what you need to know about extension points.
+## What is an extension point?
 
-Firstly, you will need to define an extension point ID. This is basically just a string describing the part of the UI where the extension point lives. Extension developers should be able to figure out where in the UI the extension will be added by reading the extension point ID.
+An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can be either be links or simple React components (that can basically impelement anything). It's up to you how you are displaying the handling or displaying the extensions, we have some examples below.
+
+You can also share contextual information with the extensions using a `context` object or component props (see the examples below).
+
+### Requirements
+
+- **an extension point ID** _(string)_
+  - In case it's an extension point in core Grafana, it must start with `grafana/`
+  - In case it's inside a plugin, it must start with `plugin/<PLUGIN_ID>/`
+  - It must be unique
+- **an app plugin** - _currently only app plugins can create extension points_
 
 :::note
 
-Extension points living in core Grafana must start with `grafana/` and extension points living in plugins must have IDs starting with `plugins/` followed by the plugin ID, for example, `plugins/<PLUGIN_ID>/`.
+Consider the design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
 
 :::
-
-The second thing you need to consider is how to design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI.
-
-Finally, consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
 
 ## Create an extension point
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -31,13 +31,13 @@ Finally, consider if there is any information from the current view that should 
 
 ## Create an extension point
 
-Use the `getPluginLinkExtensions` method in `@grafana/runtime` to create an extension point within your plugin.
-
 :::danger
 
 When you create an extension point in a plugin, you create a public interface for other plugins to interact with. Changes to the extension point ID or its context could break any plugin that attempts to register a link inside your plugin.
 
 :::
+
+Use the `getPluginLinkExtensions` method in `@grafana/runtime` to create an extension point within your plugin.
 
 The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -200,12 +200,6 @@ function AppMenuExtensionPoint({ referenceId }) {
 
 #### Example - rendering component extensions
 
-:::note
-
-\*\*This feature is only available in Grafana versions 10.1.0 and above.
-
-:::
-
 Component type extensions are simple React components, which gives you much more freedom in what you can make them do. You can pass contextual information to the extension components using props.
 
 ```tsx title="src/components/Toolbar.tsx"
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -81,7 +81,7 @@ The provided context object always gets frozen (turned immutable) before being s
 
 ##### `options?.limitPerPlugin` - _number (Optional)_
 
-Use thie parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
+Use this parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
 
 #### Return value
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -31,7 +31,7 @@ You can also share contextual information with the extensions using a `context` 
   - In case it's an extension point in core Grafana, it must start with `grafana/`
   - In case it's inside a plugin, it must start with `plugin/<PLUGIN_ID>/`
   - It must be unique
-- **an app plugin** - _currently only app plugins can create extension points_
+- **an app plugin** - _apart from core Grafana, currently only app plugins can create extension points and register extensions._
 
 :::note
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -57,7 +57,7 @@ In case the `context` object is created dynamically, make sure to wrap it into a
 
 :::
 
-The `usePluginExtensions` React hooks returns the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
+The `usePluginExtensions` React hooks return the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
 
 #### Syntax
 
--- a/docusaurus/docs/ui-extensions/create-an-extension-point.md
+++ b/docusaurus/docs/ui-extensions/create-an-extension-point.md
@@ -19,7 +19,7 @@ keywords:
 
 An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can be either be links or simple React components (that can basically impelement anything). It's up to you how you are displaying the handling or displaying the extensions, we have some examples below.
 
-You can also share contextual information with the extensions using a `context` object or component props (see the examples below).
+You can also share contextual information with the extensions using a `context` object or component props (see the following examples).
 
 ### Requirements
 
@@ -31,7 +31,7 @@ You can also share contextual information with the extensions using a `context` 
 
 :::note
 
-Consider the design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
+Consider the design of the UI of the extension point so it supports a scenario where multiple extensions can be added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the extension's functionality.
 
 :::
 
@@ -45,15 +45,15 @@ When you create an extension point in a plugin, you create a public interface fo
 
 You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
 
-### `usePluginExtensions()`
+### The `usePluginExtensions()` hook
 
 :::note
 
-**Performance tip:** in case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
+In case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
 
 :::
 
-The `usePluginExtensions` react hooks returns the list of extensions that are registered for a certain extension point ID. The hook will dynamically update its return value when the list of extensions changes - this usually happens when extensions are registered during runtime due to dynamic plugin loading.
+The `usePluginExtensions` React hooks returns the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
 
 #### Syntax
 
@@ -67,15 +67,21 @@ usePluginComponentExtensions(options); // Only returns extensions that have type
 
 ##### `options.extensionPointId` - _string_
 
-The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`, for example: `plugins/myorg-super-app`.
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
 
 ##### `options?.context` - _object (Optional)_
 
-An arbitrary object, that contains information related to your extension point which you would like to share with the extensions, for example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, for those you can pass contextual information using the component props. **Note:** the provided context object always gets frozen (turned immutable) before being shared with the extensions.
+As an arbitrary object, it contains information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions because you can pass contextual information using the component props. 
+
+:::note
+
+The provided context object always gets frozen (turned immutable) before being shared with the extensions.
+
+:::
 
 ##### `options?.limitPerPlugin` - _number (Optional)_
 
-It can be used to limit maximum how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
+Use thie parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
 
 #### Return value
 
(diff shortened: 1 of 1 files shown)


docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
   - apps
 ---
 
-Providing an extension point within your UI enables other plugins to contribute supplementary capabilities, and leverage the context within the current view to guide users to a potential next step. Here's what you need to know about extension points.
+## What is an extension point?
```
josmperez, line 18: When the H1 (# Create an extension point) is added above this H2, it creates stacked headers which Grafana docs try to avoid. I suggest no H2  or adding a sentence above the H2 to introduce the topic as a whole.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 
-Firstly, you will need to define an extension point ID. This is basically just a string describing the part of the UI where the extension point lives. Extension developers should be able to figure out where in the UI the extension will be added by reading the extension point ID.
+An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can be either be links or simple React components (that can basically impelement anything). It's up to you how you are displaying the handling or displaying the extensions, we have some examples below.
```
josmperez, line 20:
```suggestion
An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can either be links or React components that can impelement virtually anything. It's up to you how to display the handling or displaying of extensions. Refer to the following examples below:
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 
-:::note
+You can also share contextual information with the extensions using a `context` object or component props (see the examples below).
```
josmperez, line 22:
```suggestion
You can also share contextual information with the extensions using a `context` object or component props (see the following examples).
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 
-Extension points living in core Grafana must start with `grafana/` and extension points living in plugins must have IDs starting with `plugins/` followed by the plugin ID, for example, `plugins/<PLUGIN_ID>/`.
+### Requirements
 
-:::
+- **an extension point ID** _(string)_
+  - In case it's an extension point in core Grafana, it must start with `grafana/`
+  - In case it's inside a plugin, it must start with `plugin/<PLUGIN_ID>/`
+  - It must be unique
+- **an app plugin** - _currently only app plugins can create extension points_
```
josmperez, line 30:
```suggestion
- App plugins - _currently only app plugins can create extension points_
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 
-The second thing you need to consider is how to design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI.
+:::note
 
-Finally, consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
+Consider the design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
```
josmperez, line 34:
```suggestion
Consider the design of the UI of the extension point so it supports a scenario where multiple extensions can be added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the extension's functionality.
```

josmperez: Grafana docs prefer sentences < 25 words in length.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 :::
 
-The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
+
+### `usePluginExtensions()`
+
+:::note
+
+**Performance tip:** in case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
```
josmperez, line 52:
```suggestion
In case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 When you create an extension point in a plugin, you create a public interface for other plugins to interact with. Changes to the extension point ID or its context could break any plugin that attempts to register a link inside your plugin.
 
 :::
 
-The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
+
+### `usePluginExtensions()`
+
+:::note
```
josmperez, line 50:
```tip
:::note
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
 :::danger
 
 When you create an extension point in a plugin, you create a public interface for other plugins to interact with. Changes to the extension point ID or its context could break any plugin that attempts to register a link inside your plugin.
 
 :::
 
-The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
+
+### `usePluginExtensions()`
```
josmperez, line 48:
```suggestion
### The `usePluginExtensions()` hook
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+:::note
+
+**Performance tip:** in case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
+
+:::
+
+The `usePluginExtensions` react hooks returns the list of extensions that are registered for a certain extension point ID. The hook will dynamically update its return value when the list of extensions changes - this usually happens when extensions are registered during runtime due to dynamic plugin loading.
```
josmperez, line 56:
```suggestion
The `usePluginExtensions` React hooks returns the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+usePluginExtensions(options);
+usePluginLinkExtensions(options); // Only returns extensions that have type `type="link"`
+usePluginComponentExtensions(options); // Only returns extensions that have type `type="component"`
+```
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`, for example: `plugins/myorg-super-app`.
```
josmperez, line 70:
```suggestion
The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+##### `options?.context` - _object (Optional)_
+
+An arbitrary object, that contains information related to your extension point which you would like to share with the extensions, for example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, for those you can pass contextual information using the component props. **Note:** the provided context object always gets frozen (turned immutable) before being shared with the extensions.
```
josmperez, line 74:
```suggestion
As an arbitrary object, it contains information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions because you can pass contextual information using the component props. 

:::note

The provided context object always gets frozen (turned immutable) before being shared with the extensions.

:::
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+##### `options?.limitPerPlugin` - _number (Optional)_
+
+It can be used to limit maximum how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```
josmperez, line 78:
```suggestion
Use thie parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+  isLoading: boolean;
+  extensions: PluginExtensionComponent[]
+}
+```
+
+(For more information check the type definitions of [`PluginExtensionLink`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L27) and [`PluginExtensionComponent`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L35C13-L35C37).)
+
+#### Example - rendering link extensions (static context)
+
+In the following example we render a link component all link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
```
josmperez, line 108:
```suggestion
The following example shows how to render a link component all link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
```

josmperez: The above sentence is unclear ("render a link component all link-type...") but I'm not sure how to fix it.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+        );
+      })}
+    </div>
+  );
+}
+```
+
+#### Example - rendering link extensions (dynamic context)
+
+In this example we are creating the context object dynamically. Although this is a common practice, we have to be aware, that the `usePluginLinkExtensions()` hook will rerender in the following scenarios:
```
josmperez, line 148:
```suggestion
The following example shows how to create the context object dynamically. Although this is a common practice, you should be aware that the `usePluginLinkExtensions()` hook will re-render in the following scenarios:
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+- IF the `context` object changes _(so the extensions can react to the context changes)_
```
josmperez, line 150:
```suggestion
- If the `context` object changes _(so the extensions can react to the context changes)_
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+- IF the extension-registry changes
```
josmperez, line 151:
```suggestion
- If the extension-registry changes
```

josmperez: Only italics for emphasis if absolutely necessary.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+Because of these we have to make sure, that we only change the `context` object if its content changes, otherwise we will cause unnecessary rerenders. The following example shows how to approach these scenarios in a safe way:
```
josmperez, line 153:
```suggestion
Be sure to only change the `context` object if its content changes; otherwise, you could create unnecessary re-renders. The following example shows how to approach these scenarios in a safe way:
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+    </div>
+  );
+}
+```
+
+#### Example - rendering component extensions
+
+:::note
+
+**Available in Grafana >=10.1.0** <br /> (_Component type extensions are only available in Grafana 10.1.0 and above._)
```
josmperez, line 196:
```suggestion
**This feature is only available in Grafana versions 10.1.0 and above.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+:::
+
+Component type extensions are simple React components, which gives much more freedom on what they are able to do. You can pass contextual information to the extension components using props.
```
josmperez, line 200:
```suggestion
Component type extensions are simple React components, which gives you much more freedom in  what you can make them do. For example, you can pass contextual information to the extension components using props.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+            </div>
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+```
+
+### `getPluginExtensions()` - _deprecated_
```
josmperez, line 233:
```suggestion
### The `getPluginExtensions()` method - _deprecated_
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+The `getPluginExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+
+:::note
+
+This function only returns the state of the extensions registry (the extensions registered by plugins) at a given time. If there are extensions registered by plugins after that point in time, you won't receive them. <br />
+**We strongly suggest to use the reactive [`usePluginExtensions()`](#usepluginextensions) hook instead wherever possible.**
```
josmperez, line 240:
```suggestion
As a best practice, use the reactive [`usePluginExtensions()`](#usepluginextensions) hook instead wherever possible.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+getPluginExtensions(options);
+getPluginLinkExtensions(options); // Only returns extensions that have type `type="link"`
+getPluginComponentExtensions(options); // Only returns extensions that have type `type="component"`
+```
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`, for example: `plugins/myorg-super-app`.
```
josmperez, line 256:
```suggestion
The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+##### `options?.context` - _object (Optional)_
+
+An arbitrary object, that contains information related to your extension point which you would like to share with the extensions, for example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, for those you can pass contextual information using the component props. **Note:** the provided context object always gets frozen (turned immutable) before being shared with the extensions.
```
josmperez, line 260:
```suggestion
As an arbitrary object, it contains information related to your extension point which you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. 

This parameter is not available for component extensions; for those, you can pass contextual information using the component props. 

:::note

The provided context object always gets frozen (turned immutable) before being shared with the extensions.

:::
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+##### `options?.limitPerPlugin` - _number (Optional)_
```
josmperez, line 262:
```suggestion
##### The `options?.limitPerPlugin` method - _number (Optional)_
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,271 @@ keywords:
...
+
+It can be used to limit maximum how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```
josmperez, line 264:
```suggestion
Use this method to specify the maximum amount of extensions that should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```

josmperez (approved):
I have reviewed this from a documentation standpoint only, and have made various suggestions. No blocking issues identified.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: I guess this suggestion is maybe a typo? (This is one of the option properties). I will resolve it, please reopen if you have objections!

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Thanks! I have fixed it in https://github.com/grafana/plugin-tools/pull/881/commits/a5821d379642c396fdd2163dcecb71afb89a7e9c, however I left out the `For example` part (I think this is more like an additional information than an example), please let me know if you disagree.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
+
+##### `options?.context` - _object (Optional)_
+
+As an arbitrary object, it contains information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions because you can pass contextual information using the component props.
```
jackw, line 74:
```suggestion
An object containing information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, instead you should pass contextual information using the component props.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+  - In case it's an extension point in core Grafana, it must start with `grafana/`
+  - In case it's inside a plugin, it must start with `plugin/<PLUGIN_ID>/`
+  - It must be unique
+- **an app plugin** - _currently only app plugins can create extension points_
 
-The second thing you need to consider is how to design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI.
+:::note
 
-Finally, consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
+Consider the design of the UI of the extension point so it supports a scenario where multiple extensions can be added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the extension's functionality.
```
jackw, line 34:
```suggestion
When designing the UI make sure the extension point supports a scenario where multiple extensions can be added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin pre-fill values or other data in the extension's functionality.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+
+:::note
+
+The provided context object always gets frozen (turned immutable) before being shared with the extensions.
+
+:::
+
+##### `options?.limitPerPlugin` - _number (Optional)_
+
+Use thie parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```
jackw, line 84:
```suggestion
Use this parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+  isLoading: boolean;
+  extensions: PluginExtensionComponent[]
+}
+```
+
+(For more information check the type definitions of [`PluginExtensionLink`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L27) and [`PluginExtensionComponent`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L35C13-L35C37).)
+
+#### Example - rendering link extensions (static context)
+
+The following example shows how to render a link component all link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
```
jackw, line 114:
```suggestion
The following example shows how to render a link component as link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+
+```tsx
+import { usePluginLinkExtensions } from '@grafana/runtime';
+
+// We are defining the `context` outside of the React component for performance reasons.
+// (In case we would declare it inside the component below, then it would result in a new object on every render,
+// which would unnecessarily trigger the `usePluginLinkExtensions()` hook.)
```
jackw, line 121:
```suggestion
// We define the `context` outside of the React component for performance reasons.
// (Declaring it inside the component would result in a new object on every render,
// which would unnecessarily trigger the `usePluginLinkExtensions()` hook.)
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Hmm, I am not sure about this one - but maybe I am wrong. When we are listing the requirements, wouldn't `an app plugin` or `app plugin` sounds better than `App plugins` (in plural)?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+
+Be sure to only change the `context` object if its content changes; otherwise, you could create unnecessary re-renders. The following example shows how to approach these scenarios in a safe way:
+
+```tsx
+import { useMemo } from 'react';
+import { usePluginLinkExtensions } from '@grafana/runtime';
+
+function AppMenuExtensionPoint({ referenceId }) {
+  // Instead of just defining the object here (which would result in a new object on every render),
+  // we use `useMemo()` to only update the context object when its "dynamic" dependencies change.
```
jackw, line 167:
```suggestion
  // Instead of defining the object here (which would result in a new object on every render),
  // we use `useMemo()` to only update the context object when its "dynamic" dependencies change.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
...
+    </div>
+  );
+}
+```
+
+#### Example - rendering component extensions
+
+:::note
+
+\*\*This feature is only available in Grafana versions 10.1.0 and above.
```
jackw, line 202: Is this true? The example below uses `usePluginComponentExtensions` which I don't think is available in 10.1.0. 🤔

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: I'll resolve this - let me know if you would like any changes in this note block.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Hmm, should we keep `following` and the `:` in the last line? (I am asking because the examples are not directly following this sentence, but they are a few paragraphs below.) Maybe we should add a link instead?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,285 @@ keywords:
   - apps
 ---
 
-Providing an extension point within your UI enables other plugins to contribute supplementary capabilities, and leverage the context within the current view to guide users to a potential next step. Here's what you need to know about extension points.
+## What is an extension point?
```
jackw, line 18: I agree with Joe here. It's a bit funny to have stacked headers. Maybe we can do something like:

```suggestion

In this guide you will learn how to provide an extension point so that other app plugins can add their extensions to your plugin.

## What is an extension point?
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Thanks, that's a good point, I wasn't aware of this unfortunately 🙏 
I'll go with [Jack's suggestion](https://github.com/grafana/plugin-tools/pull/881#discussion_r1574263489) which adds a sentence above the H2, let me know what you think.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
   - apps
 ---
 
-Providing an extension point within your UI enables other plugins to contribute supplementary capabilities, and leverage the context within the current view to guide users to a potential next step. Here's what you need to know about extension points.
 
-Firstly, you will need to define an extension point ID. This is basically just a string describing the part of the UI where the extension point lives. Extension developers should be able to figure out where in the UI the extension will be added by reading the extension point ID.
+In this guide you will learn how to provide an extension point so that other app plugins can add their extensions to your plugin.
```
sympatheticmoose, line 18:
```suggestion
In this guide you will learn how to provide an extension point so that app plugins can add their extensions to your plugin.
```
nit: if the page targets Core developers

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
 
-:::note
+## What is an extension point?
 
-Extension points living in core Grafana must start with `grafana/` and extension points living in plugins must have IDs starting with `plugins/` followed by the plugin ID, for example, `plugins/<PLUGIN_ID>/`.
+An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can be either be links or simple React components (that can basically impelement anything). It's up to you how you are displaying the handling or displaying the extensions, we have some examples below.
```
sympatheticmoose, line 22: I'm not 100% sure we need to define it _again_, but not a blocking comment https://grafana.com/developers/plugin-tools/ui-extensions/#definition-of-terms

Building on @josmperez 's suggestion above and incorporating your response @leventebalogh 
```suggestion
An extension point is a place in the UI which you make extendable by other plugins using extensions. These extensions can either be links or React components that can implement virtually anything. 

In the extension point you can control how you want to handle the displaying of extensions to users. You can also share contextual information with the extensions using a `context` object or component props. Refer to the examples below to learn more.
```
An Examples subsection might be useful to break up the content and provide us a way to link from here. Note - I've also moved the content from below as we had two sections both ending with "see the examples"

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
 
-:::
+You can also share contextual information with the extensions using a `context` object or component props (see the following examples).
 
-The second thing you need to consider is how to design the UI of the extension point so it supports a scenario where multiple extensions are being added without breaking the UI.
+### Requirements
```
sympatheticmoose, line 26: Comment - requirements feels odd to me, we'd typically use this for things like Go version etc, there's other points below which might change how we want to title this section - can follow up.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
 
-Finally, consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin prefill values or other data in the functionality being added via the extension.
+- **an extension point ID** _(string)_
+  - In case it's an extension point in core Grafana, it must start with `grafana/`
+  - In case it's inside a plugin, it must start with `plugin/<PLUGIN_ID>/`
+  - It must be unique
+- **an app plugin** - _currently only app plugins can create extension points_
```
sympatheticmoose, line 32: What about Core developers who we'd direct this to as how to implement an extension point?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+When designing the UI make sure the extension point supports a scenario where multiple extensions can be added without breaking the UI. Also consider if there is any information from the current view that should be shared with the extensions added to the extension point. It could be information from the current view that could let the extending plugin pre-fill values or other data in the extension's functionality.
 
-Use the `getPluginLinkExtensions` method in `@grafana/runtime` to create an extension point within your plugin.
+:::
+
+## Create an extension point
 
 :::danger
 
 When you create an extension point in a plugin, you create a public interface for other plugins to interact with. Changes to the extension point ID or its context could break any plugin that attempts to register a link inside your plugin.
```
sympatheticmoose: The "Note" feels odd in the requirements, and the "Danger" is a bit off-putting as the start of the "Create an extension point" section.

What about a section like "Considerations for building an extension point" which could detail these points? Open to suggestions on alternative name

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
 
 :::
 
-The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
```
sympatheticmoose, line 48:
```suggestion
You can easily create an extension point using the following functions from `@grafana/runtime` to fetch extensions for a certain extension point ID:
```
Avoids the use of brackets

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+:::
+
+The `usePluginExtensions` React hooks returns the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
+
+#### Syntax
+
+```tsx
+usePluginExtensions(options);
+usePluginLinkExtensions(options); // Only returns extensions that have type `type="link"`
+usePluginComponentExtensions(options); // Only returns extensions that have type `type="component"`
```
sympatheticmoose, line 65: I personally find it odd that these are all under "usePluginExtensions()" subsection but I may be in the minority.

docusaurus/docs/ui-extensions/create-an-extension-point.md
sympatheticmoose: Its also somewhat brief as an introduction. "Create an extension point" sounds like I'm going to get a tutorial, but instead we've got more of an API doc with examples. Additionally we say functions plural, but then have 2 "sections" with one of them a deprecated method.

I don't object to the format, but I think we may need to expand this sentence. Happy to follow up with more suggestions if you and agree and would like some help with this

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
 :::
 
-The `getPluginLinkExtensions` method takes an object consisting of the `extensionPointId`, which must begin `plugins/<PLUGIN_ID>`, and any contextual information that you want to provide. The `getPluginLinkExtensions` method returns a list of extension links that your program can then loop over.
+You can easily create an extension point using the following functions (they live in `@grafana/runtime`) to fetch extensions for a certain extension point ID:
+
+### The `usePluginExtensions()` hook
+
+:::note
+
+In case the `context` object is created dynamically, make sure to wrap it into a `useMemo()` to prevent unnecesssary rerenders. [More info](#)
```
sympatheticmoose: > In case the `context` object is created dynamically,

Why would this be the case? Would describing an example help here?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+
+:::
+
+The `usePluginExtensions` React hooks returns the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
```
sympatheticmoose, line 58:
```suggestion
The `usePluginExtensions` React hooks return the list of extensions that are registered for a certain extension point ID. The hook dynamically updates its return value when the list of extensions changes. This behavior usually happens when extensions are registered during runtime due to dynamic plugin loading.
```

I'm not sure what's correct here... but if `hooks` is plural, return should be singular. The reason i'm not sure is that the section is `hook` singular, but the syntax lists 3 functions.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+usePluginExtensions(options);
+usePluginLinkExtensions(options); // Only returns extensions that have type `type="link"`
+usePluginComponentExtensions(options); // Only returns extensions that have type `type="component"`
+```
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
```
sympatheticmoose, line 72: what if Core?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+
+##### `options?.context` - _object (Optional)_
+
+An object containing information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, instead you should pass contextual information using the component props.
+
+:::note
+
+The provided context object always gets frozen (turned immutable) before being shared with the extensions.
```
sympatheticmoose, line 80:
```suggestion
The provided context object is made immutable before being shared with the extensions.
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: I think I'll remove this note for now - unfortunately it's not yet sure which official release the feature will land in.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+
+#### Parameters
+
+##### `options.extensionPointId` - _string_
+
+The unique identifier of your extension point. It must begin with `plugins/<PLUGIN_ID>`. For example: `plugins/myorg-super-app`.
+
+##### `options?.context` - _object (Optional)_
+
+An object containing information related to your extension point that you would like to share with the extensions. For example: `{ baseUrl: '/foo/bar' }`. This parameter is not available for component extensions, instead you should pass contextual information using the component props.
```
sympatheticmoose, line 76: > instead you should pass contextual information using the component props

We make this statement a lot... so a few questions:
- should we be providing advice to duplicate contextual information in both manners? If that is the case, why is it on the extension point developer? My worry is that it is up to the _extension_ developer to decide whether theres is a link or component, but by extending the same extension point they may get different context available, feels confusing.
- We reference it in the `Example - rendering component extensions` but maybe it is worth having an even smaller snippet which we could then link to from any time we make this statement?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+
+:::note
+
+The provided context object always gets frozen (turned immutable) before being shared with the extensions.
+
+:::
+
+##### `options?.limitPerPlugin` - _number (Optional)_
+
+Use this parameter to set the maximum value for how many extensions should be returned from the same plugin. It can be useful in cases when there is limited space on the UI to display extensions.
```
sympatheticmoose, line 86: It might be good to link to this from the section regarding designing your UI to respond to multiple plugins? But I think we also need to describe how it would be determined which plugins get displayed when there are more extensions than the limit allows

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+#### Example - rendering link extensions (static context)
+
+The following example shows how to render a link component as link-type extensions that other plugins registered for the `plugins/another-app-plugin/menu` extension point ID.
+
+```tsx
+import { usePluginLinkExtensions } from '@grafana/runtime';
+
+// We define the `context` outside of the React component for performance reasons.
+// (Declaring it inside the component would result in a new object on every render,
+// which would unnecessarily trigger the `usePluginLinkExtensions()` hook.)
```
sympatheticmoose, line 123: Do we need this detail in the docs?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+            {extension.title}
+          </a>
+        );
+      })}
+    </div>
+  );
+}
+```
+
+#### Example - rendering link extensions (dynamic context)
```
sympatheticmoose, line 154: I still find static vs dynamic a little unclear as to why the different situations occur.

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -15,40 +15,288 @@ keywords:
...
+            </div>
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+```
+
+### The `getPluginExtensions()` method - _deprecated_
```
sympatheticmoose, line 235: Do we want to keep this in the docs?

docusaurus/docs/ui-extensions/create-an-extension-point.md
```diff
@@ -101,6 +341,8 @@ Component type extensions are simple React components, which gives much more fre
 import { getPluginComponentExtensions } from '@grafana/runtime';
 
 export const Toolbar = () => {
+  // This only returns type="component" extensions
+  // Heads up! We don't specify a context object below, we pass in the contextual information as a prop to the component later.
```
sympatheticmoose, line 338: Could you help me understand why we're doing it this way and what we want to ensure the reader takes away having read this comment?

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: I think until it is actually removed, we should?

sympatheticmoose: The migration process is just to swap the hook right? (Which we could mention). 

I worry it's just noise that makes the doc look longer and more complicated. Especially given we don't want to encourage any further adoption, and we believe there's currently limited community uptake. 

I guess we need it until the core extension points are all converted. But after that I would personally just remove it.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Good one 👍

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: I actually like that idea, was thinking something similar (e.g. "Best practices"). I would probably go forward with this PR like this (to support the just-about-to-be-merged reactive extensions registry with docs), and update in an upcoming PR if that's fine with you.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Yeah, we could probably add more content here, let's try to make it better together 👍 

> but instead we've got more of an API doc with examples

True, although maybe if we start reorganising the structure of the docs soon ("Reference" and How to ..." are separate sections), then it will be good for the API reference.

> Additionally we say functions plural, but then have 2 "sections" with one of them a deprecated method.

Hmm, I guess as long as we don't remove the getter functions plural is legit?

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: No no, your suggestion is spot on (I meant talking about the three hooks, I just didn't want to type them out for making reading easier).

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Yeah, maybe you are right. I only put them here for now as their syntax, arguments and return values are basically the same and I wanted to avoid too much duplication. On the other hand if we move towards a direction of having an API reference, then we will need to separate them, you are right.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Good point, pushed a commit to fix it. 👍

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: > We reference it in the Example - rendering component extensions but maybe it is worth having an even smaller snippet which we could then link to from any time we make this statement?

Yeah, I think that's a good idea! How about creating that sub-section "Considerations..." or "Best practices" and adding this as a point under there?


> should we be providing advice to duplicate contextual information in both manners? If that is the case, why is it on the extension point developer? My worry is that it is up to the extension developer to decide whether theres is a link or component, but by extending the same extension point they may get different context available, feels confusing.

Hmm, I don't think so, the two context objects should be the same, the difference is in how the extensions receive it. Probably the communication of this is not clear though, I think we should have a chat about it and try to make sure it's easier to understand.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Yeah, good point, and some more explanation would be good indeed. I will leave it for a separate refinement PR if you don't mind, let me know if you have objections.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: We did it this way because we thought that it's more natural for React components to receive input through component props, and this pattern can save an extra step for the plugin dev, they don't need to proxy the context object that they would receive in the extension registration callback to the  extension component, they can just rely on what the extension point passes down to them. The extension point can even pass down multiple props, if they thing it's more meaningful / semantical.

> what we want to ensure the reader takes away having read this comment?

I basically wanted to raise their attention that passing down contextual information to components is different than for links. 

Probably there is some discrepancy here that we should chat through though, we should try making this as easy to understand as possible.

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: > The migration process is just to swap the hook right? (Which we could mention).

Basically yes - however if they have tests then updating those is bit more work (due to mocking).


> I worry it's just noise that makes the doc look longer and more complicated.

I agree.


Yeah, let's keep them for now (in this PR) and let's chat about when we should remove them. My other concern is that we might don't even want to remove them / deprecate them in the end, but clearly communicate what they can be used for. (We need to think about the use cases, but a function like this could be use in places outside of a React component, which can come handy. It needs some more thought.)

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: This would happen when the `context {}` object is constructed dynamically from other variables, which can change between renders. Strange that Github didn't update the markdown snippet above, as the I have linked it to an example 🤔 (Probably I did that after your comment?)

```
[More info](#example---rendering-link-extensions-dynamic-context)
```

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Yes, I think we do, it can be important help for plugin devs. (On the other this could be another candidate for having a "Considerations (...)" / "Best practices" section.)

docusaurus/docs/ui-extensions/create-an-extension-point.md
leventebalogh: Good feedback, let's discuss how the best we could explain this difference in the docs.

PR #890, opened 2024-04-24
Create plugin: CI workflow improvements

sunker:
This PR is making a few changes to the e2e related workflow bits:
* Use the right snippet in the article that walks through how to run e2e tests in ci
* Not doing a sparse checkout of the repo anymore in the e2e test job. Doing a sparse checkout have little benefits, and may cause confusion when contributors wants to go beyond the basic steps in the workflow. See [this](https://raintank-corp.slack.com/archives/C01C4K8DETW/p1713902551454569) thread for example. 
* Adds step that uploads Playwright report to GCS in case org is Grafana. Please note that even though someone were to falsely specify grafana as the org, this step would still not work for them as it requires access to id-tokens that are only available for repos within the grafana org. 

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@4.7.1-canary.890.babb00f.0
  npm install @grafana/plugin-e2e@1.1.2-canary.890.babb00f.0
  # or 
  yarn add @grafana/create-plugin@4.7.1-canary.890.babb00f.0
  yarn add @grafana/plugin-e2e@1.1.2-canary.890.babb00f.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -186,7 +186,7 @@ jobs:
           interval: 500
 
       - name: Install Playwright Browsers
-        run: {{ packageManagerName }} playwright install chromium --with-deps
+        run: {{ packageManagerName }} exec playwright install chromium --with-deps
 
       - name: Run Playwright tests
         id: run-tests
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -203,7 +203,7 @@ jobs:
           name: $\{{ matrix.GRAFANA_IMAGE.NAME }}-v$\{{ matrix.GRAFANA_IMAGE.VERSION }}-$\{{github.run_id}}-server-log
           path: grafana-server.log
           retention-days: 5
-
+  
       # If your repository is public, uploading the Playwright report will make it public on the Internet.
       # Beware not to expose sensitive information.
       - name: Upload artifacts
@@ -213,3 +213,14 @@ jobs:
           name: playwright-report-$\{{ matrix.GRAFANA_IMAGE.NAME }}-v$\{{ matrix.GRAFANA_IMAGE.VERSION }}-$\{{github.run_id}}
           path: playwright-report/
           retention-days: 5
+
+{{#if_eq orgName "grafana"}}
+      # Repos in the grafana Github org can publish the report to GCS so that they can be browsed without having to download them.
+      # Beware this will make the report public on the Internet, so do not uncomment this step if the report contains sensitive information.
+      # - name: Publish report to GCS
+      #   if: $\{{ always() && steps.run-tests.outcome == 'failure' }}
+      #   uses: grafana/plugin-actions/publish-report@main
+      #   with:
+      #     grafana-version: ${{ matrix.GRAFANA_IMAGE.VERSION }}
+      #     path: playwright-report
+{{/if_eq}}
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -182,7 +182,7 @@ jobs:
           interval: 500
 
       - name: Install Playwright Browsers
-        run: npx playwright install chromium --with-deps
+        run: {{ packageManagerName }} playwright install chromium --with-deps
 
       - name: Run Playwright tests
         id: run-tests
--- a/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
@@ -11,11 +11,7 @@ on:
       - master
       - main
 
-permissions:
-  contents: read
-{{#if_eq orgName "grafana"}}
-  id-token: write
-{{/if_eq}}
+permissions: read-all
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


mckn (approved):
LGTM!

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
```diff
@@ -11,7 +11,11 @@ on:
       - master
       - main
 
-permissions: read-all
+permissions:
+  contents: read
+{{#if_eq orgName "grafana"}}
+  id-token: write
+{{/if_eq}}
```
jackw, line 18: Are we sure we want to introduce Grafana specific code to create-plugin?

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
```diff
@@ -218,3 +217,14 @@ jobs:
           name: playwright-report-$\{{ matrix.GRAFANA_IMAGE.NAME }}-v$\{{ matrix.GRAFANA_IMAGE.VERSION }}-$\{{github.run_id}}
           path: playwright-report/
           retention-days: 5
+
+{{#if_eq orgName "grafana"}}
+      # Repos in the grafana Github org can publish the report to GCS so that they can be browsed without having to download them.
+      # This will make the report public on the Internet, so beware to skip this step if the report contains sensitive information.
```
jackw, line 223: What sort of sensitive information could this expose? Could having it as a default be rather risky?

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
leventebalogh: I am kind of leaning towards what Jack says, and try to keep this more generic (not introducing internal plugin development related knowledge in the templates).

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
```diff
@@ -141,11 +145,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          sparse-checkout: |
-            provisioning
-            tests
-            .config
```
leventebalogh, line 148: Did we remove this because it didn't give us much in a performance perspective, or because we were missing something?

mckn: I think you can find the reason for removing this in the PR description.

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
mckn: I guess I can see your worries. It is easy it becomes a slippery slope with lots of Grafana internal specific logic added to the template. On the other hand, we are creating lots of pluggins internally so it might be worth having some of this stuff added via the template. I'm a bit torn 🙈

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
mckn: hmm, I'm starting to think that a better approach would be to upload artifacts to github instead (if we want to enable this functionality). Then it would work for all users regarding if it is within the grafana org or not.

tolzhabayev: uploading to github would not give you a possibility to link to it - you would have to download the report and run `npx playwright show-report`

uploading to GCS gives you a way to share a link with someone who doesnt want to download unzip and run report locally - so kind of a convenience feature.

tolzhabayev: > What sort of sensitive information could this expose? Could having it as a default be rather risky?

Same as any e2e test report more or less - it contains your browser state, screenshots of the execution, etc. and if you have something sensitive in there it will be in the report. Since it is quite impossible to automatically know about what could theoretically be sensitive on the screen there is no automated solution to this - https://github.com/microsoft/playwright/issues/19992 

Hence the guidance in the comments above the action - if you are for some reason using real credentials or have sensitive information in the flow - you have to handle it yourself or not publish it to a public bucket.

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
tolzhabayev: so what is the alternative? Another separate workflow for internal devs?

jackw: Yep my biggest concern here isn't so much this single addition but that we open the door for a community tool to become peppered with our own internal teams "needs". I would vote that we create separate workflows for our internal teams. From what I've gathered they each have their own ideas and would be good to pool those folk together to create something that they can share internally. I'm pretty sure most of them edit/delete this workflow anyhow.

Additionally this template variable is based on user input. If I want my plugin to be called `grafana-awesome-panel` I might (through trial and error) come to the conclusion that to do that I need to use `grafana` as the org name when I scaffoled. Eventually I submit my plugin for review and I'm asked to change the org name unaware that my plugin was scaffolded with specifics used by Grafanas own plugin teams.

packages/create-plugin/templates/github/ci/.github/workflows/ci.yml
sunker: Yep, and the same security concern applies when uploading artifacts to GH as the report becomes public on the Internet. Playwright docs have [example workflows](https://playwright.dev/docs/ci-intro#setting-up-github-actions) where reports are being uploaded to GH artifacts, bit they don't mention security concerns.

sunker: I'm rather safe than sorry - let's comment out the upload artifacts step.

sunker:
Thanks for the review folks. Have removed the grafana specific bits and commented out the report uploading step (see [this](https://github.com/grafana/plugin-tools/pull/890/commits/babb00f1d74cdf1ceefc134e7cf73213392d61ab) commit). Let me know what you think.

PR #901, opened 2024-05-06
Chore: removing version from docker compose file as it is deprecated

tolzhabayev:
Specifying version in docker-compose file has been deprecated for a while and now it shows a warning. It still works properly but we can start removing it for newly scaffolded plugins as it has no value.

It removes warnings like this:

```
docker-compose.yaml: `version` is obsolete
```
More on why https://docs.docker.com/compose/compose-file/04-version-and-name/
<!--

- Automatically closes linked issue when the Pull Request is merged.

Usage: "Fixes #<issue number>", or "Fixes (paste link of issue)"

### DIFF
--- a/packages/create-plugin/templates/common/docker-compose.yaml
+++ b/packages/create-plugin/templates/common/docker-compose.yaml
@@ -1,5 +1,3 @@
-version: '3.0'
-
 services:
   grafana:
     user: root


academo (approved):

leventebalogh (approved):
🚀

PR #902, opened 2024-05-07
PME: Removed dependency on tsquery

mckn:
This PR will remove the dependency on `tsquery` package since it doesn't seem to work properly for both TS4 and TS5. I replaced it with our own implementation to find the correct identifiers that seems to work for both TS4 and TS5.

It would be nice to be able to write some integration tests that check this for different versions of typescript. Opened an [issue](https://github.com/grafana/plugin-tools/issues/904) for doing this.

I have tested the canary version with a plugin using TS4 and it does work 🎉 

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/plugin-meta-extractor@0.0.2-canary.902.bd718a3.0
  # or 
  yarn add @grafana/plugin-meta-extractor@0.0.2-canary.902.bd718a3.0
  ```
</details>

### DIFF
--- a/packages/plugin-meta-extractor/src/utils.ts
+++ b/packages/plugin-meta-extractor/src/utils.ts
@@ -1,6 +1,5 @@
 import * as ts from 'typescript';
 import createDebug from 'debug';
-import * as tsquery from '@phenomnomnominal/tsquery';
 import { PluginExtensionTypes } from './types';
 
 const _debug = createDebug('plugin-meta-extractor');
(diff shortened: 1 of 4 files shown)


packages/plugin-meta-extractor/src/utils.ts
```diff
@@ -39,6 +39,48 @@ export function getExtensionConfigs(functionName: string, ast: ts.Node, checker:
   return extensionConfigs;
 }
 
+function createExtensionMatcher(functionName: string): (node: ts.Node) => boolean {
```
leventebalogh, line 42: Nit: as this seems more like a general function, maybe we could call it something like `createNodeMatcher` or `createFuncitonMatcher`?

leventebalogh (approved):
Nice 👏

PR #903, opened 2024-05-08
Chore: Adding plugin meta extractor to GH templates

mckn:
Adding plugin meta extractor to the GH template so you can select that as an option when opening issues or feature requests.

Fixes #


leventebalogh (approved):
🚢🇮🇹

PR #922, opened 2024-05-24
Docs: Remove prompts content from update command page

jackw:
This PR removes the **Prompts** content from the create plugin update command docs page as they don't exist when the command if run these days...

image

Fixes #


leventebalogh (approved):
🚀

sunker (approved):
🚀

PR #930, opened 2024-05-28
Plugin E2E: Allow overriding grafanaAPICredentials

### LINKED ISSUE #934: Feat: Make it possible to switch admin client user
### Which areas does this feature request relate to

- [ ] Create Plugin
- [ ] Sign Plugin
- [X] Plugin E2E
- [ ] Plugin Meta Extractor
- [ ] Documentation

### Problem

In a few places in the code, plugin-e2e uses the Grafana API to administrate users, data sources etc. Before this PR, the Grafana API was always called on behalf of the admin:admin user (the default server admin account) as it has all the necessary permissions. This is working fine in most cases, but in the test environment for the SLO app, this account doesn't exist so they're getting 401 errors when setting up users (more details on how to create users [here](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/use-authentication#plugins-that-dont-use-rbac)).

### Solution

Should be possible to override the default admin user

### Alternatives

_No response_

### Additional context

_No response_

### Are you interested in contributing the solution?

- [X] Yes
- [ ] No

sunker:

In a few places in the code, plugin-e2e uses the Grafana API to administrate users, data sources etc. Before this PR, the Grafana API was always called on behalf of the admin:admin user (the default server admin account) as it has all the necessary permissions. This is working fine in most cases, but in the test environment for the SLO app, this account doesn't exist so they're getting 401 errors when setting up users (more details on how to create users [here](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/use-authentication#plugins-that-dont-use-rbac)).  

With this PR, the Grafana API is still called on behalf of the admin:admin user by default. However, consumers now have the option to override this user by providing a `grafanaAPIUser` in the playwright config. `grafanaAPIUser` is a Playwright [option](https://playwright.dev/docs/test-configuration#basic-configuration), meaning it can be defined on a global level and on a project level. Also, when specifying the user to use when running the e2e test (not the same thing as the `grafanaAPIUser`), it's now possible to set `skipCreateUser` prop. This gives the consumer the option to **not** create the user in the database. 

This is an example where one user is being used when calling the Grafana API and another one is being used when running the e2e tests.

```ts
export default defineConfig({
    ...
 testDir: './tests',
   use: {
     baseURL: 'http://localhost:3000',
     grafanaAPIUser: {
       user: 'server-admin',
       password: 'server-admin',
     },
    },
    projects: [
    {
      name: 'auth',
      testDir: pluginE2eAuth,
      testMatch: [/.*\.js/],
      use: {
        user: {
          user: 'slo',
          password: 'slo',
          skipCreateUser: true, // this user already exist in the db, so no need to create it during test setup
        },
      },
    },
    {
      name: 'run-tests',
      use: {
        ...devices['Desktop Chrome'],
        storageState: 'playwright/.auth/slo.json',
      },
      dependencies: ['auth'],
    }
  ],
});
```

Fixes #934

<details>
  <br />

  ```bash
  npm install @grafana/plugin-e2e@1.3.0-canary.930.a85b4bb.0
  # or 
  yarn add @grafana/plugin-e2e@1.3.0-canary.930.a85b4bb.0
  ```
</details>

### DIFF
--- a/packages/plugin-e2e/playwright.config.ts
+++ b/packages/plugin-e2e/playwright.config.ts
@@ -49,7 +49,6 @@ export default defineConfig<PluginOptions>({
         user: {
           user: 'admin',
           password: 'admin',
-          skipCreateUser: true,
         },
       },
     },
--- a/packages/plugin-e2e/playwright.config.ts
+++ b/packages/plugin-e2e/playwright.config.ts
@@ -22,15 +22,10 @@ export default defineConfig<PluginOptions>({
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: 'http://localhost:3000',
-
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
     grafanaAPIUser: {
       user: 'admin',
-      password: 'admin',
-    },
-    httpCredentials: {
-      username: 'admin',
       password: 'admin',
     },
     featureToggles: {
--- a/packages/plugin-e2e/playwright.config.ts
+++ b/packages/plugin-e2e/playwright.config.ts
@@ -25,6 +25,10 @@ export default defineConfig<PluginOptions>({
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
+    grafanaAPIUser: {
+      user: 'admin',
+      password: 'admin',
+    },
     httpCredentials: {
       username: 'admin',
       password: 'admin',
@@ -41,6 +45,13 @@ export default defineConfig<PluginOptions>({
       name: 'authenticate',
       testDir: './src/auth',
       testMatch: [/.*auth\.setup\.ts/],
+      use: {
+        user: {
+          user: 'admin',
+          password: 'admin',
+          userExists: true,
+        },
+      },
     },
     // Login to Grafana with new user with viewer role and store the cookie on disk for use in other tests
     {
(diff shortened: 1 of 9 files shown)


packages/plugin-e2e/playwright.config.ts
```diff
@@ -25,6 +25,10 @@ export default defineConfig<PluginOptions>({
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
+    grafanaAPIUser: {
```
jackw, line 28: nit: Besides a user account are there other ways to auth with the API? Maybe `grafanaAPIAuth` or `grafanaAPICredentials` would work rather than mentioning user?

packages/plugin-e2e/src/auth/auth.setup.ts
```diff
@@ -1,7 +1,7 @@
 import { test as setup } from '../';
 
 setup('authenticate', async ({ login, createUser, user }) => {
-  if (user) {
+  if (user && !user.skipCreateUser) {
```
jackw, line 4: Not sure I'm understanding the skipCreateUser. Doesn't createUser [already handle cases](https://github.com/grafana/plugin-tools/pull/930/files#diff-ac3488fe952ba2e1dc2e3f2af971744712ff325fb4fff817d3b55f4b21117834R43) where the user already exists?

jackw (approved):
LGTM! 🚀 

Left a couple of nits. Could we add something to the docs about this?

packages/plugin-e2e/playwright.config.ts
sunker: yep, 100% agree

packages/plugin-e2e/src/auth/auth.setup.ts
sunker: No that's right, user is only created if it doesn't exist, so this is not entirely necessary. I guess I thought this would make this more clear from an end-user standpoint, but that's not the case so I'll remote this prop. Thanks for feedback.

PR #935, opened 2024-05-30
Docs: Plugin E2E - Add section about user management

sunker:

This PR adds a section that describes how user are being managed in plugin-e2e. This feature was introduced in [this](https://github.com/grafana/plugin-tools/pull/930) PR. 

Fixes #

### DIFF
--- a/docusaurus/docs/e2e-test-a-plugin/authentication.md
+++ b/docusaurus/docs/e2e-test-a-plugin/authentication.md
@@ -21,7 +21,7 @@ To be able to interact with the Grafana UI, you need to be logged in to Grafana.
 
 If your plugin doesn't use RBAC, you can use the default server administrator credentials to login.
 
-In the following example, there's a [setup project](https://playwright.dev/docs/test-global-setup-teardown#setup-example) called `auth`. This project invokes a function in the `@grafana/plugin-e2e` package that logins to Grafana using `admin:admin`. The authenticated state is stored on disk with this file name pattern: `<plugin-root>/playwright/.auth/<username>.json`.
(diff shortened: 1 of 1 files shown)


docusaurus/docs/e2e-test-a-plugin/authentication.md
```diff
@@ -94,3 +94,28 @@ export default defineConfig<PluginOptions>({
   ]
 })
 ```
+
+## Managing users
+
+When a `user` is defined in a setup project like in the RBAC example above, `plugin-e2e` attempts to create the user using the Grafana HTTP API. This action requires elevevated permissions, so by default the server administrator credentials `admin:admin` will be used. If the end-to-end tests are targeting the [development environment](../get-started/set-up-development-environment.mdx) scaffolded with the `create-plugin`, this works fine but for other test environments the server adminstator password has likely been changed. In that case, you can provide the correct credentials by setting `grafanaAPICredentials` in the global options.
```
jackw, line 100: nit: minor changes to try to make the sentences flow more easily due to a lot of use of use, using and user. 

```suggestion
When a `user` is defined in a setup project (like in the RBAC example above) `plugin-e2e` will use the Grafana HTTP API to create the user account. This action requires elevated permissions, so by default the server administrator credentials `admin:admin` will be used. If the end-to-end tests are targeting the [development environment](../get-started/set-up-development-environment.mdx) scaffolded with `create-plugin`, this will work fine. However for other test environments the server administrator password may be different. In that case, you can provide the correct credentials by setting `grafanaAPICredentials` in the global options.
```

jackw (approved):
Awesome! Left a nit in an attempt to make the paragraph flow more easily. Feel free to ignore it if you disagree. 🚀

docusaurus/docs/e2e-test-a-plugin/authentication.md
sunker: Much better, thanks!

PR #944, opened 2024-06-05
Plugin E2E: Fix menuitemclick in Grafana <=9.1.0

sunker:

This PR changes CI so that e2e tests are targeting a larger range of Grafana versions. Since plugin-e2e APIs are supposed to be compatible with Grafana =>8.5.5, it's important that we cover these older versions in tests. 

While doing this, I discovered that clicking menu items doesn't work as expected in versions older than 9.2.0, so adding a fix for that. 

Fixes #

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -23,7 +23,8 @@ jobs:
         id: resolve-versions
         uses: grafana/plugin-actions/e2e-version@main
         with:
-          version-resolver-type: version-support-policy
+          version-resolver-type: plugin-grafana-dependency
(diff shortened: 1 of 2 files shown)


jackw (approved):
LGTM 🚀 

I'm surprised that we need to support all the way to 8.5.x in plugin-e2e. 🤔

sunker:
> I'm surprised that we need to support all the way to 8.5.x in plugin-e2e. 🤔

Yeah it's so we can support enterprise plugins. But should probably drop support for 8.x now that G11 is out the door. I'll discuss with David.

PR #961, opened 2024-06-14
Docs: reorganization of panel plugin topics

josmperez:
Complete 4th stage of the plugin-tools site reorganization project. Move/rename site, fix broken links.

Note that there is only 1 redirect created in this PR despite moves of 3 files. The other 2 files will be redirected to the corresponding topic in "how-to-guides/data-source-plugins".

Fixes: https://github.com/grafana/plugin-tools/issues/925

Instructions for testing redirects: docusaurus/website/README.md#Redirects

Testing: Refer to [design doc](https://docs.google.com/document/d/1T3v41Ik5hOYNL0xMtPLOiK38IPGUlvsEgzFkDKvsROM/edit)

PR #985, opened 2024-07-01
Move panel topics to how to guides

sympatheticmoose:

#961 was getting a bit messy so started afresh with the proposed changes

Fixes #

This is identical in _intent_ to #961 except for the following changes:
- doesn't rename custom panel options file (simpler)
- uses "Read data from a data source" rather than "Read data frames" - this is up for debate, but wanted to show an alternative which led with the desired outcome rather than the mechanism
- the error handling for panels document is identical to the one in extend a plugin as the content is all about panel plugins. We _could_ remove the original or keep until we have ones per plugin type which are actually relevant. 

When I say identical in intent that means I have changed other things as well:
- fixed issues with dataframes and other files being deleted
- fixed index files for how-to and panel-plugins to be relevant 
- fixed build issue from redirects


josmperez (approved):
LGTM.

josmperez:
Thanks for giving your attention to these issues. I think this looks good.

academo (approved):

PR #1021, opened 2024-07-22
Docs: Add notes on 'build targets' on Get Started

### LINKED ISSUE #1012: Fix mismatch in 'Build the backend' instructions
[Comment](https://github.com/grafana/plugin-tools/pull/993/files#r1677442802) from JackW:

There is a mismatch between amd and arm64 and the docker container which we should likely put a note about here. Previously we forced amd in the docker platform but that causes the backend delve dev env to fail to run on apple silicon.

Also not really sure if this should be added here but we have more than just build:[arch] as available targets. e.g.

| `build:backend` | build a production build for the current platform | `mage -v build:backend` |

Maybe we circle back to the whole backend build commands and expand them later on.

josmperez:
(Draft) Add a note about a problem running builds on Apple silicon to Get Started topic; add a second example to the Mage description.

Fixes https://github.com/grafana/plugin-tools/issues/1012

### DIFF
--- a/docusaurus/docs/get-started/get-started.mdx
+++ b/docusaurus/docs/get-started/get-started.mdx
@@ -232,7 +232,7 @@ To build for production, run:
 
 ### Build the backend
 
-If your plugin includes a [backend](../key-concepts/backend-plugins/index.md) component, you can build using mage:
+If your plugin includes a [backend](../key-concepts/backend-plugins/index.md) component, you can build using Mage:
 
 ```shell
 mage -v build:linux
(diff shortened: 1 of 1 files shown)


docusaurus/docs/get-started/get-started.mdx
```diff
@@ -243,13 +243,20 @@ mage -v build:linux
...
 
 List all available Mage targets for additional commands:
 
 ```bash
 mage -l
 ```
 
+:::note
+
+There is a mismatch between AMD, ARM64, and the Docker container which may cause the backend Delve debugger dev env to fail to run on Apple silicon.
```
academo, line 256: @oshirohugo can you confirm this is still the case?

oshirohugo: This was improved, but it seems it still not perfect and it will only work in Apple silicon after two runs.
@jackw you worked on that, can you confirm?

jackw: @oshirohugo I'll do some testing here and get back to you. It was a while since I tried this debugger.

josmperez: @jackw Do you still feel this needs further testing, or are we ok to merge?

Issue #995, opened 2024-07-09
Simplify prompts for create-plugin


sympatheticmoose:
Create plugin is an opinionated tool for scaffolding plugins. We aim to provide developers with everything they need to develop a plugin which follows our best practices, and strive towards a simple, frictionless experience to getting started.

Following the removal of a generator framework in `create-plugin `we have more flexibility in our [prompts](https://grafana.com/developers/plugin-tools/reference/prompts) and the output from `create-plugin`. This gives us an opportunity to revisit the interactive questions we ask to try and improve the developer experience. 

We would welcome any comments, suggestions or concerns from plugin developers to help improve this experience.

### Summary

- Moves plugin type question to start as more logical and important starting point
- Adds basic descriptors for plugin types to make clearer on purpose/use case for each
- Gives indication of why a backend component might be needed
- Removes (3) prompts to simplify and shorten interaction: description, CI/Release actions and levitate action. We assume users can delete actions if required and that their inclusion will have no significant negative impact regardless. As in #771 we will need to ensure that our documentation supports discovery and removal of non-essential elements. An Empty description can be flagged as part of validator checks as this is more important when the plugin is to be shared.
- Shortens basic prompts language
- Removes output of each generated file as this offers no value and clutters the terminal
- Adds new success messages to indicate presence of e2e tests and GH actions, slight rewording for others <verb> <action>

### Prompts

```
- Select plugin type
  - App (Custom pages, UI Extensions and bundling other plugins)
  - Data source (Query data from a custom source)
  - Panel (New visualization for data or a widget)
  - App with Scenes (Create dynamic dashboards in app pages)
- [If data source] Does your plugin require a backend to support server-side functionality (alerting, advanced authentication, public dashboards, etc)? [y/N]
- [If app] Does your plugin require a backend to support server-side functionality (calling external APIs, custom backend logic, advanced authentication, etc)? [y/N]
- Enter a name for your plugin
- Enter your organization name (usually your Grafana Cloud org)
```

### Success output

```
✔︎ Scaffolded <plugin-ID> <plugin-type> plugin (with Go backend)
✔︎ Added basic e2e test with Playwright
✔︎ Provisioning provided for [ app / data source instance / basic dashboard and TestData data source instance / app and TestData data source instance]
✔︎ Configured Development environment (Docker)
✔︎ Added default GitHub actions for CI, releases and Grafana compatibility 
✔︎ Updated Grafana go sdk to <version> (latest)

## What's next?

[...]

PR #1024, opened 2024-07-31
update prompts docs

sympatheticmoose:

Updates the references to prompts from the create plugin tool to go alongside the simplify prompts PR

This appears to be the only place we directly reference in the docs site and the generated output has not changed _significantly_ (only thing I noticed was more specifics if running ARM macs)

Part of #995 


leventebalogh (approved):
🚀

PR #1033, opened 2024-08-02
Docs: Improve documentation for plugins ReqAction

gamab:

This PR clarifies the access control for plugin includes and routes (specified by the `action` and `reqAction` fields, respectively). While Grafana can restrict access based on the specified actions, it does not currently support fine-grained authorization for individual resources, except for the plugin itself or its associated data sources. 

Fixes https://github.com/grafana/identity-access-team/issues/769


docusaurus/docs/reference/metadata.md
```diff
@@ -240,18 +240,18 @@ For data source plugins. Proxy routes used for plugin authentication and adding
...
+| `jwtTokenAuth` | [object](#jwttokenauth) | No       | For data source plugins. Token authentication section used with an JWT OAuth API.                                                                                                     |
+| `method`       | string                  | No       | For data source plugins. Route method matches the HTTP verb like GET or POST. Multiple methods can be provided as a comma-separated list.                                             |
+| `path`         | string                  | No       | For data source plugins. The route path that is replaced by the route URL field when proxying the call.                                                                               |
+| `reqAction`    | string                  | No       | The RBAC action a user must have to use this route. **Warning**: unless the action targets the plugin or one of its datasources, only the action is verified, not what it applies to. |
```
leventebalogh, line 250: **Nit** 
```suggestion
| `reqAction`    | string                  | No       | The RBAC action a user must have to use this route. **Warning**: unless the action targets the plugin (or a nested datasource plugin), only the action is verified, not what it applies to. |
```

docusaurus/docs/reference/metadata.md
```diff
@@ -113,7 +113,7 @@ To try this feature out, follow this [example](https://github.com/grafana/grafan
 
 | Property     | Type    | Required | Description                                                                                                                                                                                     |
 | ------------ | ------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `action`     | string  | No       | The RBAC action a user must have to see this page in the navigation menu.                                                                                                                       |
+| `action`     | string  | No       | The RBAC action a user must have to see this page in the navigation menu. **Warning**: unless the action targets the plugin, only the action is verified, not what it applies to.               |
```
leventebalogh, line 116: Hmm, maybe it's just me, and in that case sorry, but what is exactly the difference between verifying the action itself versus what it is applied to?

gamab: Thanks for the comment. It shows how confusing that is, which isn't good security wise :disappointed:

Let's take an example, let's say the prometheus plugin defines a route:
```json
    {
      "method": "POST",
      "path": "/config/v1/rules",
      "reqRole": "Editor",
      "reqAction": "folders:write"
    }
 ```
 Then anyone that can edit **any** folder will have access to the data source route.
 
 On the other hand, with:
```json
    {
      "method": "POST",
      "path": "/config/v1/rules",
      "reqRole": "Editor",
      "reqAction": "datasources:write"
    }
 ```
We make sure that you can access this data source route only if you can edit this specific data source.

leventebalogh: I see, thanks for the clarification mate! 🙏 

Hmm, I have one more thing that is not totally clear (and probably comes from my lack of knowledge): by just looking at `datasources:write` I wouldn't immediately think that it's meant for this particular datasource, but to any datasource (=if the user can edit any datasource).

gamab: I understand your point. I wrote it this way in a first place. But, some developers interpreted it as applying specifically to the datasource involved in the current request being proxied, which also makes sense. However, this interpretation led to a security vulnerability (CVE-2024-6322), which we now have fixed and these docs aim to clarify this new behavior.

josmperez (approved):
LGTM.

josmperez (approved):
From a doc clarity and style perspective, LGTM.

PR #1038, opened 2024-08-07
Docs: Add Ratings functionality

### LINKED ISSUE #1006: docs: provide rating and feedback mechanism
Similar to what we have on the main [grafana docs](https://github.com/grafana/website/blob/9a0f206c9ea84d7301e410291e78beaa1001129b/layouts/partials/docs/rating.html)

Screenshot 2024-07-12 at 10 51 46

It would be useful to be able to gather insights from users on which pages are more/less helpful and directly encourage feedback.

A quick search suggests no docusaurus plugin exists for this and instead the recommendation seems to be to [swizzle](https://docusaurus.io/docs/swizzling)

jackw:
This PR introduces a Ratings component to allow users to give feedback on how useful they find a page on our docs site.

The main docs website uses faro which I've brought across for the ride. I'm not sure if the behaviour is correct that faro is only setup and sends measurements if the cookies are agreed to (same as rudderstack).

**What does it look like**:
Original buttons state:
Screenshot 2024-08-07 at 10 18 43
Once rated buttons state:
Screenshot 2024-08-07 at 10 18 49

Fixes #1006


docusaurus/website/docusaurus.config.dev.js
```diff
@@ -74,6 +74,12 @@ const config = {
       sdkUrl: 'https://rsdk.grafana.com',
     },
     canSpamUrl: 'https://grafana-dev.com/canspam',
+    faroConfig: {
+      url: '/connect/af1fca71911a9641ebdffddb56889e97',
```
tolzhabayev, line 78: I think this is a production key

tolzhabayev (approved):
Generally great addition, I am not 100% sure we will keep using the same way as website does but for now it's fine to attach ourselves to them. I am mostly thinking about data retention which is way higher if we just use rudderstack.

PR #1069, opened 2024-08-21
Create Plugin: Fix support yarn pnp with virtual modules

jackw:
It looks like yarn pnp's resolution hijacking causes the current virtual public path setup to fail as it redirects everything to it's zipped dependencies which causes the virtual module to fail to be found. This PR addresses it by moving the file to a virtual_modules directory and including this directory in module resolution webpack settings.

Supporting yarn pnp is a bit of a thorny issue, won't merge this until we reach some consensus.

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.3.3-canary.1069.afa1a66.0
  # or 
  yarn add @grafana/create-plugin@5.3.3-canary.1069.afa1a66.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/_package.json
+++ b/packages/create-plugin/templates/common/_package.json
@@ -51,6 +51,7 @@
     "sass-loader": "13.3.1",
     "style-loader": "3.3.3",
     "swc-loader": "^0.2.3",
+    "terser-webpack-plugin": "^5.3.10",
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.2.0",
     "typescript": "4.8.4",
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -22,7 +22,7 @@ const pluginJson = getPluginJson();
 const cpVersion = getCPConfigVersion();
 
 const virtualPublicPath = new VirtualModulesPlugin({
-  'node_modules/grafana-public-path.js': `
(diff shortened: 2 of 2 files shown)


domasx2 (approved):

packages/create-plugin/templates/common/_package.json
```diff
@@ -51,6 +51,7 @@
     "sass-loader": "13.3.1",
     "style-loader": "3.3.3",
     "swc-loader": "^0.2.3",
+    "terser-webpack-plugin": "^5.3.10",
```
academo, line 54: is this necessary?

jackw: For yarn pnp it is required for builds to run as without it we access a transitive dependency and yarn errors out with something like:

```
Error: Your application tried to access terser-webpack-plugin, but it isn't declared in your dependencies; this makes the require call ambiguous and unsound.
```

academo (approved):

leventebalogh (approved):
Generally looks good to me - although are we sure we would like to support Yarn PNP? If yes, do we know which plugins are currently having issues with this?

jackw:
After further discussions within the plugins platform team we've reached the consensus that we don't want to shoulder the burden of supporting Yarn berrys PnP functionality for plugin tooling.

If you are a plugin developer using yarn berry and it's PnP functionality we should strongly advise you switch to using [the tried and true way](https://yarnpkg.com/features/linkers#nodelinker-node-modules) to install node.js projects by setting `nodeLinker: node-modules` in the .yarnrc.yml file.

PR #1083, opened 2024-08-30
Create Plugin): Bump Webpack to >5.94.0

jackw:
Given [this advisory](https://github.com/advisories/GHSA-4vvj-4cpr-p986) we should bump the version so when `upgrade` is run the plugin is no longer vulnerable. (I don't believe any plugins actually have the auto public path code in them but better to stay up to date.)

Fixes #


sunker (approved):
LGTM

PR #1089, opened 2024-09-03
Renovate: Add gomod manger and package rule for plugin sdk

sunker:

This PR enables Renovate to manage updates to the grafana-plugin-sdk-go. As soon as there's a new minor or patch release available for the sdk, Renovate will raise a PR that bumps its version in the gomod file for create-plugin backend templates.

Fixes #

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,7 +5,7 @@
   "separateMajorMinor": false,
   "labels": ["dependencies", "release", "patch"],
   "reviewers": ["team:grafana/plugins-platform-frontend"],
-  "enabledManagers": ["regex"],
+  "enabledManagers": ["regex", "gomod"],
   "customManagers": [
     {
       "customType": "regex",
@@ -23,10 +23,17 @@
   ],
   "packageRules": [
     {
+      "matchManagers": ["regex"],
       "matchPackagePatterns": ["@grafana/*", "grafana/grafana-enterprise"],
       "matchUpdateTypes": ["minor", "patch", "major"],
       "groupName": "grafana dependencies",
       "groupSlug": "all-grafana"
+    },
+    {
+      "reviewers": ["team:grafana/plugins-platform-backend"],
+      "matchManagers": ["gomod"],
+      "matchDepNames": ["grafana-plugin-sdk-go"],
+      "matchUpdateTypes": ["minor", "patch"]
     }
   ]
 }


.github/renovate.json
```diff
@@ -23,10 +23,17 @@
...
   "packageRules": [
     {
+      "matchManagers": ["regex"],
       "matchPackagePatterns": ["@grafana/*", "grafana/grafana-enterprise"],
       "matchUpdateTypes": ["minor", "patch", "major"],
       "groupName": "grafana dependencies",
       "groupSlug": "all-grafana"
+    },
+    {
+      "reviewers": ["team:grafana/plugins-platform-backend"],
```
sunker, line 33: I _think_ reviewers can be overridden on package rule level.

.github/renovate.json
```diff
@@ -23,10 +23,17 @@
...
+      "matchManagers": ["gomod"],
```
xnyo, line 34: Is there a way to test this? I'm not sure how renovate works, but I suppose it somehow runs `go get -u github.com/grafana/grafana-plugin-sdk-go` in order to update the dependency.

If this is the approach it uses, then this won't work properly. This is because the `go.mod` and `go.sum` files we have in `packages/create-plugin/templates/backend*` aren't valid, as they contain templating snippets:

https://github.com/grafana/plugin-tools/blob/40a745edefbb4580ffb93ee92ca04017f8b1dae4/packages/create-plugin/templates/backend/go.mod#L1

Same for the `.go` files:

https://github.com/grafana/plugin-tools/blob/40a745edefbb4580ffb93ee92ca04017f8b1dae4/packages/create-plugin/templates/backend/pkg/main.go#L8

Attempting to run `go get -u ...` or `go mod tidy` in one of the tempalte's folders results in the following:

```bash
╭─ ~/grafana/plugin-tools/packages/create-plugin/templates/backend on main !3                                                  
╰─❯ go mod tidy
go: errors parsing go.mod:
go.mod:1: usage: module module/path

╭─ ~/grafana/plugin-tools/packages/create-plugin/templates/backend on main !3                                                  
╰─❯ go get -u github.com/grafana/grafana-plugin-sdk-go
go: errors parsing go.mod:
go.mod:1: usage: module module/path
```

sunker: I think Renovate uses regex to update the go.mod file, so this bit is working fine even though the file contains templates. After that it runs `go get`, and as you say this won't work given the module path is invalid (tested and confirmed in [this](https://github.com/sunker/plugin-tools/pull/40/files) PR in a forked repo). So thanks for catching this @xnyo! 

I assume there's no way to force `go get` and `go mod tidy` to complete even with an invalid module path? If not, I'll close this PR and think of other ways of solving this. Could possibly be solved using a regex manager (would be messy) or by finding other ways to template orgname and pluginname.

xnyo: > I assume there's no way to force go get and go mod tidy to complete even with an invalid module path?

@sunker I'm not aware of any way to do it, no

---

However, when a new plugin is scaffolded with `create-plugin`, it will automatically run `go get -u ...` and `go mod tidy`, as long as `go` is installed on the developer's machine. See here:

https://github.com/grafana/plugin-tools/blob/40a745edefbb4580ffb93ee92ca04017f8b1dae4/packages/create-plugin/src/commands/generate.command.ts#L54-L56

https://github.com/grafana/plugin-tools/blob/40a745edefbb4580ffb93ee92ca04017f8b1dae4/packages/create-plugin/src/commands/generate/update-go-sdk-and-packages.ts#L10-L63

So I'm not sure if we need to use renovate at all for this. Are there any other benefits that I'm glossing over if we do it via renovate instead?

sunker: Ah sweat, I wasn't aware of this. I'll close the PR then. Thanks @xnyo!

PR #1101, opened 2024-09-09
Chore: Let depedabot handle plugin-e2e dep updates

sunker:

This PR lets dependabot manage dependency updates in the plugin-e2e package.

According to plugin-e2e peer deps, the package is compatible with `>=1.41.2 <2.0.0` of Playwright. For that reason, it makes sense to always use latest version of Playwright internally when testing plugin-e2e. 

Fixes #


jackw (approved):
LGTM 🚀

PR #1127, opened 2024-09-20
Feat: Adding provenance publish config

tolzhabayev:
Linking npm package builds better to source code and CI builds using sigstore.

More context in: https://docs.npmjs.com/generating-provenance-statements

I don't think we have to force publishing new versions just because of this change. I would rather just wait for regular releases to happen to those packages after this change is merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,6 +186,9 @@ jobs:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
       NX_BRANCH: ${{ github.event.number || github.ref_name }}
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - name: Generate token
         id: generate_token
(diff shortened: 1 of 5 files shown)


jackw (approved):
LGTM! 🚀

.github/workflows/ci.yml
```diff
@@ -186,6 +186,9 @@ jobs:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
       NX_BRANCH: ${{ github.event.number || github.ref_name }}
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    permissions:
```
tolzhabayev, line 189: @jackw I've added this because I assume it will be required. fyi

jackw: Looking at [the docs](https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow) we do need to set the `id-token`. However are we using this github token or the `github-app-token` further down?

tolzhabayev: My assumption is that they are a bit different and it needs the specific id-token. I guess we will see once we run it :)

PR #1133, opened 2024-09-25
Switch to createDataFrame

### LINKED ISSUE #763: Frontend DS template and tutorial use deprecated `MutableDataFrame`
Warning message: `'MutableDataFrame' is deprecated. use standard DataFrame, or create one with PartialDataFrame`

Example run https://github.com/grafana/plugin-tools/actions/runs/7987513869

And in our docs https://grafana.com/developers/plugin-tools/tutorials/build-a-data-source-plugin#returning-data-frames

Will need to check more docs, saw in Explore queries as well

tolzhabayev:
Move from deprecated `MutableDataFrame` to the function the creates `DataFrame`'s

Fixes #763


sunker (approved):
LGTM

packages/create-plugin/templates/datasource/src/datasource.ts
```diff
@@ -77,7 +76,7 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
       }
     } catch (err) {
       let message = '';
-      if (_.isString(err)) {
+      if (typeof err === 'string') {
```
tolzhabayev, line 79: I couldn't find any other places we are using `lodash` on scaffolding so I removed it - the developers can add it back when needed and it is removed by webpack on bundle anyway.

tolzhabayev: @jackw anything against that ☝️

sunker: 👍 

This method is a bit of a mess, so would be nice to improve at some point. Maybe grafana/data could expose a function that returns message from an unknown error.

PR #1137, opened 2024-09-26
Chore: Fixing path-to-regexp

tolzhabayev:
https://github.com/grafana/plugin-tools/security/dependabot/70

Bumps path-to-regexp package and overrides it for serve-handler due to https://github.com/vercel/serve-handler/issues/211


jackw (approved):
LGMT! 🚀

PR #1140, opened 2024-09-26
[e2e-plugins] - Export PageFixture type

shelldandy:
this is useful when writing interfaces that might use the page component, since currently we'd have to use the `Page` type from playwright, but in reality the type should be PageFixture which is currentlty unexported.

This PR fixes that

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.4.0-canary.1140.8d70b63.0
  npm install @grafana/plugin-e2e@1.9.0-canary.1140.8d70b63.0
  # or 
  yarn add @grafana/create-plugin@5.4.0-canary.1140.8d70b63.0
  yarn add @grafana/plugin-e2e@1.9.0-canary.1140.8d70b63.0
  ```
</details>

### DIFF
--- a/packages/plugin-e2e/src/fixtures/page.ts
+++ b/packages/plugin-e2e/src/fixtures/page.ts
@@ -1,8 +1,5 @@
-import { TestFixture, Page } from '@playwright/test';
-import { PlaywrightArgs } from '../types';
+import { PageFixture } from '../types';
 import { overrideFeatureToggles } from './scripts/overrideFeatureToggles';
-
-type PageFixture = TestFixture<Page, PlaywrightArgs>;
 
 /**
  * This fixture ensures the feature toggles defined in the Playwright config are being used in Grafana frontend.
--- a/packages/plugin-e2e/src/types.ts
+++ b/packages/plugin-e2e/src/types.ts
@@ -6,6 +6,8 @@ import {
   PlaywrightWorkerOptions,
   Response,
   TestInfo,
+  TestFixture,
+  Page,
 } from '@playwright/test';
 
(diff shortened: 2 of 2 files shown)


sunker:
Thanks for raising this PR @shelldandy! 

Can you elaborate a little bit on where this would be useful? I wonder if this is something we could cater for within the plugin-e2e lib.

shelldandy:
@sunker sure thing, so im migrating the e2e tests for the enterprise datasources squad and we have some helper functions to DRY the code on some repeated actions:

image

So I wanted to have proper typing on the interface being used, currently im using `Page` from playwright but in reality the true type would be PageFixture

sunker:
> @sunker sure thing, so im migrating the e2e tests for the enterprise datasources squad and we have some helper functions to DRY the code on some repeated actions:

Thanks for clarifying @shelldandy! While migrating from grafana/e2e to grafana/plugin-e2e, you need to keep in mind that the two are fundamentally different. Plugin E2E was built with plugin e2e testing from the start, whereas the grafana/e2e was not. All fixtures, pages models and methods of plugin-e2e are there to simplify plugin testing, so my hope is that there should be no need for dry methods. And if there are, it should be added to the plugin-e2e lib. Any suggestions/feature requests of PRs would be very appreciated!  

Regarding the runQuery method in your example above. This looks like an entire test to me, so not sure if this is something one would want to repeat across multiple plugins? 

Also dropped a few comments in [this](https://github.com/grafana/plugins-private/pull/2391) PR.

shelldandy:
Roger that we have decided to give that approach a try, thanks!

PR #1159, opened 2024-10-02
Chore: bump uplot to 1.6.31

tolzhabayev:

Bump `@grafana/data` package to include this change https://github.com/grafana/grafana/pull/93952

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2635,9 +2635,9 @@
       }
     },
     "node_modules/@braintree/sanitize-url": {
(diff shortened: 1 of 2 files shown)


tolzhabayev:
@leventebalogh I am not sure if that is alright

jackw (approved):
LGTM! 🚀 

I can't see an issue with relying on the canary (if there's no stable version available yet). This is a dev dep and used for tests in a package that we aren't using in production right now.

PR #1167, opened 2024-10-08
Renovate: Prevent capturing template expression in version

jackw:
Create plugin update doesn't bump grafana/schema leaving packages out of sync.

image

From what I can tell Renovate doesn't update `@grafana/schema` [in PRs](https://github.com/grafana/plugin-tools/pull/1156). I can't be 💯 sure on this but the regex is capturing the trailing template expression on that line as part of the version which I think throws renovate off somewhere. By adding the trailing `,` we capture just the version information.

**Before**
image

**After**
image

[regex101 link](https://regex101.com/r/FXmhZe/1) for those that want to play around.

Fixes #

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -9,7 +9,7 @@
     {
       "customType": "regex",
       "fileMatch": ["_package.json"],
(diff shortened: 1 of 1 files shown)


leventebalogh (approved):
Nice catch! 🚀

PR #1169, opened 2024-10-08
Chore: Override cookie to latest version

jackw:
Forces an override of the npm package cookie to resolve a [dependabot alert](https://github.com/grafana/plugin-tools/security/dependabot/78).

Apologies for the noise in the lock file. NPM won't override nested dependencies if it is mentioned in the lock file. So I had to remove the docusaurus packages first to clean the lock file, add the override, then add the docusaurus packages back which brought in a load of updates.

cookie monster

Fixes #

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -199,9 +199,9 @@
       }
     },
     "node_modules/@algolia/client-common": {
-      "version": "5.6.1",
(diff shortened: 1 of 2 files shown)


tolzhabayev (approved):
🍪

PR #1170, opened 2024-10-08
CI: Switch to using renovate to handle dependency bumps

jackw:
This PR disables dependabot npm dep handling and enables renovates instead.

Why? Because I would like to not have to care about labelling dev dep bumps with `no-changelog` and let them auto merge if they pass.

I've tried my best to carry across the previous dependabot groupings and set a limit of concurrent PRs to 10. Any vulnerability alerts will be labelled with an additional `security` label.

Fixes #


.github/dependabot.yml
```diff
@@ -1,43 +1,12 @@
 # Set update schedule for GitHub Actions
-
 version: 2
 updates:
   # Check for updates to GitHub Actions every month
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
-      interval: 'monthly'
+      interval: 'weekly'
```
tolzhabayev, line 8: why weekly?

tolzhabayev (approved):
lgtm!

PR #1196, opened 2024-10-10
Website: Use esbuild to bundle js

jackw:
Just an idea to see if we can trim some time off our website builds...

Fixes #


sunker:
> Just an idea to see if we can trim some time off our website builds...

What's your findings so far?

jackw:
> What's your findings so far?

Right now it makes pretty much no difference whatsoever! :D Think I'm closing this. We can revisit if in the future we find builds are taking a long time.

PR #1197, opened 2024-10-10
CI: Use check-labels to label renovate PRs

jackw:

This PR removes the default `patch` label from renovate PRs and  updates the check-labels action to label renovate PRs with `no-changelog` or `patch` depending on if only the lock file is touched or not. The repos lock file changes don't affect consumers of packages so there is no need for those PRs to create releases or entries in changelogs. 

Additionally it removes the cancel-in-progress for the check-labels workflow as we (@ashharrison90 and myself) believe it is confusing renovate by cancelling workflow runs and preventing auto merge from working. Thanks Ash for the help! 🥇 

Fixes #

### DIFF
--- a/.github/actions/check-labels/index.js
+++ b/.github/actions/check-labels/index.js
@@ -25,6 +25,7 @@ async function run() {
 
     const files = await getPullRequestFiles({ octokit });
 
+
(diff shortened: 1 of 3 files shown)


sunker (approved):
Well done spotting this! 🚀

PR #1210, opened 2024-10-11
Create Plugin: Remove mkdirp package

jackw:
I don't think we're using any of the benefits of the mkdirp package and internally I believe it calls the native function anyway so this PR removes it.

Fixes #


leventebalogh (approved):
👏

academo (approved):

PR #1213, opened 2024-10-14
Docs: Add documentation about setMigrationHandler for panels

### LINKED ISSUE #1128: Docs: Add docs about setMigrationHandler
Add docs about how to use setMigrationHandler in panels and perhaps other plugins such as datasources

academo:

Adds a documentation page for the `setMigrationHandlers` for panels.

closes https://github.com/grafana/plugin-tools/issues/1128


docusaurus/docs/how-to-guides/panel-plugins/options-migration-handler.md
```diff
@@ -0,0 +1,151 @@
...
+  })
+  // define the migration handler
+  .setMigrationHandler(migrationHandler);
+```
+
+:::note
+
+The migration handler is only called when the stored panel version differs from the loaded panel version. (e.g. after an update)
+
+The migration handler will be called until the user manually edits and saves the panel.
```
andresmgot, line 45: maybe... 

```suggestion
The migration handler is only called when the installed plugin version differs from the version used to generate a panel. Changes made to the panel are not automatically persisted, users need to manually save the dashboard after a panel is migrated.
```

docusaurus/docs/how-to-guides/panel-plugins/options-migration-handler.md
```diff
@@ -0,0 +1,151 @@
+---
+id: migration-handler-for-panels
```
andresmgot, line 2: not sure how to structure this but I was thinking on a generic page with general information about plugin migrations (e.g. when to use it) and then split in different pages:

 - Migrations for panel plugins
 - Migration for frontend datasources
 - Migration for backend datasources
 - (Maybe in the future?) Migration for apps

I guess this is fine for the first guide but we may need to restructure?

andresmgot (commented):
Mostly LGTM! I have some small sugestions / questions.

docusaurus/docs/how-to-guides/panel-plugins/options-migration-handler.md
academo: I think it is best to keep these migrations guide under each plugin type and maybe we can create some phantom menu entries for a dedicated migrations section.

if i were a dev I'd look for information about panels under the panels section.

docusaurus/docs/how-to-guides/panel-plugins/options-migration-handler.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+### Version-specific adjustments
+
+You might want to run specific adjustments based on the previously installed plugin version.
+
+you can use the `pluginVersion` property. This property:
+
+- is empty if it is the first time a migration handler is ever called. e.g. the plugin didn't implement a migration handler before.
+- is set to the version of the plugin when the panel was last saved after a migration handler was called.
```
andresmgot, line 126:
```suggestion
You might want to base migration decissions based on the version that was used to write a panel. For this, you can use the `pluginVersion` property. This property is empty the first time the migration handler is ever used but after that it will be set to the plugin version used to save the panel.

For example, imagine the following history for a plugin:

 - In v1, the plugin didn't have any migration code.
 - In v2, the plugin introduced the first migration code.
 - In v3, the plugin changed again with new migration steps.
 
In that scenario, the migration handler would look like this:

```ts title="migrationHandler.ts"
import { config } from '@grafana/runtime';

function migrationHandler(panel: PanelModel<SimpleOptions>) {
  const options = Object.assign({}, panel.options);
  const pluginVersion = panel?.pluginVersion ?? '';

  if (pluginVersion === '') {
    // Plugin version was v1.x
    // Needs logic to migrate v1 -> v3
    options.displayMode = 'compact';
    options.displayType = 'linePlot';
  }
  
  if (pluginVersion.startsWith('2.') {
    // Panel was last saved with version v2.x
    // Needs logic to migrate v2 -> v3
    options.displayMode = 'compact';
  }

  return options;
}
```

academo: @josmperez What do you think?

andresmgot (approved):
LGTM from the technical perspective.

PR #1243, opened 2024-10-18
Fix: Dont default navigation page permissions to admin

### LINKED ISSUE #1207: Bug: App templates restrict pages to admin users

- [x] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E
- [ ] Plugin Meta Extractor

### Package versions

5.5.3

### What happened?

The plugin.json template files that are used to scaffold plugins contain page includes that are restricted to only being accessible by Admin users.

To create an app plugin with routes that are accessible by all users by default.

Restricting routes should IMHO be something a plugin developer buys into rather than accidentally creating apps that only admins have access to. Note that due to the env vars used with the developer docker env this goes unnoticed:

```
ENV GF_AUTH_ANONYMOUS_ORG_ROLE "Admin"
ENV GF_AUTH_ANONYMOUS_ENABLED "true"
ENV GF_AUTH_BASIC_ENABLED "false"
```

1. Scaffold plugin
2. Login as a viewer
3. Notice that there are no links to any app pages

### Environment

N/A

### Additional context

This likely needs addressing in our plugin-examples repo too.

jackw:
There have been a [recent bug report](https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1728381252390489) that scenes apps can only be accessed by admin users. After some back and forth with the developers it was confirmed that the plugin.json contained `role: Admin` which limits page access to only Admin users.

This PR removes these role properties from the plugin.json templates for apps and apps with scenes leaving it for a plugin developer to "buy in" to limiting custom page access to certain roles. It also cleans up the build an app tutorial which has an incorrect `roles` property and adds a tip admonition explaining how to limit access to pages.

I've intentionally left the app config page with role access as admin as I believe anything else would likely cause errors if viewers or editors attempted to access / edit the app configs.

Fixes #1207

If / when this PR is ready to merge will likely need to carry across whatever changes we make here to the plugin-examples repo app examples.

### DIFF
--- a/docusaurus/docs/tutorials/build-an-app-plugin.md
+++ b/docusaurus/docs/tutorials/build-an-app-plugin.md
@@ -77,7 +77,6 @@ Let's add a new page to the navigation menu:
            "type": "page",
            "name": "New Page",
            "path": "/a/%PLUGIN_ID%/new-page",
-           "roles": "Admin",
            "addToNav": true,
            "defaultNav": false
        }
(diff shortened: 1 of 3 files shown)


leventebalogh (approved):
LGTM 👍

PR #1250, opened 2024-10-25
Dependencies: Follow e2e-selectors modified tag

sunker:

In this PR, we follow the `grafana/e2e-selectors@modified` tag in scaffolded plugin and in the plugin-e2e package. I put them in two different package rules since they needs two different set of labels - plugin-e2e needs to be released immediately and create-plugin can probably wait. 

Fixes #

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -36,7 +36,7 @@
     },
     {
       "automerge": true,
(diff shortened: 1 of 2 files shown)


jackw (approved):
LGTM! 🚀

PR #1272, opened 2024-10-29
E2E: Improve support to interact with panel edit options

mckn:
This PR introduce an abstraction to interact with the panel edit options. This is useful when developing panel plugins and you want to verify that different combinations of settings works as expected. The goal is to make it easier and more robust to write tests that interacts with the panel editor. This will also give us a layer where we can cater for differences in the panel editor between Grafana versions. 

One example of this is the recent change of the switch component that changed role from checkbox to switch. Developers writing tests interacting with the switch directly would need to cater for this in the test which will make it harder and more complex to test your plugin against multiple versions of Grafana.

> Note: A more robust way to test this is to provision one panel per setting. So you never interact with the panel editor while running tests. You only test that the setting has the expected behaviour on your panel. This might not be possible in all scenarios e.g. when adding a custom panel editor or having cascading settings.

This is how the configuration of the panel edit options map into the test APIs.

- [x] addNumberInput -> getNumberInput
- [x] addSliderInput -> getSliderInput
- [x] addTextInput -> getTextInput
- [x] addSelect -> getSelect()
- [x] addMultiSelect -> getMultiSelect()
- [x] addRadio -> getRadio()
- [x] addBooleanSwitch -> getSwitch
- [x] addColorPicker -> getColorPicker
- [x] addTimeZonePicker -> getSelect
- [x] addUnitPicker -> getUnitPicker
- [x] addFieldNamePicker  -> getSelect
- [x] addDashboardPicker -> getSelect

Fixes #

This is a suggestion and I would like to get more eyes on this to get better naming, structure and ways to interact with the panel editor in your test.

On thing that isn't that great is that while writing the tests you use the "labels" visible in the UI to select options to interact with. This is great for the options that the plugin adds because the plugin controls the labels that they add. But for the standard options or options that is added by Grafana we can still run into the problem that a plugin developer writes a test and interacts with the `Standard options` label. If we, for some reason, change that in a Grafana version the tests will stop working and the tests will not be able to run across multiple versions of Grafana.

<details>
  <br />

  ```bash
  npm install website@3.6.0-canary.1272.059bc30.0
  npm install @grafana/create-plugin@5.14.0-canary.1272.059bc30.0
  npm install @grafana/plugin-e2e@1.15.0-canary.1272.059bc30.0
  npm install @grafana/plugin-meta-extractor@0.1.0-canary.1272.059bc30.0
  npm install @grafana/plugin-types-bundler@0.3.0-canary.1272.059bc30.0
  npm install @grafana/sign-plugin@3.1.0-canary.1272.059bc30.0
  # or 
  yarn add website@3.6.0-canary.1272.059bc30.0
  yarn add @grafana/create-plugin@5.14.0-canary.1272.059bc30.0
  yarn add @grafana/plugin-e2e@1.15.0-canary.1272.059bc30.0
  yarn add @grafana/plugin-meta-extractor@0.1.0-canary.1272.059bc30.0
  yarn add @grafana/plugin-types-bundler@0.3.0-canary.1272.059bc30.0
  yarn add @grafana/sign-plugin@3.1.0-canary.1272.059bc30.0
  ```
</details>


sunker (commented):
Nice! I think I prefer option 1 as option 2 would make us slightly more vulnerable to changes in the core Playwright api. 

Question: Would it be better to replace `input` and `switch` with `getByRole(role: 'input' | 'switch' | ..., { name: string }: Options)`? This would make our abstraction look closer to the playwright api? 

```ts
const group = panelEditPage.getOptionsGroup('Clock panel')
await group.getByRole('switch', { name: 'enableClock' }).uncheck()
```

mckn:
> Nice! I think I prefer option 1 as option 2 would make us slightly more vulnerable to changes in the core Playwright api.
> 
> Question: Would it be better to replace `input` and `switch` with `getByRole(role: 'input' | 'switch' | ..., { name: string }: Options)`? This would make our abstraction look closer to the playwright api?
> 
> ```ts
> const group = panelEditPage.getOptionsGroup('Clock panel')
> await group.getByRole('switch', { name: 'enableClock' }).uncheck()
> ```

Yeah, but on the other hand 'input' is not a role. In the scenario where with the switch-bug you would still need to know if you are trying to fetch a `combobox` or a `switch` role.

mckn:
@sunker I added some more code and tests. Lets have a sync about this before I continue further but it feels kind of OK to work with when writing tests.

sunker:
Nice! I had a quick look at the helper methods in the panel options API and this is what you can do: 

```
addNestedOptions (can skip this for now)
addNumberInput (textbox)
addSliderInput (not sure, textbox I guess?)
addTextInput (textbox or textarea)
addSelect (combobox)
addMultiSelect (combobox)
addRadio (radiogroup)
addBooleanSwitch (switch in recent versions) 
addColorPicker (don't know - can probably skip for now)
addTimeZonePicker (combobox)
addUnitPicker (comobobox)
addFieldNamePicker (combobox)
addDashboardPicker (combobox) 
```

So basically we need to support textbox, textarea, combobox, radiogroup and switch. So I guess we'll have to choose between: 
```ts
group.getByRole(role: textbox | textarea | combobox | radiogroup | switch, options: { ... })
// or implementing each specifically like in this PR. e.g
group.getRadioButtonGroup(label: string)
```

I don't have strong opinions here. The tricky bit will be to get them to work in >= 8.5.0.

mckn:
> Nice! Great tests!
> 
> I had a quick look at the helper methods in the panel options API and this is what you can do:
> 
> ```
> addNestedOptions (can skip this for now)
> addNumberInput (textbox)
> addSliderInput (not sure, textbox I guess?)
> addTextInput (textbox or textarea)
> addSelect (combobox)
> addMultiSelect (combobox)
> addRadio (radiogroup)
> addBooleanSwitch (switch in recent versions) 
> addColorPicker (don't know - can probably skip for now)
> addTimeZonePicker (combobox)
> addUnitPicker (comobobox)
> addFieldNamePicker (combobox)
> addDashboardPicker (combobox) 
> ```
> 
> So basically we need to support textbox, textarea, combobox, radiogroup and switch. So I guess we'll have to choose between:
> 
> ```ts
> group.getByRole(role: textbox | textarea | combobox | radiogroup | switch, options: { ... })
> // or implementing each specifically like in this PR. e.g
> group.getRadioButtonGroup(label: string)
> ```
> 
> I don't have strong opinions here. The tricky bit will be to get them to work in >= 8.5.0.

I'm happy with both as well. The only thing I feel is a bit negative with the `getByRole` is that it diverge from the standard `getByRole` since we are introducing "new" roles.

But I'm happy to proceed with that approach as well. Maybe ask for some more input here?

sunker:
> The only thing I feel is a bit negative with the `getByRole` is that it diverge from the standard `getByRole` since we are introducing "new" roles.

You mean because they only make up a subset of all roles that exists? 

> Maybe ask for some more input here? 

Sounds good! Maybe Jack? He's affected by the recent change in his traffic light panel. :)

mckn:
> You mean because they only make up a subset of all roles that exists?

Yes, and we would need to have custom roles that match to our components. E.g. in the scenario with the recent change the switch got changed from a 'checkbox' to a 'switch' and in those cases we would not want the developer to need to update their tests. In case they have `getByRole('checkbox', ...);`.

I understood your suggestion as we don't use the default roles but use our own that maps to our components rather than the html roles. Otherwise we don't resolve issues as the one with the switch.

packages/plugin-e2e/src/models/components/Select.ts
```diff
@@ -0,0 +1,18 @@
+import { Locator } from '@playwright/test';
+import { PluginTestCtx } from '../../types';
+
+export class Select {
```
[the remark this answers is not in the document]
sunker: Do we need this abstraction? Why not just return the Playwright locator from `getSelect`? That way they can interact with the core playwright apis directly. 

Clicking an option would be something like this:
```ts
const group = panelEditPage.getOptionsGroup('Clock panel')
// open select
await group.getSelect('Something').click()
// select an option
await panelEditPage.getByGrafanaSelector(selectors.components.Select.option).getByText('Option 1').click()
```

packages/plugin-e2e/src/models/components/PanelEditOptionsGroup.ts
```diff
@@ -0,0 +1,30 @@
...
+    // we need to add logic to select by a switch or a checkbox role depending on grafana version.
+    const id = await this.getByLabel(label).getByRole('checkbox').getAttribute('id');
+    return this.getByLabel(label).locator(`label[for='${id}']`);
+  }
+
+  getInput(label = ''): Locator {
+    return this.getByLabel(label).getByRole('textbox');
+  }
+
+  getSelect(label = ''): Select {
```
sunker, line 22: I guess a panel option cannot exist without a label? So should probably not make this arg optional. Also, we may need to pass optional options in the future so we can for example let people override playwright options such as `exact`, `disabled` etc (see second arg of playwrights native `getByRole` or `getByLabel`.

mckn: There are a couple of examples but I noticed that they occur when the label of the field is the same as the option group. Clock panel timezone is an example of this.

packages/plugin-e2e/src/models/components/Select.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async selectOption(values: string, options?: SelectOptionsType): Promise<string> {
+    const menu = await openSelect(this.element, options);
+    return selectByValueOrLabel(values, menu, options);
+  }
+}
+
+export async function openSelect(element: Locator, options?: SelectOptionsType): Promise<Locator> {
+  await element.getByRole('combobox').click(options);
+  return element.page().getByLabel('Select options menu', {
```
sunker, line 18: Use `selectors.components.Select.menu` instead

packages/plugin-e2e/src/models/components/Select.ts
```diff
@@ -0,0 +1,41 @@
...
+    exact: true,
+  });
+}
+
+export async function selectByValueOrLabel(
+  labelOrValue: string,
+  menu: Locator,
+  options?: SelectOptionsType
+): Promise<string> {
+  if (!labelOrValue) {
```
sunker, line 28: Don't need this check?

packages/plugin-e2e/src/models/pages/PanelEditPage.ts
```diff
@@ -185,4 +182,34 @@ export class PanelEditPage extends GrafanaPage {
...
+    return this.getCustomOptions('Data links');
+  }
+
+  getThresholdsOptions(): PanelEditOptionsGroup {
+    return this.getCustomOptions('Thresholds');
+  }
+
+  private getOptionsGroupLocator(label: string): Locator {
+    if (semver.gte(this.ctx.grafanaVersion, '11.4.0')) {
+      return this.ctx.page.getByTestId(`data-testid Options group ${label}`);
```
sunker, line 210: Change to `this.getByGrafanaSelector(selectors.components.OptionsGroup.group(label))`

packages/plugin-e2e/src/models/components/PanelEditOptionsGroup.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Select } from './Select';
+import { MultiSelect } from './MultiSelect';
+import { Switch } from './Switch';
+import { gte } from 'semver';
+import { RadioGroup } from './RadioGroup';
+
+export class PanelEditOptionsGroup {
+  constructor(private ctx: PluginTestCtx, public readonly element: Locator, private groupLabel: string) {}
+
+  getRadio(label: string): RadioGroup {
```
sunker, line 14: Should be called `RadioGroup` no?

packages/plugin-e2e/src/matchers/toBeSwitched.ts
```diff
@@ -0,0 +1,25 @@
+import { expect, MatcherReturnType } from '@playwright/test';
+import { getMessage } from './utils';
+
+import { Switch } from '../models/components/Switch';
+
+export async function toBeSwitched(
```
sunker, line 6: Do we need this matcher?

packages/plugin-e2e/tests/as-admin-user/panel/panelEdit.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  await clockMode.check('Time');
+  await expect(clockMode).toHaveChecked('Time');
+});
+
+test('checking switch', async ({ gotoPanelEditPage }) => {
+  const panelEdit = await gotoPanelEditPage({ dashboard: { uid: 'mxb-Jv4Vk' }, id: '5' });
+  const clockOptions = panelEdit.getCustomOptions('Clock');
+  const monospaceFont = await clockOptions.getSwitch('Font monospace');
+
+  await expect(monospaceFont).toBeSwitched({ on: false });
```
sunker, line 39: I think I rather use the standard API: 
```await expect(monospaceFont.locator).not.toBeChecked()```

Comment (sunker) [reply packages/plugin-e2e/tests/as-admin-user/panel/panelEdit.spec.ts:39]: Alternatively, we keep the custom matcher but use the same name and signature as the Playwright one:
```await expect(monospaceFont).not.toBeChecked()```

Comment (mckn) [reply packages/plugin-e2e/tests/as-admin-user/panel/panelEdit.spec.ts:39]: Yeah, I started with that but ran into problems with the standard method "toBeChecked" being overwritten.

File: packages/plugin-e2e/src/models/components/PanelEditOptionsGroup.ts (hunk 6 above)
Comment (mckn) [reply packages/plugin-e2e/src/models/components/PanelEditOptionsGroup.ts:14]: I changed it to Radio to align with the plugins API. But I agree with you just changed it to be consistent.

File: packages/plugin-e2e/src/models/components/ColorPicker.ts [hunk 9]
```diff
@@ -0,0 +1,20 @@
+import { Locator } from '@playwright/test';
+import { PluginTestCtx } from '../../types';
+import { ComponentBase } from './ComponentBase';
+import { SelectOptionsType } from './types';
+
+export type SelectableColors = 'red' | 'blue' | 'orange' | 'green' | 'yellow';
```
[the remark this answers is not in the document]
mckn: Yeah, but I just wanted to limit the options during testing to make it easier to select the proper colors. Should be good enough for tests, no?

packages/plugin-e2e/src/matchers/toHaveSelected.ts
```diff
@@ -0,0 +1,126 @@
...
+  value: string | RegExp,
+  options?: ContainTextOptions
+): Promise<MatcherReturnType> {
+  let actual = '';
+
+  try {
+    actual = await select
+      .locator('div[class*="-grafana-select-value-container"]')
+      .locator('div[class*="-singleValue"]')
+      .innerText(options);
```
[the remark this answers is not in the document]
mckn: hmm, that is a valid concern. But I think we can fix that in the `index.ts` file.

packages/plugin-e2e/src/models/components/PanelEditOptionsGroup.ts
```diff
@@ -0,0 +1,59 @@
...
+    return this.getByLabel(label).getByRole('textbox');
+  }
+
+  getNumberInput(label: string): Locator {
+    return this.getByLabel(label).getByRole('spinbutton');
+  }
+
+  getSliderInput(label: string): Locator {
+    return this.getNumberInput(label);
+  }
```
mckn, line 38: These are still returning regular Locators since they are so simple components. What are your gut feeling. Should I encapsulate them in an abstraction as well?

packages/plugin-e2e/src/models/components/Select.ts
mckn: A slightly different selector got added in 11.4

https://github.com/grafana/grafana/blob/main/packages/grafana-e2e-selectors/src/selectors/components.ts#L881

packages/plugin-e2e/src/matchers/toHaveSelected.ts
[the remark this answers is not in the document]
mckn: The `ctx` is available with that the selectors as well. So I will see if I can change these to selectors from the grafana-selectors package istead.

sunker (commented):
This is starting to look really good!

packages/plugin-e2e/src/matchers/toHaveSelected.ts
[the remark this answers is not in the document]
mckn: hmm, I don't think we pass the data-testid when rendering the select component. Or I can't find it in the dom.

.github/workflows/playwright.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         with:
           version-resolver-type: plugin-grafana-dependency
           grafana-dependency: '>=8.5.0'
-          # limit: 0 # Uncomment to test all versions since 8.5.0. Useful when testing compatibility for new APIs.
+          limit: 0
```
sunker, line 28: no that all tests pass, we can revert the change in this file :)

packages/plugin-e2e/src/matchers/toHaveSelected.ts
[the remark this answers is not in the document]
sunker: Okay, can we add a new one then?

packages/plugin-e2e/src/models/components/ColorPicker.ts
```diff
@@ -0,0 +1,27 @@
...
+    const colorInput = this.getContainer().getByTestId('input-wrapper').getByRole('textbox');
+    await colorInput.hover(options);
+    await colorInput.fill(rgbOrHex, options);
+  }
+
+  private getContainer(): Locator {
+    if (gte(this.ctx.grafanaVersion, '8.7.0')) {
+      return this.ctx.page.locator('#grafana-portal-container');
+    }
+    return this.ctx.page.locator('body > div').last();
```
sunker, line 25: if you could add a new testid in Grafana and use it here that would be great!

packages/plugin-e2e/src/models/components/Select.ts
sunker: Let's use the selector here too

packages/plugin-e2e/src/models/components/Switch.ts
```diff
@@ -0,0 +1,41 @@
...
+    return target.uncheck({ force: true, ...options });
+  }
+
+  private async getSwitch(options: CheckOptionsType): Promise<Locator> {
+    if (gte(this.ctx.grafanaVersion, '11.3.0')) {
+      return this.element;
+    }
+
+    const id = await this.element.getAttribute('id', options);
+    return this.group.locator(`label[for='${id}']`);
```
sunker, line 39: Nit: Maybe reverse the order here so that backwards compatibility becomes the "exception" rather than the default case. That way it will be easier to remove legacy code in the future.

packages/plugin-e2e/src/models/components/UnitPicker.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  async selectOption(value: string, options?: SelectOptionsType): Promise<void> {
+    await this.element.getByRole('textbox').click();
+    const option = await this.getOption(value, options);
+    await option.click(options);
+  }
+
+  private async getOption(selector: string, options?: SelectOptionsType): Promise<Locator> {
+    const steps = selector.split('>').map((step) => step.trim());
+    const container = this.ctx.page.locator('div[class="rc-cascader-menus"]');
```
sunker, line 19: Can we introduce a new data test id here too?

packages/plugin-e2e/tests/as-admin-user/panel/panelEdit.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+  const axisOptions = panelEdit.getCustomOptions('Axis');
+  const timeZonePicker = axisOptions.getSelect('Time zone');
+
+  await timeZonePicker.selectOption('Europe/Stockholm');
+  await expect(timeZonePicker).toHaveSelected('Europe/Stockholm');
+});
+
+function getPanelContent(grafanaVersion: string, page: Page): Locator {
+  if (gte(grafanaVersion, '9.5.0')) {
+    return page.locator('div[class*="-panel-content"]');
```
sunker, line 147: Can use `selectors.components.Panels.Panel.content`?

sunker (commented):
Great stuff! From my standpoint we're ready to ship this, but I think we should try to add proper selector for a few of the elements in Grafana first.

packages/plugin-e2e/src/models/components/UnitPicker.ts
mckn: I'm afraid we don't have control over the rc cascader menu. Didn't find a way to inject it with our current cascader component. Will see how we have interacted wit this component in other tests.

mckn: Can't find any tests in Grafana that tries to do anything with UnitPicker

mckn:
Everything is fixed except that I have not updated the dependency on the canary package and hence not using the selector in the ColorPicker. Will do that asap.

sunker (approved):
Great stuff!! 🕺  

I think what's left to do know is:
* update all docs and snippets so that they use the new awesome API
* add [api-test](https://github.com/grafana/grafana/tree/main/e2e/plugin-e2e/plugin-e2e-api-tests) in grafana/grafana to verify the new APIs won't break
* update create-plugin scaffolding so that the new APIs are used
* update examples repo to use the new APIs

It's totally fine with me to do this later in follow up PRs, but the docs should probably be addressed quickly after this PR is merged.

Awesome work on this!

PR #1281, opened 2024-10-31
Create Plugin: Add `buildMode` property to plugin.json asset 

sunker:

This PR adds a simple webpack plugin the inserts a `buildMode` prop to the plugin.json asset. Knowing if the plugin was build for dev or production might be useful in a different scenarios. For example, in the plugin extensions API we put error messages in the dev console in case Grafana is in dev mode. However, most Grafana contributors would consider these errors spam. It would be better if we could only log errors related to a certain plugin in case that specific plugin was built for development. 

Should `buildMode` be added to the plugin.json schema? Note that this is only added to the asset build time.

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.8.0-canary.1281.9616e88.0
  # or 
  yarn add @grafana/create-plugin@5.8.0-canary.1281.9616e88.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
@@ -11,17 +11,22 @@ export class BuildModeWebpackPlugin {
           stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONS,
         },
         async () => {
-          const pluginJsonAsset = compilation.getAsset('plugin.json');
-          const pluginJsonString = pluginJsonAsset ? pluginJsonAsset?.source.source().toString() : '{}';
-          const pluginJsonWithBuildMode = JSON.stringify(
-            {
-              ...JSON.parse(pluginJsonString),
-              buildMode: compilation.options.mode,
-            },
-            null,
-            4
-          );
-          compilation.updateAsset('plugin.json', new webpack.sources.RawSource(pluginJsonWithBuildMode));
+          const assets = compilation.getAssets();
+          for (const asset of assets) {
+            if (asset.name.endsWith('plugin.json')) {
+              console.log(asset.name);
+              const pluginJsonString = asset.source.source().toString();
+              const pluginJsonWithBuildMode = JSON.stringify(
+                {
+                  ...JSON.parse(pluginJsonString),
+                  buildMode: compilation.options.mode,
+                },
+                null,
+                4
+              );
+              compilation.updateAsset(asset.name, new webpack.sources.RawSource(pluginJsonWithBuildMode));
+            }
+          }
         }
       );
     });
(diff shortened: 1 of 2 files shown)


packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+export class BuildModeWebpackPlugin {
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: PLUGIN_NAME,
+          stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONS,
+        },
+        async () => {
+          const pluginJsonAsset = compilation.getAsset('plugin.json');
```
jackw, line 14: I know this is draft but does this work with nested plugins?

sunker: No 😱  Fixed in https://github.com/grafana/plugin-tools/pull/1281/commits/9616e88d5514f83d9f9a9967acbf993385359162

packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+      compilation.hooks.processAssets.tap(
+        {
+          name: PLUGIN_NAME,
+          stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONS,
+        },
+        async () => {
+          const assets = compilation.getAssets();
+          for (const asset of assets) {
+            if (asset.name.endsWith('plugin.json')) {
+              console.log(asset.name);
```
mckn, line 17: Should we keep this one?

packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
```diff
@@ -0,0 +1,34 @@
+import * as webpack from 'webpack';
+
+const PLUGIN_NAME = 'BuildModeWebpack';
+
+export class BuildModeWebpackPlugin {
+  apply(compiler: webpack.Compiler) {
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(
```
mckn, line 8: Is this done as late as possible in the compilation process? Just thinking it would be good to only do this if the build seems to be successful.

mckn (approved):
LGTM! Added two questions.

packages/create-plugin/templates/common/.config/webpack/BuildModeWebpackPlugin.ts
sunker: Yeah I think it's at a late state. Tested with a build that failed, and then this prop is not added to the asset.

PR #1290, opened 2024-11-01
Renovate: Restrict selector updates to plugin-e2e

sunker:

Only the plugin-e2e package needs to follow the `modified` dist-tag of `grafana/e2e-selectors`. However, currently this package rule is affecting also the `_package.json` file for scaffolded plugins.  This PR puts a restriction on the e2e-selector package rule so that it only applies to the package.json in plugin-e2e workspace. 

Fixes #


mckn (approved):
LGTM!

PR #1363, opened 2024-11-28
Document user storage

andresmgot:

Document the new user storage feature. It's common to all plugin types (panel / datasource/ app), that's why I didn't include it in a subfolder. Let me know if this is the best approach for this.

Documentation for https://github.com/grafana/grafana/issues/91664

### DIFF
--- a/docusaurus/docs/how-to-guides/add-user-storage.md
+++ b/docusaurus/docs/how-to-guides/add-user-storage.md
@@ -47,6 +47,7 @@ export function QueryEditor(props: Props): ReactElement {
   ];
   const storage = usePluginUserStorage();
   useEffect(() => {
+    // Load the default query type from user storage
     storage.getItem('queryType').then((value) => {
       if (value && !query.queryType) {
         onChange({
@@ -55,13 +56,14 @@ export function QueryEditor(props: Props): ReactElement {
         });
       }
     });
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, []);
 
   const onChangeQueryType = (type: SelectableValue<string>) => {
     onChange({
       ...query,
       queryType: type.value!,
     });
+    // Save the query type to user storage to be used by default for the next time
     storage.setItem('queryType', type.value!);
   };
 
--- a/docusaurus/docs/how-to-guides/add-user-storage.md
+++ b/docusaurus/docs/how-to-guides/add-user-storage.md
@@ -11,11 +11,11 @@ keywords:
 
 # Add user storage to your plugin
 
-Add user storage to your plugin to store user-specific data in the Grafana backend. This data is stored per user and is accessible only to that user. It will also persist independently of the user browser. Note that this data is not encrypted and should not be used to store sensitive information. The typical use case for user storage is to store user preferences or settings.
+Add user storage to your plugin to store user-specific data in Grafana. This data is stored in the Grafana database and is accessible only to that user. Note that this data is not encrypted and should not be used to store sensitive information. The typical use case for this feature is to store user preferences or other settings.
 
 :::important
 
-This feature is only available in Grafana 11.5 and later.
+This feature is only available in Grafana 11.5 and later. It also requires the feature flag `userStorageAPI` to be enabled. If a plugin uses this feature but it's not enabled in the Grafana instance, the browser `localStorage` will be used as the storage mechanism instead.
 
 :::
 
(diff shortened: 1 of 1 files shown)


docusaurus/docs/how-to-guides/add-user-storage.md
```diff
@@ -0,0 +1,80 @@
...
+    });
+  }, []);
+
+  const onChangeQueryType = (type: SelectableValue<string>) => {
+    onChange({
+      ...query,
+      queryType: type.value!,
+    });
+    // Save the query type to user storage to be used by default for the next time
+    storage.setItem('queryType', type.value!);
```
academo, line 67: I prefer we don't encourage this type of code in our docs. (non-null assertion)

```suggestion
    type.value && storage.setItem('queryType', type.value);
```

academo (approved):
I prefer we don't use the non-null assertion in the code example but the doc looks good to me. nice work :D

docusaurus/docs/how-to-guides/add-user-storage.md
```diff
@@ -0,0 +1,80 @@
...
+    {
+      label: 'Timeseries',
+      value: 'timeseries',
+    },
+    {
+      label: 'Table',
+      value: 'table',
+    },
+  ];
+  const storage = usePluginUserStorage();
```
sunker, line 48: This will result in a broken plugin in case the runtime is <11.5.0. I think all examples where new Grafana features are being used in a plugin should be backwards compatible. Can you add a runtime check to see if the hook is exposed by the runtime package?

docusaurus/docs/how-to-guides/add-user-storage.md
```diff
@@ -0,0 +1,80 @@
...
+  useEffect(() => {
+    // Load the default query type from user storage
+    storage.getItem('queryType').then((value) => {
+      if (value && !query.queryType) {
+        onChange({
+          ...query,
+          queryType: value,
```
sunker, line 55: Even though this example proves how this API can be used, I'm not sure this is a valid use case? Do we really want to overwrite the stored queryType every time the query editor is opened?

docusaurus/docs/how-to-guides/add-user-storage.md
```diff
@@ -0,0 +1,80 @@
...
+keywords:
+  - grafana
+  - plugins
+  - plugin
+  - storage
+---
+
+# Add user storage to your plugin
+
+Add user storage to your plugin to store user-specific data in the Grafana backend. This data is stored per user and is accessible only to that user. It will also persist independently of the user browser. Note that this data is not encrypted and should not be used to store sensitive information. The typical use case for user storage is to store user preferences or settings.
```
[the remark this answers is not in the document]
andresmgot: Added the extra information to the "important" note below.

docusaurus/docs/how-to-guides/add-user-storage.md
andresmgot: That would clutter the doc quite a bit. I agree that may be the best for a complete example in the examples repo but here, for this summarized example, I'd rather keep it as simple as possible (I also don't see we are doing this with other docs examples like this [usePluginLinks](https://github.com/grafana/plugin-tools/blob/userStorage/docusaurus/docs/how-to-guides/ui-extensions/create-an-extension-point.md#creating-an-extension-point-for-links)).

We are already saying that the minimum version for this is 11.5 so hope that's enough. It would be nice to have a generic section / doc for "how to use new features maintaining Grafana version compatibility" that we could link everytime we document a new feature so we don't clutter all the examples 🤔

docusaurus/docs/how-to-guides/add-user-storage.md
andresmgot: yeah, likely not, but it's as simple as it can get, people then can apply it to their more-complex scenarios. Also, just for the sake of correctness, this doesn't change the _default_ queryType every time the query editor is opened but only when the value is changed in the editor.

docusaurus/docs/how-to-guides/add-user-storage.md
sunker: No that's fair enough. I'm planning to add an article on this topic. Let's leave it like this for now 👍

docusaurus/docs/how-to-guides/add-user-storage.md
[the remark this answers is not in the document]
sunker: 👍  Added a minor suggestion

docusaurus/docs/how-to-guides/add-user-storage.md
andresmgot: that was my first option but that example is not that complete either. It's setting every panel as favorite when you click one 😄 . I thought of using the `panelId` so only one gets "favorited" but then it will mark the same panelId in all dashboards. We could use the dasboard UID... at that point I discarded the example for being too complex 😆

josmperez (approved):
I've made some suggestions in the latest commit. Please take a look. LGTM.

docusaurus/docs/how-to-guides/add-user-storage.md
sunker: Okay then :)

sunker (approved):

PR #1369, opened 2024-12-02
Create-plugin: Remove unused dependencies according to Knip report

jackw:

Attempts to clean up dependencies according to the Knip reports in Node CI workflow runs.

Fixes #

### DIFF
--- a/packages/create-plugin/templates/common/_package.json
+++ b/packages/create-plugin/templates/common/_package.json
@@ -7,9 +7,8 @@
     "test": "jest --watch --onlyChanged",
     "test:ci": "jest --passWithNoTests --maxWorkers 4",
     "typecheck": "tsc --noEmit",
-    "format": "prettier --write ./src",
     "lint": "eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .",
-    "lint:fix": "{{ packageManagerName }} run lint{{#if isNPM}} --{{/if}} --fix",{{#if usePlaywright}}
+    "lint:fix": "{{ packageManagerName }} run lint{{#if isNPM}} --{{/if}} --fix && prettier --write --list-different .",{{#if usePlaywright}}
     "e2e": "playwright test",{{/if}}{{#if useCypress}}
     "e2e": "{{ packageManagerName }} exec cypress install && {{ packageManagerName }} exec grafana-e2e run",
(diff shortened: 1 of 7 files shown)


packages/create-plugin/templates/common/_package.json
```diff
@@ -7,6 +7,7 @@
     "test": "jest --watch --onlyChanged",
     "test:ci": "jest --passWithNoTests --maxWorkers 4",
     "typecheck": "tsc --noEmit",
+    "format": "prettier --write ./src",
```
tolzhabayev, line 10: wouldn't we want to just add this 
https://github.com/prettier/prettier-eslint as devDependency

and then `eslint --fix` would also fix the prettier errors alongside the others?

jackw: Is that an eslint plugin? Looks like it's a module to use within a node script. 🤔 

Also see https://prettier.io/docs/en/integrating-with-linters and  https://github.com/grafana/eslint-config-grafana/pull/22 which shows how slow the eslint-prettier plugin is.

How about we add the `prettier --write` to `lint:fix` rather than introduce a `format` command?

leventebalogh: > How about we add the prettier --write to lint:fix rather than introduce a format command?

I think that's a good idea! 👍

leventebalogh (approved):
Nice, LGTM 👍

packages/create-plugin/templates/common/_package.json
tolzhabayev: Yeah it used to be "the way" a couple of years ago, but since then things have seem to have changed in the JS world. Kind of making it unified by surfacing prettier errors during linting and fixing them same way too.

I want us to add the `lint` step into our internal plugin github action workflows and for that we need to standardize on a command(s) to run (https://github.com/grafana/plugin-ci-workflows/issues/9)

I am happy with just running `npm run lint` and ignoring the prettier completely, I have no strong opinions if it should be strictly enforced or just an optional "run from time to time" thing.

So fine for me to add `prettier --write` to `lint:fix` or even keep it separate if that makes more sense.

tolzhabayev: I think adding `prettier --write` to `lint:fix` is a good middle ground so let's go for that

Issue #952, opened 2024-06-11
Docs: Add guidance on depending on a plugin for local development


sympatheticmoose:
### Which areas does this feature request relate to

- [ ] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E
- [ ] Plugin Meta Extractor
- [X] Documentation

### Problem

There is no documentation for creating a plugin which depends on an already existing plugin

### Solution

We should explain the following about how create plugin's development env works: 

- the plugin’s dist folder (built plugin) is linked to /var/lib/grafana/plugins/{{ pluginId }}
- once grafana server boots it runs a plugin loader [Load function](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) which scans this folder for plugins and initializes them (runs them)
- this path has no dependency verification although other plugin.json validation, manifest and signatures checks are happening

So in order to install the plugin with it’s dependency you would need to either add the dependency into GF_INSTALL_PLUGINS=dependency (which devs probably are doing) or build and install your whole plugin via grafana cli or GF_INSTALL_PLUGINS=your-plugin-zip-via-url (which is quite inconvenient, but would work)

### Are you interested in contributing the solution?

- [ ] Yes
- [ ] No

josmperez:
Adding to Q3 roadmap.

PR #1375, opened 2024-12-03
Docs: Add info about dependencies for local development

josmperez:
Add info about working with dependencies for local development

Note: I'm aware that there are some ambiguities in the text that need to be resolved. 
- First, where are the environment variables added? 
- Second, for GF_INSTALL_PLUGINS=dependency and GF_INSTALL_PLUGINS=your-plugin-zip-via-url can we provide an example? Or use placeholder syntax for placeholders?

Fixes # https://github.com/grafana/plugin-tools/issues/952

### DIFF
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -442,7 +442,7 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
 ```
 ## Add a dependency for local development
 
-Grafana’s plugin development environment is designed to simplify the process of creating and testing plugins. Here’s how it operates:
+Grafana’s plugin development environment is designed to allow for plugins to be dependent upon other plugins if necessary. Here’s how it operates:
 
 Plugin distribution path:
 
@@ -452,7 +452,7 @@ Plugin initialization:
 
 When the Grafana server starts, its plugin loader runs a [Load](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) function. This scans the `/var/lib/grafana/` plugins directory for plugins, validates `plugin.json` files, checks signatures and manifests, and initializes valid plugins.
 
-Dependency management:
+Plugin dependency management:
 
 Although some validations occur on the server side, the plugin path itself doesn't include dependency validation. Developers are responsible for ensuring dependencies are correctly managed.
 
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -468,8 +468,6 @@ For example:
 
 tells Grafana to install that the clock panel plugin.
 
-Note that this approach is commonly used during development but requires manual tracking and updates for dependencies.
-
 ### Build and install the plugin via the Grafana CLI
 
 * Use `GF_INSTALL_PLUGINS=your-plugin-zip-via-url` for installation.
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -454,7 +454,7 @@ When the Grafana server starts, its plugin loader runs a [Load](https://github.c
 
 Plugin dependency management:
 
-Although some validations occur on the server side, the plugin path itself doesn't include dependency validation. Developers are responsible for ensuring dependencies are correctly managed.
+Although some validations occur on the server side, the plugin path itself doesn't include validation of other plugins your plugin may be dependent upon. You are responsible for ensuring that your plugin's dependencies are correctly managed.
 
 To manage dependencies during local plugin development, you can use one of the following methods:
 
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -450,7 +450,7 @@ Refer to [Grafana documentation](https://grafana.com/docs/grafana/latest/setup-g
 
 Plugin initialization:
 
-When the Grafana server starts, its plugin loader runs a [Load](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) function. This scans the `/var/lib/grafana/` plugins directory for plugins, validates `plugin.json` files, checks signatures and manifests, and initializes valid plugins.
+When the Grafana server starts, its plugin loader runs a [Load](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) function. This scans the defined [plugins directory](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#plugins) for plugins, validates `plugin.json` files, checks signatures and manifests, and initializes valid plugins.
 
 Plugin dependency management:
 
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -462,6 +462,12 @@ To manage dependencies during local plugin development, you can use one of the f
 
 * Add dependencies explicitly to your Grafana instance with `GF_INSTALL_PLUGINS=dependency`.
 
+For example:
+
+`GF_INSTALL_PLUGINS=grafana-clock-panel`
+
+tells Grafana to install that the clock panel plugin.
+
 Note that this approach is commonly used during development but requires manual tracking and updates for dependencies.
 
 ### Build and install the plugin via the Grafana CLI
--- a/docusaurus/docs/get-started/set-up-development-environment.mdx
+++ b/docusaurus/docs/get-started/set-up-development-environment.mdx
@@ -12,6 +12,7 @@ description: Set up your development environment with Docker for Grafana plugin 
   - continuous integration
   - automation
   - configuration
+  - dependency
 sidebar_position: 20
 ---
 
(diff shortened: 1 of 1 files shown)


academo:
> First, where are the environment variables added?

Environment variables are defined in your terminal environment, usually in  your profile file, which can be bashrc, zshrc, and many others, or by exporting them before running commands `export FOO=bar`

Working with environment variables is a well-known thing in development and IMO we should not be explaining that concept to developers, it'd be close to tell them how to install an editor or IDE, or how to install and open a terminal. 

Furthermore, there are so many possible variations of how to do it that whatever we write most likely won't fit the case of the person reading, so it is simpler to say "set an environment variable" and let the developer do it its own way.

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
 -"dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
 +"dev": "webpack -w -c ./webpack.config.ts --env development",
 ```
+## Add a dependency for local development
+
+Grafana’s plugin development environment is designed to simplify the process of creating and testing plugins. Here’s how it operates:
+
+Plugin distribution path:
+
+The plugin’s dist folder (built plugin) is linked to `/var/lib/grafana/plugins/{{pluginId}}` in the Grafana server environment.
```
academo, line 449: I don't think is a good idea to be naming paths like this in the documentation because this can become very confusing quickly.

This text is true, but only if you are running grafana from a binary, in linux or mac. if you are running it with docker this is not longer valid because docker configuration often mount local folders, or if you clone and build the grafana repository it is also not the case. 

Better to link the official documentation https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#plugins

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+Plugin initialization:
+
+When the Grafana server starts, its plugin loader runs a [Load](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) function. This scans the `/var/lib/grafana/` plugins directory for plugins, validates `plugin.json` files, checks signatures and manifests, and initializes valid plugins.
```
academo, line 453: again, let's not mention paths specifically here
```suggestion
When the Grafana server starts, its plugin loader runs a [Load](https://github.com/grafana/grafana/blob/b53f14ca83eeec6b0f624c2dbbaa417d08bbadbf/pkg/plugins/manager/loader/loader.go#L55-L56) function. This scans the defined [plugins directory](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#plugins) for plugins, validates `plugin.json` files, checks signatures and manifests, and initializes valid plugins.
```

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+Dependency management:
+
+Although some validations occur on the server side, the plugin path itself doesn't include dependency validation. Developers are responsible for ensuring dependencies are correctly managed.
```
academo, line 457: I generally don't understand where this is going in terms of dependency management, maybe what we want to say here is that grafana doesn't `resolve` dependencies automatically?

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+To manage dependencies during local plugin development, you can use one of the following methods:
+
+### Use the `GF_INSTALL_PLUGINS` environment variable
+
+* Add dependencies explicitly to your Grafana instance with `GF_INSTALL_PLUGINS=dependency`.
```
academo, line 463: an example of this is any plugin we have in the catalog, so

`GF_INSTALL_PLUGINS=grafana-clock-panel`

tells grafana to install that plugin.

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+Note that this approach is commonly used during development but requires manual tracking and updates for dependencies.
```
academo, line 465: It is not clear what does `requires manual tracking and updates for dependencies.` mean. what is the intention in this text?

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,32 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+### Build and install the plugin via the Grafana CLI
```
academo, line 467: this section is not using grafana CLI at all

academo (commented):
I am not sure i understand the idea behind this document, my main problem is that I don't know what do we mean by "dependency". can you maybe clarify that?

sympatheticmoose:
> I am not sure i understand the idea behind this document, my main problem is that I don't know what do we mean by "dependency". can you maybe clarify that?

@academo Does the issue help or is it still unclear? https://github.com/grafana/plugin-tools/issues/952

academo:
> > I am not sure i understand the idea behind this document, my main problem is that I don't know what do we mean by "dependency". can you maybe clarify that?
> 
> @academo Does the issue help or is it still unclear? #952

yes, that makes sense! I was almost sure we were talking about plugin depending on other plugins but in the current docs I could not see where it says that. If I skipped it and didn't read it, then maybe we should repeat it a couple times to not put things out of context?

It is easy to think of a plugin dependency as a library dependency, or grafana version dependency only

sympatheticmoose:
> It is easy to think of a plugin dependency as a library dependency, or grafana version dependency only

💯

josmperez:
> > > I am not sure i understand the idea behind this document, my main problem is that I don't know what do we mean by "dependency". can you maybe clarify that?
> > 
> > 
> > @academo Does the issue help or is it still unclear? #952
> 
> yes, that makes sense! I was almost sure we were talking about plugin depending on other plugins but in the current docs I could not see where it says that. If I skipped it and didn't read it, then maybe we should repeat it a couple times to not put things out of context?
> 
> It is easy to think of a plugin dependency as a library dependency, or grafana version dependency only

I've made a commit that clarifies that this additional section pertains specifically to plugins that depend on other plugins. Does this address your concerns? Do you see any other changes necessary?

academo:
> > > > I am not sure i understand the idea behind this document, my main problem is that I don't know what do we mean by "dependency". can you maybe clarify that?
> > > 
> > > 
> > > @academo Does the issue help or is it still unclear? #952
> > 
> > 
> > yes, that makes sense! I was almost sure we were talking about plugin depending on other plugins but in the current docs I could not see where it says that. If I skipped it and didn't read it, then maybe we should repeat it a couple times to not put things out of context?
> > It is easy to think of a plugin dependency as a library dependency, or grafana version dependency only
> 
> I've made a commit that clarifies that this additional section pertains specifically to plugins that depend on other plugins. Does this address your concerns? Do you see any other changes necessary?

I left other comments that also require some changes or clarification.

docusaurus/docs/get-started/set-up-development-environment.mdx
josmperez: done.

docusaurus/docs/get-started/set-up-development-environment.mdx
josmperez: i've tweaked this paragraph to make it clearly refer to plugin dependencies.

docusaurus/docs/get-started/set-up-development-environment.mdx
josmperez: added this example.

docusaurus/docs/get-started/set-up-development-environment.mdx
josmperez: Because it's not clear, I may have misunderstood something. Removing note.

docusaurus/docs/get-started/set-up-development-environment.mdx
josmperez: removed erroneous phrase

josmperez:
> > First, where are the environment variables added?
> 
> Environment variables are defined in your terminal environment, usually in your profile file, which can be bashrc, zshrc, and many others, or by exporting them before running commands `export FOO=bar` but it is much different if you are defining environment variables for a production server for example.
> 
> Working with environment variables is a well-known thing in development and IMO we should not be explaining that concept to developers, it'd be close to tell them how to install an editor or IDE, or how to install and open a terminal.
> 
> Furthermore, there are so many possible variations of how to do it that whatever we write most likely won't fit the case of the person reading, so it is simpler to say "set an environment variable" and let the developer do it its own way.

Thanks for the thorough review. I've made corrections/changes in response to your comments. Please re-review.

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,36 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
 -"dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
 +"dev": "webpack -w -c ./webpack.config.ts --env development",
 ```
+## Add a dependency for local development
```
academo, line 443:
```suggestion
## Add a plugin as a dependency for local development
```

docusaurus/docs/get-started/set-up-development-environment.mdx
```diff
@@ -439,3 +440,36 @@ Update the `scripts` in the `package.json` to use the extended Webpack configura
...
+
+Grafana’s plugin development environment is designed to allow for plugins to be dependent upon other plugins if necessary. Here’s how it operates:
+
+Plugin distribution path:
```
academo, line 447:
```suggestion
Plugins location directory:
```

academo (commented):
This looks better but, are we somewhere actually putting how to define the dependency in the `plugin.json`?

So far what we have tells the user they can define a dependency and (maybe?) they have to install it themselves, but how do they define the dependency?

josmperez:
> This looks better but, are we somewhere actually putting how to define the dependency in the `plugin.json`?
> 
> So far what we have tells the user they can define a dependency and (maybe?) they have to install it themselves, but how do they define the dependency?

What's the fix for this? Can you make a suggestion that gives an example please?

Issue #1254, opened 2024-10-25
EPIC: `create-plugin update` command improvements


tolzhabayev:
### Goals

We would like to improve the `create-plugin update` command so it can do more sophisticated updates to the plugins. We are planning to call these tailored and more sophisticated updates **"migrations"**.

**Migrations could:**
- update github workflows
- update specific NPM dependencies, or groups of NPM dependencies
- update and/or add files in the project (not just the ones inside the .config folder)
- update package.json scripts

Another key goal is to make it possible to **reliably test changes to the update command**, something that we are lacking now. As from now on mutating operations can only happen through migration scripts, we will achieve this goal by testing the migration scripts.

### Tasks
- [x] Define the metadata for the migrations (id, description, information on which versions to run it for)
- [x] Create a migration manager to orchestrate the entire update process when the command is run.
- [x] Create a context for the migration manager to perform file system operations. It should provide a way to make it easy
- [x] Adjust log messages so they display the necessary information in an easy-to-read format
- [x] Create a way to test migration scripts
- [x] Add an automated test that checks if the migration scripts are idempotent
- [x] Create a way to share helpers with the migration scripts: the goal is to make it possible for devs writing migrations to perform the most common tasks: updating dependencies, formatting code, running package manager commands, build config files from the handlebars templates.
- [x] Add an eslint rule that forbids migrations from importing `node:fs` to make sure they always use migration manager context

### Ongoing work
The ongoing work is currently available on the [**`updates-as-migrations-pt1`**](https://github.com/grafana/plugin-tools/compare/main...updates-as-migrations-pt1) branch.

If we want we can incrementally roll these things out without changing the current update command behaviour and use an arg of some sort to fine tune before flipping the switch.

Fixes: #820  
Related: 
- #261
- #61

leventebalogh:
I think now we basically have the possibility to all the things that we outlined in the issue (even if only behind a `--experimental-updates` flag for now). Is there any reason to keep this issue still open?

Issue #820, opened 2024-03-14
Bug: create plugin update clobbers overridden build/build:dev/dev scripts


sd2k:
### Which package(s) does this bug affect?

- [X] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E

### Package versions

```shell
@grafana/create-plugin@4.2.2
```

### What happened?

I have a custom webpack.config.ts so modified the scripts in my package.json to point to `./webpack.config.ts` instead of `./.config/webpack/webpack.config.ts`. When I run `yarn create @grafana/plugin update`, it always overrides the scripts back to using the original webpack config file.

### What you expected to happen

If the scripts have been customised, they should be ignored by the update process.

### How to reproduce it (as minimally and precisely as possible)

1. Create a plugin
2. Modify the build script to point to a different webpack config
3. Run `yarn create @grafana/plugin update`
4. Notice that the build script has been modified

### Environment

```shell
❯ npx envinfo --system --binaries --browsers --npmPackages="@grafana/*"

  System:
    OS: macOS 14.4
    CPU: (8) arm64 Apple M1
    Memory: 474.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/Library/Caches/fnm_multishells/30971_1710406991088/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/30971_1710406991088/bin/yarn
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/30971_1710406991088/bin/npm
  Browsers:
    Chrome: 122.0.6261.129
    Firefox: 123.0.1
    Safari: 17.4
  npmPackages:
    @grafana/eslint-config: ^6.0.1 => 6.0.1
```
```


### Additional context

_No response_

sympatheticmoose:
@jackw mind taking a quick look? If just needs putting on the backlog as a bug feel free to do so and remove yourself as assignee if not able to look at in the near term

jackw:
This is a known issue that we're gonna need to put some thought into to help devs selectively update the various parts of their project when running the command. I'm gonna remove myself for the time being and try to get a proposal together on how that could look.

jackw:
The initial update to 6+ might clobber the script but once done this issue should on longer happen as any future updates will only run new migrations which won't clobber scripts. I'm going to go ahead and close this.

PR #1403, opened 2024-12-12
Custom webpack

idastambuk:
Hey all, I ran `create-plugin update` in one of our plugins that had a webpack.config.ts in its root with some plugin-specific loaders. The script overwrote them to use the ./config/webpack.config.ts, which made some build steps in the CI fail. 
I'm not sure if this is the best way to do it, but I tried to add a check that, if the plugin already has a `webpack.config.ts` file, it will use that file in `build` and `dev` scripts. 

Fixes #

### DIFF
--- a/packages/create-plugin/src/types.ts
+++ b/packages/create-plugin/src/types.ts
@@ -26,5 +26,6 @@ export type TemplateData = {
   reactRouterVersion: string;
   usePlaywright: boolean;
   useCypress: boolean;
+  hasCustomWebpackConfig: boolean;
   pluginExecutable?: string;
 };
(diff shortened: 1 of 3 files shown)


jackw:
@idastambuk we really appreciate you looking into a fix for this. It's a known issue that was [reported here](https://github.com/grafana/plugin-tools/issues/820). However due to how the update command currently operates and, as you've witnessed, breaks things in multiple ways we've decided to [rewrite it so it's smarter](https://github.com/grafana/plugin-tools/issues/1254) when it comes to making changes.

Given we already have a way forward that be implemented soon and will solve this issue (along with the others listed in the epic) I'm going to go ahead and close this.

PR #1414, opened 2024-12-18
feat: add create plugin version to sign request

### LINKED ISSUE #1413: Extend sign plugin request - add more debugging information
**Why?**

In order to offer better support for signature related issues we should additionally to sign-plugin version collect the version of create-plugin tooling being used.

**How?**
- Extend GCOM to accept the new optional field ([Issue](https://github.com/grafana/grafana-com/issues/12313))
- Extend sign plugin logic to fetch the `version` from `.config/.cprc.json` file similar to [version collection](https://github.com/grafana/plugin-tools/blob/2086d38bb26fa0b8b300073a6e355437ee1b6399/packages/sign-plugin/src/commands/sign.command.ts#L31C14-L31C24)

**To consider**
The collection should not impede with anything and if the information can not be collected - json can not be parsed, cprc is missing or in a wrong format, version can not be parsed, no access, whatever - it should not block the signature requests but just omit this new optional field.

s4kh:

Add create-plugin tooling version to sign plugin request - better support for signature related issues

Fixes https://github.com/grafana/plugin-tools/issues/1413

https://github.com/grafana/grafana-com/pull/12324 should be merged before


packages/sign-plugin/src/utils/getCreatePluginVersion.ts
```diff
@@ -0,0 +1,21 @@
+import { existsSync, readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = fileURLToPath(new URL('.', import.meta.url));
+
+export const getCreatePluginVersion = () => {
+  try {
+    const crpcJSONPath = resolve(__dirname, '.config', '.crpc.json');
```
jackw, line 9: Hmm... are you sure this resolves to the plugins `.cprc.json` file? 🤔

s4kh: Yes, but only if you run the `npx @grafana/sign-plugin@latest` inside the plugin directory.

packages/sign-plugin/src/utils/getCreatePluginVersion.ts
```diff
@@ -0,0 +1,18 @@
+import { existsSync, readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+export const getCreatePluginVersion = (distDir: string) => {
```
s4kh, line 4: @jackw not sure how we can get the `.cprc.json` when custom dist directory is passed.
```
npx @grafana/sign-plugin@latest --distDir path/to/directory

```

jackw: Do we need to know where the dist directory is? Generally the `./.config/.cprc.json` is in the root of the project where the command is run. Would `process.cwd()` not work to get the location of the `./.config/.cprc.json` file?

s4kh: My understanding is that `dist/` is usually inside the plugin's root directory where the `.config/.cprc.json` is located, thus one up `../` would be the plugin's root, hopefully. If they run the script outside plugin's root directory and passed in `distDir` directory is not in the project's root, finding `.cprc.json` somewhat impossible.

oshirohugo (approved):

PR #1418, opened 2024-12-19
Create Plugin: Change scaffolded tests

### LINKED ISSUE #1408: Feat: Scenes App e2e tests should run in CI
Looks like we filter out running e2e tests for scenes-apps in our [ci gh workflow](https://github.com/grafana/plugin-tools/blob/main/.github/workflows/ci.yml#L149-L167) but we have [tests in the templates](https://github.com/grafana/plugin-tools/tree/main/packages/create-plugin/templates/scenes-app/tests).

 We should remove these checks and make sure the smoke tests run in our pipeline.

sunker:

This PR makes a few changes to the scaffolded tests: 
* It adds a few more smoke tests for some of the scaffolded plugin types.
* It adds an env variable to control anonymous auth. That way it can be disabled when running the e2e tests in CI. 
* It makes sure also the e2e tests for scenes are run in CI. 

Fixes https://github.com/grafana/plugin-tools/issues/1408

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,8 @@ jobs:
 
       - name: Wait for grafana server
         uses: grafana/plugin-actions/wait-for-grafana@main
+        with:
+          url: http://localhost:3000/login
 
       - name: Run e2e tests
         id: run-e2e-tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,7 +129,6 @@ jobs:
         if: ${{ matrix.hasBackend == true }}
 
       - name: Install playwright dependencies
-        if: ${{ matrix.workingDir != 'myorg-nobackend-scenesapp' }}
         run: npm exec playwright install chromium
         working-directory: ./${{ matrix.workingDir }}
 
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -147,22 +146,20 @@ jobs:
...
-        if: ${{ matrix.workingDir != 'myorg-nobackend-scenesapp' }}
-        run: docker compose up -d
+        run: ANONYMOUS_AUTH_ENABLED=false docker compose up -d
         working-directory: ./${{ matrix.workingDir }}
 
       - name: Wait for grafana server
         uses: grafana/plugin-actions/wait-for-grafana@main
-        if: ${{ matrix.workingDir != 'myorg-nobackend-scenesapp' }}
+        with:
+          url: http://localhost:3000/login
```
jackw, line 155: For my own understanding - why do we need this?

jackw (approved):
LGTM! 🚀

.github/workflows/ci.yml
sunker: By default, the `wait-for-grafana` Action completes when the server responds with a 200 on  `http://localhost:3000/`. However, now that anonymous auth is disabled (see [here](https://github.com/grafana/plugin-tools/blob/0f845843061c087fd1f30069d58ff1f2fd13bda6/.github/workflows/ci.yml#L149)) the server will respond with a 302, causing this Action to eventually timeout. So basically just changing the url to poll. When the login page returns a 200, the server is ready.

PR #1419, opened 2024-12-19
CI: Run e2e tests against two Grafana versions

sunker:

This PR extends the CI workflow so that the e2e tests are not only run against the latest version of Grafana, but also against the lowest supported Grafana version. In addition to that, the reports are now being deployed to GH pages. 

Note that I have forced two tests to fail just to showcase what it looks like (see PR comment below).

Fixes #

### DIFF
--- a/packages/create-plugin/templates/common/playwright.config
+++ b/packages/create-plugin/templates/common/playwright.config
@@ -21,8 +21,6 @@ export default defineConfig<PluginOptions>({
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',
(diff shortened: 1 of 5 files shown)


packages/create-plugin/templates/common/playwright.config
```diff
@@ -21,8 +21,6 @@ export default defineConfig<PluginOptions>({
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
```
sunker, line 25: This speeds up the CI process a bit. Seems to work fine in GH workflows

mckn (commented):
LGTM! Should we also add a check if MIN_VERSION and LATEST_VERSION is the same, then we probably can skip running MIN_VERSION tests.

sunker:
> LGTM! Should we also add a check if MIN_VERSION and LATEST_VERSION is the same, then we probably can skip running MIN_VERSION tests.

@mckn I don't think it can be the same. We set `grafanaDependency` manually and it should always be the last minor of the previous major.

mckn (approved):
LGTM!

PR #1435, opened 2025-01-07
Website: Use .env files with Docusaurus config

jackw:

This PR replaces the multiple docusaurus config files with a single file and .env files to set specific dev / prod config properties. It also moves the redirects into a json file to reduce the size of the config file and make it easier for docs team to manage without getting caught up in docusaurus config.

Not sure we _need_ this PR but the multiple inherited configs feel overly complex for the small number of variables we need to change per env.

The only visible change this PR should (currently) make is the `announcementBar` now appears in both dev and prod envs. I don't see any harm in it appearing in both envs but if this was intentionally only available on prod let me know and I'll sort it.

Fixes #


docusaurus/website/docusaurus.config.ts
```diff
@@ -1,35 +1,105 @@
...
+  url: PORTAL_URL,
   baseUrl: 'developers/plugin-tools/',
-  plugins,
+  onBrokenLinks: 'throw',
+  onBrokenMarkdownLinks: 'warn',
+  favicon: 'img/favicon.png',
+  // GitHub pages deployment config.
+  // If you aren't using GitHub pages, you don't need these.
+  organizationName: 'grafana', // Usually your GitHub org/user name.
+  projectName: 'plugin-tools', // Usually your repo name.
```
jackw, line 37: @tolzhabayev we don't need this anymore, right? 🤔

docusaurus/website/.env.development
```diff
@@ -0,0 +1,10 @@
+# This file is used to set environment variables for development deployments of the website.
+# DO NOT PUT SENSITIVE INFORMATION IN THIS FILE.
+DEV_PORTAL_HOST=grafana-dev.com
+DEV_PORTAL_RUDDERSTACK_URL=https://rs.grafana-dev.com
+DEV_PORTAL_RUDDERSTACK_WRITE_KEY=1w02fcWseyqcwsJA9CSKRkfEOfU
```
sunker, line 5: Just to be totally sure, this key is not a secret?

sunker (approved):
Looks good to me, but probably good if someone who's been more involved in website setup take a look too.

docusaurus/website/.env.development
jackw: I don't believe it's a secret. It's a [unique id for the source](https://www.rudderstack.com/docs/resources/glossary/#write-key) of the events. It is passed through to the frontend and can be accessed via the browser network tab if someone were inclined to steal it.

image

docusaurus/website/docusaurus.config.ts
tolzhabayev: Correct, I think it can be removed

tolzhabayev (approved):

PR #1439, opened 2025-01-08
CI: Use NPX when scaffolding plugin

jackw:

This PR changes the "Generate plugin" step in the ci gh workflow to use NPX. This brings the CI closer to the command a user would run to scaffold a plugin which allows create-plugin to detect NPM as the package manager rather than falling back to yarn.

Fixes #


leventebalogh (approved):
LGTM 👍

PR #1478, opened 2025-01-24
Chore: Improve ESlint config and fix NX config

jackw:

This PR adds missing ignores to the eslint config which currently watches hundreds of nx cache js files along with dist and create plugin template files.

Additionally I spotted multiple issues with old input paths for nx cache invalidation which are fixed in this PR too.

Fixes #


leventebalogh (approved):
LGTM 👍

PR #1479, opened 2025-01-24
Feat: Create Plugin updates as migrations

jackw:

This PR introduces the building blocks to allow developers to write migrations, focused on a specific update task, which are run sequentially via the `create-plugin` update command. This is hidden behind a flag `--experimental-updates` which should allow us to iterate further on this without impacting current usage.

See [epic](https://github.com/grafana/plugin-tools/issues/1254) for more info and [design doc](https://docs.google.com/document/d/1A0w29hewy16pKiNgnBVlk2LTlV3FtAOvHvSp1iL50BQ/edit?tab=t.0#heading=h.5sybau7waq2q) for even more more info.

[Read more about how to add migrations in the CONTRIBUTION guide &rarr;](https://github.com/grafana/plugin-tools/blob/updates-as-migrations-pt1/packages/create-plugin/CONTRIBUTING.md#migrations)

Related #1254 

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.18.0-canary.1479.13372846760.0
  # or 
  yarn add @grafana/create-plugin@5.18.0-canary.1479.13372846760.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/CONTRIBUTING.md
+++ b/packages/create-plugin/CONTRIBUTING.md
@@ -33,7 +33,8 @@ npm install
 ├── src // Executable code
 │   ├── bin // the entrypoint file
 │   ├── commands // Code that runs commands
-│   └── utils // Utilities used by commands
+│   ├── utils // Utilities used by commands
+│   └── migrations // Migrations for updating create-plugins
 └── templates // Handlebars templates
     ├── _partials // Composable parts of a template
     ├── app // Templates specific to scaffolding an app plugin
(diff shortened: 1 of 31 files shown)


mckn (commented):
Great work on this! Looks really great 🙌🏻 I have some thoughts follow up questions.

- Have you considered having two functions as part of the migration script. One that does the migration and another one that checks if the migration is applicable. Then in the migration manager we could check if a migration is applicable prior to running it. Benefit would be so that you can skip a migration e.g. if the current plugin doesn't have/use that part of the scaffolded plugin.
- Instead of having the version could we just have an array of migrations? Then they will automatically be ordered and you can use the function that checks if a migration is applicable to check version compatibility.
- What is the thoughts regarding applying changes to the create-plugin. Are we planning on using migrations to change the create-plugin project as well?

packages/create-plugin/package.json
```diff
@@ -37,7 +37,7 @@
     "generate-datasource-backend": "tsc && npm run clean-generated && CREATE_PLUGIN_DEV=true node ./dist/bin/run.js --pluginName='Sample datasource' --orgName='sample-org' --pluginType='datasource' --hasBackend",
     "lint": "eslint --cache ./src",
     "lint:fix": "npm run lint -- --fix",
-    "test": "vitest",
+    "test": "vitest --passWithNoTests",
```
academo, line 40: can we instead make it ignore the particular example test?

academo (approved):
excellent work on this! I only left a couple nits.

packages/create-plugin/src/commands/update.migrate.command.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    console.log(`Running migrations from ${projectCpVersion} to ${packageCpVersion}.`);
+
+    const commitEachMigration = argv.commit;
+    const migrations = getMigrationsToRun(projectCpVersion, packageCpVersion);
+    await runMigrations(migrations, { commitEachMigration });
+    printHeader('the update command completed successfully.');
+    console.log('');
+  } catch (error) {
+    printHeader('the update command encountered an error.');
```
academo, line 26: add a note here to instruct the user they can run `git reset hard --origin [ref]` (calculate the ref using git) so users don't get lost in the case of error.

Maybe even add a link to our github repository for them to report the issue?

mckn (approved):
LGTM!

packages/create-plugin/package.json
jackw: Thanks for pointing this out. I should have noted that it is temporary and will be removed when we add the first migration to the codebase. 

The issue isn't the example test it's that right now we have no migrations which causes the [migrations.test.ts](https://github.com/grafana/plugin-tools/blob/updates-as-migrations-pt1/packages/create-plugin/src/migrations/migrations.test.ts) that validates migrationScript paths are valid fails with `no test found in suite`. I can ignore that file temporarily but it felt like the lesser of two evils to use this flag instead given what "temporary" means. 😄

packages/create-plugin/CONTRIBUTING.md
```diff
@@ -81,3 +82,118 @@ _Work in progress._
...
+
+1. Checks the current project version (`projectCpVersion`) and package version (`packageCpVersion`).
+2. If the project version is greater than or equal to the package version, exits early.
+3. Identifies all migrations needed between the project version and package version.
+4. Executes migrations sequentially.
+5. If the `--commit` flag is passed, it will commit changes after each migration.
+
+#### How to add a migration?
+
+1. Create a new migration script file with a descriptive name (e.g. `add-webpack-profile.ts`)
```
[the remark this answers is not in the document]
jackw: Yeah I like it to. 👍 

From my experience with db migrations they tend to prefix with timestamp. Do we wanna go for that or shall we keep it simple with `000-` or maybe the targeted create plugin version? Not really sure what the benefits are over keeping it simple and incrementing from `000`. 🤷

PR #1563, opened 2025-02-21
POC: codegen from plugin.json

hugohaggmark:
Initial POC that illustrates how we could generate Typescript files from the `plugin.json` file.

https://github.com/user-attachments/assets/f168a8f3-7812-4700-bd92-1a845b8a328f

- The code generation should be made more extendable but I left that part out for now.
- Naming and placement of the files is something we should discuss if moving forward with this solution
- This could be extended to enforce rules in the `plugin.json`, for example For data source plugins, if the plugin supports alerting. Requires backend to be set to true. Those types of rules could be added here
- Need to add the code generation to the webpack build pipeline too
- Do I dare to mention this could be made for backend go files as well if any of the props in `plugin.json` have immediate impact on the backend that is

Fixes #https://github.com/grafana/grafana/issues/100898


packages/create-plugin/templates/common/.config/types/pluginSchema.d.ts
leventebalogh, line 1: I was wandering if we could generate this on the fly from the [plugin.json schema file](https://github.com/grafana/grafana/blob/main/docs/sources/developers/plugins/plugin.schema.json) (so we have a single source of truth)?

For example using [quicktype](https://github.com/glideapps/quicktype) in the beginning of the build process (just an idea, maybe there's a better tool for the job)? 
```
quicktype plugin.schema.json -o types.ts
```

leventebalogh (commented):
Nice one @hugohaggmark, this looks like a good direction to me! 🚀 

I guess one of the main benefits of an auto-generated type file for these metadata would be the following: _(over just importing the plugin.json, which is what plugins are usually doing now)_
- easier to discover available properties without looking at the JSON
- better type safety during development (now the code actually "knows" what includes are available) 

Obviously not part of the POC, but if we take this further we could probably somehow "generalise" the generation methods, so it's easy to introduce a new type?

packages/create-plugin/templates/common/.config/types/pluginSchema.d.ts
hugohaggmark: yup, sounds like a good idea

hugohaggmark:
One thing that I haven't added to this POC yet is that we need to run the code generation before we build too, otherwise the build could potentially break if someone was using the code generated classes/functions/enums.

tolzhabayev:
@hugohaggmark closing due to https://github.com/grafana/plugin-tools/pull/1770

packages/create-plugin/templates/common/.config/codegen/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  console.log(`${colors.green}[codegen]${colors.reset} ${colors.yellow}${message}${colors.reset}`);
+}
+
+export function logError(message: string) {
+  console.log(`${colors.green}[codegen]${colors.reset} ${colors.red}${message}${colors.reset}`);
+}
+
+export function getCodegenDirPath() {
+  return path.join(process.cwd(), SOURCE_DIR, CODEGEN_DIR);
+}
```
leventebalogh, line 50: Could we maybe reuse [the `@libs/output` lib](https://github.com/grafana/plugin-tools/tree/main/libs/output)?

PR #1575, opened 2025-02-25
docs: update links for migration handler

sympatheticmoose:
Fixes 404

Fixes #


academo (approved):

andresmgot (approved):
LGTM but I wonder if we should point to https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/feature-toggles/#development-feature-toggles instead.

sympatheticmoose:
@andresmgot interesting... that doc implies Grafana needs to be running in `dev` mode to enable them... but we don't reference that in the doc and it also is a bit surprised

andresmgot:
yeah, that's correct, [it requires running Grafana in dev mode](https://github.com/grafana/grafana/blob/main/pkg/services/featuremgmt/registry.go#L478)

sympatheticmoose:
Going to merge this just to fix the 404 and then we can continue the discussion about whether other changes should be made 👍

PR #1596, opened 2025-03-06
Update documentation to use proxyUrl instead of baseUrl

academo:
The code has been updated to use `proxyUrl` instead of `baseUrl` to better reflect its purpose as a proxy URL for the Grafana data source. The documentation should be updated to match this change to maintain consistency and clarity. The changes include updating the variable name in the code examples and related comments.


academo:
this is a test PR as part of a hackathon. ignore

PR #1615, opened 2025-03-12
Plugin E2E: Fix panel options group collapsable in 11.6.0

jackw:
This PR fixes the breaking tests that use `panelEdit.getCustomOptions('Data links')`. DataViz had the label change behind a feature flag that was removed in grafana/grafana/pull/100309.

This PR updates the tests and fixes some unused imports and stray `require()` across the e2e test suite. It also removes the `docker compose pull` from playwright gh workflow which feels redundant as the following command always pulls a new image anyhow.

I'm not entirely sure what the best solution is for these group labels. While this PR fixes it for the tests in this repo it won't solve it for plugin developers. If they don't already, maybe the e2e selectors should include labels for the standard config options? 🤔 

Fixes #

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -69,9 +69,8 @@ jobs:
           password: ${{ env.DOCKERHUB_PASSWORD }}
 
       - name: Start Grafana
+        working-directory: ./packages/plugin-e2e
         run: |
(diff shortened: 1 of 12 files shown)


oshirohugo (approved):

hugohaggmark (commented):
Great work on this 🎉

PR #1643, opened 2025-03-21
CI: Add optionDependencies to devDependencies group

jackw:
We need this to keep the devDependencies that have optionalDependencies listed in package.json aligned.

Dev dependencies in this repo are auto merged but due to the NPM bug related to optionalDependencies, we're seeing rust bindings PRs being created that should likely be bumped at the same time as their dependent NPM packages.

e.g. when `rollup` is bumped we should also bump `@rollup/rollup-linux-x64-gnu`.

Fixes #


leventebalogh (approved):
LGTM 👍

PR #1680, opened 2025-04-04
Fix: create-plugin common Webpack config types

MattIPv4:

When extending the Webpack config, the current create-plugin-provided config generates type-errors.

cc https://github.com/grafana/grafana-setupguide-app/pull/100 (private)

N/A

<details>
  <br />

  ```bash
  npm install website@3.6.3-canary.1680.14470780971.0
  npm install @grafana/create-plugin@5.19.3-canary.1680.14470780971.0
  npm install @grafana/eslint-plugin-plugins@0.3.4-canary.1680.14470780971.0
  npm install @grafana/plugin-e2e@1.19.5-canary.1680.14470780971.0
  npm install @grafana/plugin-meta-extractor@0.5.3-canary.1680.14470780971.0
  npm install @grafana/plugin-types-bundler@0.3.2-canary.1680.14470780971.0
  # or 
  yarn add website@3.6.3-canary.1680.14470780971.0
  yarn add @grafana/create-plugin@5.19.3-canary.1680.14470780971.0
  yarn add @grafana/eslint-plugin-plugins@0.3.4-canary.1680.14470780971.0
  yarn add @grafana/plugin-e2e@1.19.5-canary.1680.14470780971.0
  yarn add @grafana/plugin-meta-extractor@0.5.3-canary.1680.14470780971.0
  yarn add @grafana/plugin-types-bundler@0.3.2-canary.1680.14470780971.0
  ```
</details>


packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -34,7 +35,11 @@ __webpack_public_path__ =
 `,
 });
 
-const config = async (env): Promise<Configuration> => {
+export type Env = {
+  [key: string]: true | string | Env;
+};
```
hugohaggmark, line 40: Maybe this should be?
```ts
export type Env = Record<string, boolean | string | Env>
```

hugohaggmark (commented):
I think this looks good but I'm going to let someone with more experience with these parts approve the changes as they're more than just type changes.

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
MattIPv4: Ah yeah, they're the same (I think?), happy to change to keep it cleaner 👍

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -34,7 +35,9 @@ __webpack_public_path__ =
 `,
 });
 
-const config = async (env): Promise<Configuration> => {
+export type Env = Record<string, boolean | string | Env>
```
MattIPv4, line 38: Ah, this can't be `false`, if you do `--env test=false` you get `'false'` as a string

```suggestion
export type Env = Record<string, true | string | Env>
```

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
MattIPv4: FYI I had to revert this after checking locally, using a `Record` here causes TS to generate a circular reference error.

packages/create-plugin/templates/common/.config/webpack/utils.ts
```diff
@@ -50,14 +50,17 @@ export async function getEntries(): Promise<Record<string, string>> {
...
 
-      result[entryName] = module;
-      return result;
-    }, result);
-  }, {});
+        innerResult[entryName] = module;
+        return innerResult;
+      }, result);
+    },
+    {} as Record<string, string>
```
leventebalogh, line 64: Did this part generate a typescript error as well when you tried to import the base webpack config? I think it's better with the change, as it's inferring more, only asking to make sure I understand correctly.

leventebalogh (dismissed):
Thanks for the fix @MattIPv4, looks good to me 👍

packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
```diff
@@ -9,6 +9,7 @@ import CopyWebpackPlugin from 'copy-webpack-plugin';
 import ESLintPlugin from 'eslint-webpack-plugin';
 import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin';
 import path from 'path';
+// @ts-expect-error Plugin is not typed
```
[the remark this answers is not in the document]
MattIPv4: As soon as I do this locally, the base Webpack config blows up with a bunch of errors -- it looks like `@types/webpack-livereload-plugin` has a dependency of `@types/webpack@4`, which this ends up pulling in somehow and TS then does not agree that the types match for plugins.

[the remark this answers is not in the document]
MattIPv4: We can probably roll our own types for that as well based on what definitely-typed has:

```ts
declare module 'webpack-livereload-plugin' {
    import { ServerOptions } from "https";
    import { Compiler, Plugin, Stats, Compilation } from "webpack";
    
    interface Options extends Pick<ServerOptions, "cert" | "key" | "pfx"> {
        /**
         * protocol for livereload `<script>` src attribute value
         * @default protocol of the page, either `http` or `https`
         */
        protocol?: string | undefined;
        /**
         * The desired port for the livereload server.
         * If you define port 0, an available port will be searched for, starting from 35729.
         * @default 35729
         */
        port?: number | undefined;
        /**
         * he desired hostname for the appended `<script>` (if present) to point to
         * @default hostname of the page, like `localhost` or 10.0.2.2
         */
        hostname?: string | undefined;
        /**
         * livereload `<script>` automatically to `<head>`.
         * @default false
         */
        appendScriptTag?: boolean | undefined;
        /**
         * RegExp of files to ignore. Null value means ignore nothing.
         * It is also possible to define an array and use multiple anymatch patterns
         */
        ignore?: RegExp | RegExp[] | null | undefined;
        /**
         * amount of milliseconds by which to delay the live reload (in case build takes longer)
         * @default 0
         */
        delay?: number | undefined;
        /**
         * create hash for each file source and only notify livereload if hash has changed
         * @default false
         */
        useSourceHash?: boolean | undefined;
    }

    class LiveReloadPlugin extends Plugin {
        readonly isRunning: boolean;
        constructor(options?: Options);
    
        apply(compiler: Compiler): void;
    
        start(watching: any, cb: () => void): void;
        done(stats: Stats): void;
        failed(): void;
        autoloadJs(): string;
        scriptTag(source: string): string;
        applyCompilation(compilation: Compilation): void;
    }
    
    export = LiveReloadPlugin;
}
```

Doing that locally seems to then get all the types synced again with `@types/webpack-livereload-plugin` uninstalled.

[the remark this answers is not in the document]
MattIPv4: I'm testing in https://github.com/grafana/grafana-setupguide-app:

```
pnpm why @types/webpack

devDependencies:
@types/webpack-livereload-plugin 2.3.6
└── @types/webpack 4.41.40
```

```
pnpm typecheck

> grafana-setupguide-app@1.6.1 typecheck /Users/mattcowley/GitHub/grafana/grafana-setupguide-app
> tsc --noEmit

.config/webpack/webpack.config.ts:194:7 - error TS2322: Type 'VirtualModulesPlugin' is not assignable to type 'false | "" | 0 | WebpackPluginInstance | ((this: Compiler, compiler: Compiler) => void) | null | undefined'.
  Type 'VirtualModulesPlugin' is not assignable to type 'WebpackPluginInstance'.
    Types of property 'apply' are incompatible.
      Type '(compiler: import("/Users/mattcowley/GitHub/grafana/grafana-setupguide-app/node_modules/.pnpm/@types+webpack@4.41.40/node_modules/@types/webpack/index").Compiler) => void' is not assignable to type '(compiler: import("/Users/mattcowley/GitHub/grafana/grafana-setupguide-app/node_modules/.pnpm/webpack@5.98.0_@swc+core@1.11.11_@swc+helpers@0.5.15__webpack-cli@5.1.4/node_modules/webpack/types").Compiler) => void'.
        Types of parameters 'compiler' and 'compiler' are incompatible.
          Type 'Compiler' is missing the following properties from type 'Compiler': _pluginCompat, _plugins, plugin, apply, and 17 more.

194       virtualPublicPath,
          ~~~~~~~~~~~~~~~~~

.config/webpack/webpack.config.ts:243:7 - error TS2322: Type 'ESLintWebpackPlugin | ForkTsCheckerWebpackPlugin | LiveReloadPlugin' is not assignable to type 'false | "" | 0 | WebpackPluginInstance | ((this: Compiler, compiler: Compiler) => void) | null | undefined'.
  Type 'LiveReloadPlugin' is not assignable to type 'false | "" | 0 | WebpackPluginInstance | ((this: Compiler, compiler: Compiler) => void) | null | undefined'.
    Type 'LiveReloadPlugin' is not assignable to type 'WebpackPluginInstance'.
      Types of property 'apply' are incompatible.
        Type '(compiler: import("/Users/mattcowley/GitHub/grafana/grafana-setupguide-app/node_modules/.pnpm/@types+webpack@4.41.40/node_modules/@types/webpack/index").Compiler) => void' is not assignable to type '(compiler: import("/Users/mattcowley/GitHub/grafana/grafana-setupguide-app/node_modules/.pnpm/webpack@5.98.0_@swc+core@1.11.11_@swc+helpers@0.5.15__webpack-cli@5.1.4/node_modules/webpack/types").Compiler) => void'.
          Types of parameters 'compiler' and 'compiler' are incompatible.
            Type 'Compiler' is missing the following properties from type 'Compiler': _pluginCompat, _plugins, plugin, apply, and 17 more.

243       ...(env.development
          ~~~~~~~~~~~~~~~~~~~
244         ? [
    ~~~~~~~~~~~
... 
257           ]
    ~~~~~~~~~~~
258         : []),
    ~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: .config/webpack/webpack.config.ts:194
```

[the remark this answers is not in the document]
jackw: Took it for a spin with PNPM and looks like there's some hoisting madness that doesn't show up in my plugin that uses NPM. Either way the livereload types package aren't compatible with Webpack5. As you've noted it's using webpack 4 types and we're getting conflicts.

I've opened a PR in DefinitelyTyped to try to resolve this. I guess we roll our own for the time being to get past this.

As an aside I would suggest taking a closer look at your plugins setup - globbing the entire codebase in [tsconfig](https://github.com/grafana/grafana-setupguide-app/blob/main/tsconfig.json#L20) is probably not a good idea from either a performance or maintenance perspective. Additionally the [.npmrc file](https://github.com/grafana/grafana-setupguide-app/blob/main/.npmrc) in your plugin has been edited and the pnpm settings create-plugin scaffolds with have been removed. 

When scaffolded it looks like:

```
# This file is required for PNPM

# PNPM 8 changed the default resolution mode to "lowest-direct" which is not how we expect resolutions to work
resolution-mode="highest"

# Make sure the default patterns are still included (https://pnpm.io/npmrc#public-hoist-pattern)
public-hoist-pattern[]="*eslint*"
public-hoist-pattern[]="*prettier*"

# Hoist all types packages to the root for better TS support
public-hoist-pattern[]="@types/*"
public-hoist-pattern[]="*terser-webpack-plugin*"
```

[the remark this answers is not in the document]
MattIPv4: > I guess we roll our own for the time being to get past this.

Will do 👍 

> globbing the entire codebase in [tsconfig](https://github.com/grafana/grafana-setupguide-app/blob/main/tsconfig.json#L20) is probably not a good idea from either a performance or maintenance perspective.

Yeh, I've played around with that a bit -- landed on the globs as we have a bunch of TS/JS files outside of src, including dotfiles, that we also want TS to check fully.



> Additionally the [.npmrc file](https://github.com/grafana/grafana-setupguide-app/blob/main/.npmrc) in your plugin has been edited and the pnpm settings create-plugin scaffolds with have been removed.

It never had those as it was a Yarn plugin originally that we swapped over to pnpm recently due to Yarn 1 having some nasty dependency resolution bugs. I'll get it updated with the recommended bits from create-plugin ❤️

packages/create-plugin/templates/common/.config/types/webpack-rules.d.ts
```diff
@@ -25,8 +25,8 @@ declare module '*.webp' {
 }
 
 declare module '*.svg' {
-  const content: string;
-  export default content;
+  const src: string;
```
jackw, line 28: I'd rather not rename this file as I don't think these declarations are specific to Webpack. Any project that uses a bundler that handles importing of images / fonts would benefit from them.

jackw (commented):
Thanks for hanging in there with this one @MattIPv4. This all looks good although please see my comment about the types file rename.

packages/create-plugin/templates/common/.config/types/webpack-rules.d.ts
MattIPv4: Would you be amenable to `bundler-rules.d.ts` or similar? I found `custom.d.ts` to be too generic when there are now also types for actual modules in here as well?

jackw: I guess that's a better name for it, yeah.

jackw (approved):
LGTM! 🚀

PR #1718, opened 2025-04-11
Chore: no need to create issues for stale branches

tolzhabayev:
Seems like this action opens a bunch of issues for every stale branch, this PR should change it - we can just cleanup the branches without creating new issues.


jackw (approved):
🚀

PR #1728, opened 2025-04-15
Chore: fixes e2e tests waiting until "networkidle"

hugohaggmark:
The playwright team enabled the `PlzDedicatedWorker` feature in chromium [here](https://github.com/microsoft/playwright/pull/34400). This caused all e2e tests on scaffolded plugins to fail because the tests where waiting for the `networkidle` to show up but it never did. The errors looked something like this:
```log
- navigating to "http://localhost:3000/XYZ", waiting until "networkidle"
```

Removing `networkidle` from GrafanaPage [here](https://github.com/grafana/plugin-tools/blob/main/packages/plugin-e2e/src/models/pages/GrafanaPage.ts#L22) made the e2e tests on scaffolded plugins to pass but would break all e2e tests on the plugin-e2e package itself. I think we need to move away from the `waitUntil: 'networkidle'` solution at some point in the future so I created a separate issue for this [here](https://github.com/grafana/plugin-tools/issues/1734) 

In the meantime we're passing `--disable-features=PlzDedicatedWorker` as an argument to chromium at launch.

Fixes #https://github.com/grafana/plugin-tools/issues/1733

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.19.4-canary.1728.14489973977.0
  npm install @grafana/plugin-e2e@1.19.6-canary.1728.14489973977.0
  # or 
  yarn add @grafana/create-plugin@5.19.4-canary.1728.14489973977.0
  yarn add @grafana/plugin-e2e@1.19.6-canary.1728.14489973977.0
  ```
</details>


packages/create-plugin/templates/common/_package.json
```diff
@@ -23,7 +23,7 @@
     "@grafana/eslint-config": "^8.0.0",{{#if usePlaywright}}
     "@grafana/plugin-e2e": "^1.17.1",{{/if}}
     "@grafana/tsconfig": "^2.0.0",{{#if usePlaywright}}
-    "@playwright/test": "1.50.1",{{/if}}
+    "@playwright/test": "1.51.1",{{/if}}
```
jackw, line 26: I think we can put the caret back here now. I pinned it cos of this issue I couldn't figure out.

```suggestion
    "@playwright/test": "^1.51.1",{{/if}}
```

packages/create-plugin/templates/common/playwright.config
```diff
@@ -43,7 +43,13 @@ export default defineConfig<PluginOptions>({
     // 2. Run tests in Google Chrome. Every test will start authenticated as admin user.
     {
       name: 'chromium',
-      use: { ...devices['Desktop Chrome'], storageState: 'playwright/.auth/admin.json' },
+      use: {
+        ...devices['Desktop Chrome'],
+        storageState: 'playwright/.auth/admin.json',
+        launchOptions: {
+          args: ['--disable-features=PlzDedicatedWorker'], // because https://github.com/microsoft/playwright/pull/34400
```
jackw, line 50: Does the `auth` test above also need this?

packages/plugin-e2e/tests/as-admin-user/app/app-config/appConfig.spec.ts
```diff
@@ -16,6 +16,6 @@ test('should wait for plugin config settings API to respond', async ({ gotoAppCo
   );
 
   const response = configPage.waitForSettingsResponse();
-  page.getByRole('button', { name: 'Disable' }).first().click();
+  await page.getByRole('button', { name: 'Disable' }).first().click();
```
jackw, line 19: 👏

jackw (approved):
Really nice work figuring this out @hugohaggmark . 👏 👏 👏

packages/create-plugin/templates/common/playwright.config
hugohaggmark: Good question I'm not sure, it seems to work without it because I think that test only uses the HTTP api 👍

PR #1794, opened 2025-05-14
Docs: Add a cookbook category and a page for streaming data sources

academo:

Adds documentation for how to add support for streaming data in data sources.

It also adds a "cookbook" section in the docs for small and concise examples

Initially this cookbook has no structure but we can start creating sub-categories as we add more examples

Fixes #

### DIFF
--- a/docusaurus/docs/cookbook/_category_.json
+++ b/docusaurus/docs/cookbook/_category_.json
@@ -1,0 +1,9 @@
+{
+  "position": 70,
+  "label": "Cookbook",
+  "customProps": {
+    "description": "Start your building journey with one of our tutorials for learning the essentials of plugin development."
+  },
+  "collapsible": true,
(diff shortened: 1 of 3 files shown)


s4kh (approved):

andresmgot (commented):
I am not sure people would understand the difference between the "How to ..." guides that we have and these cookbook "recipes"? (When I say people, I also mean myself :)) What it's even more confusing is that we have a "Build a streaming data source plugin" guide, which is completely different from this one... so which one should people follow?

Maybe we should seek guidance from the docs team?

academo:
@andresmgot the difference is a cookbook has straight forward information, basically snippets, not a tutorial, not a "how to step by step"

in a similar topic I completely missed the streaming data source guide and that's what this is trying to cover so I'll close this PR :)

PR #1799, opened 2025-05-15
Docs: add more details on deps verification

tolzhabayev:
Adding documentation on how to verify version compatibility.

Preview: https://github.com/grafana/plugin-tools/pull/1799/files?short_path=d316f85#diff-d316f85003828dc8248eadb9be7c2ce76c46ece4fc5785ef424d84b98f1301de


oshirohugo (approved):

docusaurus/docs/get-started/best-practices.md
```diff
@@ -85,6 +85,27 @@ Is something missing from this list? [Let us know](https://github.com/grafana/pl
...
+:::note
+
+We strongly recommend to verify that the range you have defined in `grafanaDependency` is set correctly. The best way to verify this is using [`semver.satisfies`](https://www.npmjs.com/package/semver) function to check a specific version against the range you have defined.
+
+Example - your `plugin.json` defines `grafanaDependency: >=12.1.0`. In order to test if it would be installable on, for example, Grafana version `12.1.0-512e312` you can verify it with:
+
+```
+console.log(semver.satisfies("12.1.0-3212123", ">=12.1.0"));
+> false
+```
```
jackw, line 97: nit: The version hashes are different in the text (512e312) compared to the code (3212123).

I'm not entirely sure where I run this code? In my plugin?

Any reason not to point folk at an online semver checker like [semver-check](https://jubianchi.github.io/semver-check/#/%3E%3D12.1.0/12.1.0-3212123) over code examples?

tolzhabayev: Yeah I thought about that, unfortunately semver check you referenced is not doing proper semver checks :sigh: it ignored pre-release and not actually running `satisfies` so that it gives you a false result. In the link you mentioned it should say

```12.1.0-3212123 DOES NOT satisfy constraint >=12.1.0```

The only way to run it so far is in some code sandbox or locally, runkit seems to be dead https://npm.runkit.com/ (does not load anymore, maybe temporarily). So if you have ideas on where I can point to instead, happy to do it.

Comment (tolzhabayev) [reply docusaurus/docs/get-started/best-practices.md:97]: Hashes, fixed https://github.com/grafana/plugin-tools/pull/1799/commits/4ed925830d83154654ac4d6a1ba7b27ab84c7073

File: docusaurus/docs/get-started/best-practices.md [hunk 2]
```diff
@@ -85,6 +85,27 @@ Is something missing from this list? [Let us know](https://github.com/grafana/pl
...
+
+As you see it would not be installable. If you want to target this version you can adjust your `grafanaDependency` to `>=12.1.0-0`
+
+```
+console.log(semver.satisfies("12.1.0-3212123", ">=12.1.0-0"));
+> true
+```
+
+Read more on semver specifications on [semver.org](https://semver.org)
+:::
```
andresmgot, line 107: this is not being rendered as you expect, I think you need to add a blank line before the `:::` (check the "rich diff" view)

andresmgot (approved):
LGTM, I wonder if we need to be so verbose about this? I think that one line addressing the specific issue would be enough & more concise? Something like:

> If you want to target the next release (or the current `main`) and being able to install the plugin in Cloud, remember to add a prerelease to your `grafanaDependency`. For example `>=12.1.0-0` if `12.1.0` has not been released yet and Grafana Cloud versions look like `12.1.0-123123`.

docusaurus/docs/get-started/best-practices.md
jackw: If `12.1.0-3212123` doesn't satisfy `>=12.1.0` does that mean that no ranges are currently satisfied by canary versions of Grafana?

You can do run the commands in chrome devtools with dynamic import and an esm cdn...

```js
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-3212123", ">=12.1.0"));
// false
```

docusaurus/docs/get-started/best-practices.md
tolzhabayev: it is rendering correctly. Unfortunately rich diff != docusaurus rendering. It is close to but not the same.

tolzhabayev:
@jackw @andresmgot updated text and preview above a bit to make it shorted and highlight they can run it on the console (at least on some pages)

docusaurus/docs/get-started/best-practices.md
jackw: >Only on pages that don't have strict script-src CSP, like try running it on this page it will fail. I can try to rewrite it to ask people to run it on the same page they reading this doc and it will work.

Yeah, I ran it from my plugins docker dev env but same page can work too.

>Most of plugins we have in the cloud are targeting >=10/11 and that includes whatever the pre-release of 12 is.

I'm probably not following this clearly but if I run:

```js
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-3212123", ">=11"));
// false
```

I get back false. But that doesn't seem to align with what you're saying the current Grafana behaviour is with pre-release of 12. 😕 

It also seems the only range that satisfies a Grafana pre-release with this semver package is the same minor with -0 suffix.

```js
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-243283", ">=11.0.0-0"));
// false
```

```js
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-243283", ">=12.0.0-0"));
// false
```

```js
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-243283", ">=12.1.0-0"));
// true
```

tolzhabayev: Right, forgot to add the `includePrerelease` option, updated https://github.com/grafana/plugin-tools/pull/1799/commits/c00945bdaba7f56094a1b07e054e551c455dd72d

tolzhabayev:
```
const { satisfies } = await import("https://esm.sh/semver");
console.log(satisfies("12.1.0-3212123", ">=11", { includePrerelease: true }));
// true
```

docusaurus/docs/get-started/best-practices.md
```diff
@@ -85,6 +85,24 @@ Is something missing from this list? [Let us know](https://github.com/grafana/pl
 
 In the Grafana plugin ecosystem, plugin compatibility with specific Grafana versions is determined by the semantic versioning range defined in the `grafanaDependency` property of the plugin's `plugin.json` file. Plugin authors must carefully select a version range that balances broad compatibility with the need for manageable maintenance efforts.
 
+:::note
+
+We strongly recommend to verify that the range you have defined in `grafanaDependency` is set correctly. The best way to verify this is using [`semver.satisfies`](https://www.npmjs.com/package/semver) function with `includePrerelease` option set to `true` to check a specific version against the range you have defined.
+
+If you want to target the next release (or the current `main`) and being able to install the plugin in Grafana Cloud, remember to add a prerelease to your `grafanaDependency`. For example set it to `>=12.1.0-0` if `12.1.0` is what you want to target since Grafana Cloud versions look like `12.1.0-123123`
```
andresmgot, line 92:
```suggestion
If you want to target the next release (or the current `main`) and being able to install the plugin in Grafana Cloud, remember to add a prerelease to your `grafanaDependency`. For example set it to `>=12.1.0-0` if `12.1.0` is what you want to target since Grafana Cloud versions look like `12.1.0-123123`.
```

andresmgot (approved):
👍 LGTM

xnyo (approved):

jackw (approved):
LGTM! 🚀

PR #1823, opened 2025-05-21
Chore: Fix zizmor issues in template workflows

academo:

Updates our github workflow templates to pass [zizmor](https://github.com/zizmorcore/zizmor) checks

### DIFF
--- a/packages/create-plugin/templates/github/workflows/ci.yml
+++ b/packages/create-plugin/templates/github/workflows/ci.yml
@@ -11,15 +11,12 @@ on:
       - master
       - main
 
-permissions:
-  contents: write
-  id-token: write
-  pull-requests: write
-
 jobs:
   build:
     name: Build, lint and unit tests
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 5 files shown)


packages/create-plugin/templates/github/workflows/ci.yml
```diff
@@ -68,14 +67,14 @@ jobs:
 
       - name: Test backend
         if: steps.check-for-backend.outputs.has-backend == 'true'
-        uses: magefile/mage-action@v3
+        uses: magefile/mage-action@6f50bbb8ea47d56e62dee92392788acbc8192d0b # v3.1.0
```
andresmgot, line 70: are these doomed to be outdated? (I guess no automatic process will update these templates)

andresmgot (approved):
LGTM

packages/create-plugin/templates/github/workflows/ci.yml
academo: If the developer set up dependabot they'll be updated.

PR #1824, opened 2025-05-21
CI: Trigger examples update when a create-plugin release happens

academo:

Trigger an update of the examples whenever a new release of create-plugin happens

Fixes #

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -448,8 +448,6 @@ jobs:
 
       - name: Trigger examples update
         run: |
-          # give it a bit of time for the release to settle
-          sleep 5
           curl -X POST \
             -H "Accept: application/vnd.github.v3+json" \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -445,3 +445,14 @@ jobs:
           # comes from the get-secrets step (shared get-vault-secrets workflow)
           SLACK_WEBHOOK_URL: ${{ env.SLACK_WEBHOOK_URL }}
         run: npm run release
+
+      - name: Trigger examples update
+        run: |
```
jackw, line 450: Is it much of an issue if this triggers examples repo CI on every merge to main?

academo: Good point. I should add a check to only if files inside create-plugin package were touched

academo:
I realize the examples are working perfectly fine with the current daily cronjob and triggering here is rather unnecessary.

PR #1825, opened 2025-05-21
swc-loader: allow node_modules `.ts` and `.tsx` files

darrenjaneczek:

It can be convenient to use github imports between private repositories in order to share typescript files.
The previous rule excluded all `node_modules` files from the swc-loader. This adjustment to the `excludes` pattern allows for `ts` and `tsx` files.

Without this change (or an override to this rule's `exclude` field), webpack would emit this failing error:
> You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file.

Here is a workaround that I have been using on webpack.config which overrides that of the `.config` folder:

```ts
...
function isSwcLoaderRule(rule: any): rule is RuleSetRule {
  return rule?.use?.loader === 'swc-loader';
}
...


  // BEGIN FIX swc-loader to work with our internal git dependencies
  // Prevents error: "You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file."
  const swcLoaderRule = baseConfig.module?.rules?.filter(isSwcLoaderRule)[0];
  if (!swcLoaderRule) {
    throw new Error('Could not find the swc-loader rule. Has the .config/webpack.config.ts changed?');
  }
  // Exclude node_modules, except for typescript files (.ts, .tsx) possibly shared by git dependencies
  swcLoaderRule.exclude = /(node_modules)(?:(?!.*tsx?$))/;
  // END FIX
```

Having this `exclude` pattern within the boilerplate `.config` would be preferable to pasting this workaround in multiple projects webpack overrides.

### DIFF
--- a/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
+++ b/packages/create-plugin/templates/common/.config/webpack/webpack.config.ts
@@ -118,7 +118,8 @@ const config = async (env: Env): Promise<Configuration> => {
           ],
         },
         {
-          exclude: /(node_modules)/,
+          // Exclude node_modules, except for typescript files (.ts, .tsx) possibly shared by git dependencies
+          exclude: /(node_modules)(?:(?!.*tsx?$))/,
           test: /\.[tj]sx?$/,
           use: {
             loader: 'swc-loader',


jackw (changes requested):
This feels very specific to a single use case to me. Gonna discuss it further in our [slack channel](https://raintank-corp.slack.com/archives/C01C4K8DETW/p1748430316051119?thread_ts=1747175797.104609&cid=C01C4K8DETW) first.

PR #1842, opened 2025-05-27
Chore: Move all remaining secrets to vault

academo:
Updates existing workflows to use vault instead of Github secrets.

TODO: Once this is merge, we must remove all secrets from github

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,19 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
     name: Run unit tests
+    permissions:
+      contents: read
+      id-token: write
     env:
-      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
       NX_BRANCH: ${{ github.event.number || github.ref_name }}
     steps:
+      - id: get-secrets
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -138,13 +157,7 @@ jobs:
 
       - name: '@grafana/sign-plugin - use GRAFANA_ACCESS_POLICY_TOKEN to sign generate-panel plugin'
         env:
-          GRAFANA_ACCESS_POLICY_TOKEN: ${{ secrets.GRAFANA_ACCESS_POLICY_TOKEN }}
-        run: sign-plugin --rootUrls http://www.example.com --signatureType private
-        working-directory: ./${{ env.WORKING_DIR }}
-
-      - name: '@grafana/sign-plugin - use GRAFANA_API_KEY to sign generate-panel plugin'
```
academo, line 145: `GRAFANA_API_KEY` can't be generated anymore. I created https://github.com/grafana/plugin-tools/issues/1844 so we can remove this option from sign-plugin (and documentation).

.github/workflows/ci.yml
```diff
@@ -178,6 +192,20 @@ jobs:
             cmdArgs: --plugin-name='no-backend-scenes' --org-name='myorg' --plugin-type='scenesapp' --no-backend
             hasBackend: false
     steps:
+      - id: get-secrets
```
academo, line 195: not 100% sure if we can combine these two so I left it in two steps.

tolzhabayev: I think it's possible to combine (see [example](https://github.com/grafana/shared-workflows/tree/main/actions/get-vault-secrets#using-environment-variables-default)) but fine to leave them separated for now as long as they work - the steps just take longer.

tolzhabayev (approved):

PR #1865, opened 2025-06-04
fix: mention required config setting in service account how-to

sd2k:
Grafana requires the `auth.managed_service_accounts_enabled` configuration variable to be set to true in order to use service accounts (see [the code]). This commit updates the doc to mention this requirement.

I'm not sure why this didn't come up before though, perhaps it's a new requirement? It seems to have been the same for 9 months :thinking: 

[the code]: https://github.com/grafana/grafana/blob/7f63b18e826bc1aba2ba215ee329b989a63fc018/pkg/services/pluginsintegration/serviceregistration/serviceregistration.go#L25


mckn (commented):
LGTM! Leave the final approval for @wbrowne or @andresmgot

wbrowne (approved):
Makes sense! LGTM

PR #1875, opened 2025-06-11
Chore: Disabling tests for the meta extractor

mckn:
It will disable the tests for the meta extractor since they are failing when running against the latest version of Grafana. The meta extractor is a experimental package that isn't actively being developed and hence we don't want to run its tests in CI etc.

Fixes #


jackw (approved):
LGTM! 🚀

PR #1890, opened 2025-06-13
fix: don't exit Webpack on ESLint error in development

MattIPv4:

https://github.com/webpack-contrib/eslint-webpack-plugin/pull/275 fixed the default behaviour to now fail on error, which is quite annoying in development.

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@5.22.5-canary.1890.15645459200.0
  # or 
  yarn add @grafana/create-plugin@5.22.5-canary.1890.15645459200.0
  ```
</details>


MattIPv4:
Sure, there are essentially two reasons behind this:

- Ensuring that the config has the same behaviour whether you're using eslint-webpack-plugin v4 or v5 (in v4 this defaulted to false, in v5 it defaults to true)
- When you're working in development with webpack in watch mode and an IDE that auto-saves, if you auto-save invalid syntax with failOnError enabled (if the IDE auto-saves as you're still typing a change for example), the entire webpack process exits which is very annoying

jackw (approved):
LGTM! 🚀

PR #1937, opened 2025-06-30
fix: include 'tools' in followup request in LLM docs

sd2k:
... and use a loop to handle multiple tool call results from the LLM.


jewbetcha (approved):
👊

andresmgot (approved):

PR #1942, opened 2025-07-02
Create Plugin: Add package.json utils for migrations

jackw:
This PR that adds some util functions for interacting with package.json dependencies. Cursor wrote most of this.

I think we need to start breaking apart the `src/migrations/utils.ts` file into smaller modules soon and likely audit the `src/utils` files to see what can be cleaned up as there's some crossover occurring in some of what's being added in this PR.

Fixes #

<details>
  <br />

  ```bash
  npm install website@3.6.10-canary.1942.16599341801.0
  npm install @libs/version@1.0.2-canary.1942.16599341801.0
  npm install @grafana/create-plugin@5.25.8-canary.1942.16599341801.0
  npm install @grafana/sign-plugin@3.1.4-canary.1942.16599341801.0
  # or 
  yarn add website@3.6.10-canary.1942.16599341801.0
  yarn add @libs/version@1.0.2-canary.1942.16599341801.0
  yarn add @grafana/create-plugin@5.25.8-canary.1942.16599341801.0
  yarn add @grafana/sign-plugin@3.1.4-canary.1942.16599341801.0
  ```
</details>

### DIFF
--- a/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
+++ b/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
@@ -2,7 +2,7 @@ import type { Context } from '../context.js';
 import { addDependenciesToPackageJson, removeDependenciesFromPackageJson } from '../utils.js';
 
 export default function migrate(context: Context) {
-  if (context.doesFileExist('.config/.eslintrc')) {
+  if (context.doesFileExist('.config/.eslintrc') && context.doesFileExist('package.json')) {
     const eslintConfigRaw = context.getFile('.config/.eslintrc') || '';
     const [eslintComments, eslintConfigJSON] = splitEslintConfig(eslintConfigRaw);
 
--- a/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
+++ b/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
@@ -1,4 +1,5 @@
 import type { Context } from '../context.js';
+import { addDependenciesToPackageJson, removeDependenciesFromPackageJson } from '../utils.js';
 
 export default function migrate(context: Context) {
   if (context.doesFileExist('.config/.eslintrc')) {
@@ -34,16 +35,11 @@ export default function migrate(context: Context) {
     }
 
     if (context.doesFileExist('package.json')) {
-      const packageJson = JSON.parse(context.getFile('package.json') || '{}');
-
-      if (packageJson.devDependencies) {
-        delete packageJson.devDependencies['eslint-plugin-deprecation'];
-
-        packageJson.devDependencies['@typescript-eslint/eslint-plugin'] = '^8.3.0';
-        packageJson.devDependencies['@typescript-eslint/parser'] = '^8.3.0';
-      }
-
-      context.updateFile('package.json', JSON.stringify(packageJson, null, 2));
+      addDependenciesToPackageJson(context, {
+        '@typescript-eslint/eslint-plugin': '^8.3.0',
+        '@typescript-eslint/parser': '^8.3.0',
+      });
+      removeDependenciesFromPackageJson(context, [], ['eslint-plugin-deprecation']);
     }
   }
 
--- a/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
+++ b/packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
@@ -32,9 +32,7 @@ export default function migrate(context: Context) {
 
       const result = eslintComments.join('\n') + '\n' + JSON.stringify(eslintConfig, null, 2);
       context.updateFile('.config/.eslintrc', result);
-    }
 
-    if (context.doesFileExist('package.json')) {
       addDependenciesToPackageJson(context, {
         '@typescript-eslint/eslint-plugin': '^8.3.0',
         '@typescript-eslint/parser': '^8.3.0',
--- a/packages/create-plugin/src/migrations/utils.ts
+++ b/packages/create-plugin/src/migrations/utils.ts
@@ -273,7 +273,7 @@ export function isVersionGreater(incomingVersion: string, existingVersion: strin
   }
 
   // We can't determine the exact version the dist tag pointed to so we force an update so the migration changes and expected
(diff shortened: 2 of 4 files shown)


packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,174 @@ export function installNPMDependencies(context: Context) {
...
+    throw new Error(`Cannot parse ${path}: ${e}`);
+  }
+}
+
+export function addDependenciesToPackageJson(
+  context: Context,
+  dependencies: Record<string, string>,
+  devDependencies: Record<string, string> = {},
+  packageJsonPath = 'package.json'
+) {
```
hugohaggmark, line 153: What are the main arguments for using homegrown solution that we need to maintain instead of using `npm`|`yarn`|`pnpm` or 3rd party lib?

hugohaggmark (commented):
Doing a first pass and this looks great 🎉

packages/create-plugin/src/migrations/utils.ts
hugohaggmark: Thanks for the great explain, all valid and solid points 🙌

packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,174 @@ export function installNPMDependencies(context: Context) {
     execSync(installCmd, { cwd: context.basePath, stdio: 'inherit' });
   }
 }
+
+export function readJsonFile<T extends object = any>(context: Context, path: string): T {
+  if (!context.doesFileExist(path)) {
+    throw new Error(`Cannot find ${path}`);
+  }
+  try {
+    return JSON.parse(context.getFile(path) || '{}');
```
leventebalogh, line 142: Why do we need the `|| '{}'? If the file is empty, then we would like to throw the error (not able to parse) - or am I wrong?

packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,174 @@ export function installNPMDependencies(context: Context) {
...
+    } else if (currentDevDeps[dep]) {
+      // Exists in devDependencies, only update there if new version is greater
+      if (isIncomingVersionGreater(newVersion, currentDevDeps[dep])) {
+        currentDevDeps[dep] = newVersion;
+      }
+    } else {
+      // Not present, add to dependencies
+      currentDeps[dep] = newVersion;
+    }
+  }
```
leventebalogh, line 173: Nit: do we want to log something if the migration was trying to downgrade dependencies?

packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,174 @@ export function installNPMDependencies(context: Context) {
...
+
+  // Sort dependencies alphabetically for consistency
+  const sortedDeps = sortObjectByKeys(currentDeps);
+  const sortedDevDeps = sortObjectByKeys(currentDevDeps);
+
+  // Only update if there are actual changes
+  const hasChanges =
+    JSON.stringify(sortedDeps) !== JSON.stringify(currentPackageJson.dependencies || {}) ||
+    JSON.stringify(sortedDevDeps) !== JSON.stringify(currentPackageJson.devDependencies || {});
```
leventebalogh, line 199: Just to make sure I understand correctly - this also updates the package.json if there were no changes in the dependencies but the sorting order has been changed - is that something that we want?

packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,174 @@ export function installNPMDependencies(context: Context) {
...
+  if (incomingVersionCompareBy in DIST_TAGS && existingVersionCompareBy in DIST_TAGS) {
+    return (
+      DIST_TAGS[incomingVersionCompareBy as keyof typeof DIST_TAGS] >
+      DIST_TAGS[existingVersionCompareBy as keyof typeof DIST_TAGS]
+    );
+  }
+
+  if (incomingVersionCompareBy in DIST_TAGS || existingVersionCompareBy in DIST_TAGS) {
+    return true;
+  }
```
leventebalogh, line 284: Is this correct? 🤔 What if the latest version is `12.0.0`, the incoming version is `10.0.0`, while the existing version - which is also in dist tags - is `latest`?

leventebalogh (commented):
Nice changes! Left a few questions.

packages/create-plugin/src/migrations/scripts/003-update-eslint-deprecation-rule.ts
```diff
@@ -1,4 +1,5 @@
 import type { Context } from '../context.js';
+import { addDependenciesToPackageJson, removeDependenciesFromPackageJson } from '../utils.js';
```
tolzhabayev, line 2: Can you ask cursor if migration scripts should ever be changed or rather new scripts are usually added to ensure same end state after migrations?

packages/create-plugin/src/migrations/utils.ts
jackw: I've added some debug logging to the function. I don't think we want to output this for every migration, it could get noisy quickly.

packages/create-plugin/src/migrations/utils.ts
```diff
@@ -132,3 +133,176 @@ export function installNPMDependencies(context: Context) {
...
+export function isVersionGreater(incomingVersion: string, existingVersion: string): boolean {
+  const incomingIsDistTag = incomingVersion in DIST_TAGS;
+  const existingIsDistTag = existingVersion in DIST_TAGS;
+
+  if (incomingIsDistTag && existingIsDistTag) {
+    return DIST_TAGS[incomingVersion as keyof typeof DIST_TAGS] > DIST_TAGS[existingVersion as keyof typeof DIST_TAGS];
+  }
+
+  // We can't determine the exact version the dist tag pointed to so we force an update so the migration changes and expected
+  // dependency version are aligned. This should mean the codebase is more likely to continue working post-migration, even if
```
leventebalogh, line 276: Nit
```suggestion
  // dependency versions are aligned. This should mean the codebase is more likely to continue working post-migration, even if
```

packages/create-plugin/src/migrations/utils.ts
leventebalogh: Yeah, I think it's good 👍

leventebalogh (approved):
LGTM 🚀

PR #1983, opened 2025-07-22
RBAC: Updates the docs on `actionSets` with limitations on folders and dashboard actions only

eleijonmarck:
moved the actionsets section to the top to make it clearer that we only support extending the actionsets on folders


urbiz-grafana (approved):
LGMT, left minor suggestion.

tolzhabayev (approved):

PR #1988, opened 2025-07-24
fix(deps): bump rxjs to 7.8.2 to match @grafana/ui version

MattIPv4:

I'd quite like #1903 to land as I'd like to make use of some bits landed in 12.1.0. Tests were failing on that though due to the `rxjs` version in `@grafana/ui` being bumped.

With this change locally, `cd packages/create-plugin && npm ci && npm run build && ./dist/bin/run.js --plugin-name='backend' --org-name='myorg' --plugin-type='app' --backend && cd myorg-backend-app && npm install && npm run typecheck` passes, where before it would not due to an `rxjs` type-error.


jackw (approved):
Thanks @MattIPv4 , I was about to fix this myself as it's breaking CI in this repo too. 🙏

PR #1997, opened 2025-07-30
Docs: Minor updates, rewording for UI extensions 

urbiz-grafana:
Covering https://github.com/grafana/grafana-community-team/issues/313 et al.


docusaurus/docs/how-to-guides/ui-extensions/expose-a-component.md
```diff
@@ -11,7 +11,9 @@ keywords:
 sidebar_position: 30
 ---
 
-Expose components to allow app plugins to easily share functionality with other app plugins. Compared to [registering an extension](./register-an-extension), they do not require the extension provider to explicitly register a component against any extension points, and can therefore be [used by any app plugin](./use-an-exposed-component.md) with no action required by the provider.
+Expose components to allow app plugins to easily share functionality with other app plugins. 
+
+Compared to [registering an extension](./register-an-extension), when you expose a component you do not require the other extension providers to explicitly register their extensions against any extension points. Therefore the component may be [used by any app plugin](./use-an-exposed-component.md) with no action required by the provider.
```
leventebalogh, line 16: I think it's better! 

---

As we generally had an issue expressing what we mean by this, I'll try to add a few more thoughts to it, they might help.

We have **"extension points"**, either in core Grafana or in plugins, which are basically parts of the UI which can be extended by plugins. They extend these parts by _registering functionality_ (links or components) using the identifier of the extension point, and when the extension point is rendered it can fetch these registered extensions, and display them in a certain way. (How it displays them is up to the given extension point).

Screenshot 2025-08-01 at 6 54 15

In this case we called the plugins "providers", because they are ones that provide the extensions to the extension point, which is a "consumer". Due to the nature of this, the "consumer" cannot easily render the same extension(s) somewhere else as well (technically it could, but it would not be straightforward), it will first need to create a new extension point, and then ask the plugins ("providers") to register their extensions to this new extension point, too. _**This is what we mean above by "action required by the provider".**_ 

We also introduced a concept called **"exposed components"**, which is a little bit different. A plugin doesn't register an extension (link or component) to a certain extension point in the UI, but they are exposing a component as is (it has a unique id), and other plugins can grab it and render it wherever they want. If they choose to render it in other places as well, then they just reference it there as well, the "provider" (plugin that exposes it) doesn't need to get involved.
Screenshot 2025-08-01 at 7 03 27

docusaurus/docs/how-to-guides/ui-extensions/register-an-extension.md
```diff
@@ -13,23 +13,23 @@ sidebar_position: 20
 
 import ExtensionPoints from '@shared/extension-points.md';
 
-An _extension_ is a plugin-defined link or a React component that is rendered in either the core Grafana UI or in another app plugin.
+A UI extension is either a plugin-defined link or a React component that you can render in the core Grafana UI or in another app plugin.
```
leventebalogh, line 16: There's even another twist to this now: an extension can also be a generic function (which doesn't necessarily render anything, just returns a value based on a list of arguments). Due to this we started to just call them "extensions", as we thought it's less misleading. (I know, right now everything is misleading in here :))

docusaurus/docs/how-to-guides/ui-extensions/register-an-extension.md
```diff
@@ -138,15 +138,17 @@ export const plugin = new AppPlugin().addLink({
...
 ### Best practices for adding components
 
 - **Use the props** - check what props the extension point is passing to the components and use them to implement a more tailored experience.
 - **Wrap your component with providers** - if you want to access any plugin specific state in your component make sure to wrap it with the necessary React context providers (e.g. for Redux).
 - **Use the enum for Grafana extension point IDs** - if you are registering a component to one of the available Grafana extension points, make sure that you use the [`PluginExtensionPoints` enum exposed by the `@grafana/data`](https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/types/pluginExtensions.ts#L121) package.
 
-### Register a component
+### Register a component extension
+
+You can register an extension for a component. For example:
```
leventebalogh, line 151:
```suggestion
You can register a component extension. For example:
```

docusaurus/docs/key-concepts/ui-extensions.md
```diff
@@ -12,34 +12,42 @@ keywords:
...
 
-1. `Extension point` - A place in the UI where plugins can contribute new functionality to the end user. The Grafana UI exposes extension points and app plugins can also define their own extension points.
-2. `Extension` - New functionality, registered by an app plugin, appearing at an extension point. For example, an extension could provide a navigational link to bring users to a particular view, it could open a modal menu allowing the user to configure an action to take from within their current context (for instance, to create a SLO), or it could trigger background tasks.
+1. `Extension point` - A place in the UI where plugins can contribute new functionality to the end user. Both the Grafana UI and app plugins can expose extension points.
+2. `Extension` - New functionality, registered by an app plugin, appearing at an extension point. For example, an extension could provide a navigational link to bring users to a particular view, it could open a modal menu allowing the user to configure an action to take from within their current context (for instance, to create a SLO), or it could trigger background tasks. There are two types of extensions: links and components.
```
leventebalogh, line 22: Now we also have `function` extensions - however I think we don't have docs for them yet.

docusaurus/docs/key-concepts/ui-extensions.md
```diff
@@ -12,34 +12,42 @@ keywords:
...
+
+After you have added an extension point to your UI you can extend it multiple times by multiple plugins.
+
+## Where can I find extensions?
```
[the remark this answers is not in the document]
leventebalogh: Yeah, so you can only see extensions at these extensions points if there are any plugins that are extending them. An on-prem installation can have extensions as well, but the "where" depends on what plugins they have installed. It means that it's not guaranteed that they will see any extensions in a panel menu for example.

I would probably just go with the screenshot (or another screenshot later if we think this one is not expressive enough), and then list all the available core extension points with descriptions somewhere else in the docs?

leventebalogh (commented):
Nice improvements @urbiz-grafana! 👏  I've left a few comments, let me know what you think.

docusaurus/docs/key-concepts/ui-extensions.md
urbiz-grafana: Ah well, time to do them then!

leventebalogh (approved):
Nice 🚀

PR #2015, opened 2025-08-06
CI: Use self hosted runners

jackw:
This PR switches us over to use our self hosted runners which should save us some 💵 . 

Fixes #


toddtreece (approved):

xnyo (approved):
:rocket:

PR #2105, opened 2025-09-08
Chore: enable grafana dev image in Playwright workflow

hugohaggmark:
> [!IMPORTANT]  
> Merge after https://github.com/grafana/grafana/pull/110866 is merged

Changed skip-grafana-dev-image option to false.

Fixes #


sunker (approved):
Looks good! 

It's unfortunate that the table panel cells and headers don't have proper e2e selectors. If I recall correctly, I tried fixing this a year ago but the rendering is handled in a third-party package so it couldn't be done easily. I'll see if I can revisit this sometime soon. Will create an issue for it.

Issue #2107, opened 2025-09-08
Feat: Support Node 24


jackw:
### Which areas does this feature request relate to

- [x] Create Plugin
- [x] Sign Plugin
- [x] Plugin E2E
- [x] Plugin Meta Extractor
- [ ] Documentation

### Problem

Node 24 will go LTS on Oct 28th. We should make sure the packages in this repo support it.

### Solution

Something like this???

1. Update CI to run using Node 24.
2. Update any packagejson engines entries to include node 24.
3. Fix errors

### Additional context

https://github.com/nodejs/release#release-schedule

### Are you interested in contributing the solution?

- [ ] Yes
- [ ] No

tolzhabayev:
Were there any specific reason for <=22 @sunker https://github.com/grafana/plugin-tools/blob/dcbbdff4b947e3690654adad0715610babf5dd93/packages/plugin-e2e/package.json#L34?

sunker:
> Were there any specific reason for <=22 [@sunker](https://github.com/sunker)
> 
> [plugin-tools/packages/plugin-e2e/package.json](https://github.com/grafana/plugin-tools/blob/dcbbdff4b947e3690654adad0715610babf5dd93/packages/plugin-e2e/package.json#L34)
> 
> Line 34 in [dcbbdff](/grafana/plugin-tools/commit/dcbbdff4b947e3690654adad0715610babf5dd93)
> 
>  "node": ">=18 <=22" 
> ?

No not that I'm aware of.

PR #2235, opened 2025-10-20
Chore: move repo to Node 24

jackw:

Node 24 goes LTS next week. This PR does the necessary to shift the entire repo over to use Node 24.

Given that dropping older versions of node is considered a breaking change I'll open separate PRs for package specific changes to engines so we don't accidentally push out major versions wrongly.

Related: #2107


academo (approved):

leventebalogh (approved):
LGTM 🚀

tolzhabayev (approved):

PR #2268, opened 2025-11-05
Docs: Introduce eslint config for md,mdx files

jackw:

This PR lays the foundations for an eslint config that lints the docs files and their code blocks. It requires some fine tuning to make it really valuable but I think we can approach that iteratively rather than in one PR. It already caught misconfigured languages in code blocks which are fixed in this PR.

Fixes #


andresmgot (approved):
Sounds good to me

hugohaggmark (approved):

sunker (approved):
Really nice! There are some codeblocks in markdown files outside of `docusaurus/docs/`. Would it make sense to include those in the linting scope as well?

PR #2271, opened 2025-11-05
Remove cookie override

tolzhabayev:
Seems like it's not needed and docusaurus core has it properly defined. Previously pinned in https://github.com/grafana/plugin-tools/commit/e464910181840cfc09468ec79b44c3c887cb1fc7


jackw (approved):
LGTM 🚀

PR #2314, opened 2025-11-19
Vibe: Adding agents.md when scaffolding panel plugin

mckn:
It will add a agents.md file that cover the fundamentals of panel plugins. We are keeping the convention where we put generic information below `.config` folder (which should not be modified by the developer). Then we reference that file from the project `AGENTS.md` where the developers can add more project specific details.

Fixes https://github.com/grafana/grafana-community-team/issues/673

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@6.4.0-canary.2314.19664595751.0
  # or 
  yarn add @grafana/create-plugin@6.4.0-canary.2314.19664595751.0
  ```
</details>


hugohaggmark (approved):
Let's gooo 🎉

sunker (commented):
Nice! ✨ 

I was looking at some examples of agent files, and it seems to be a common pattern to list and describe all available commands early in the file. Even though the frontend scripts are specified in package.json and are mostly self-explanatory, it's not entirely obvious that one needs to have a server running before, for example, running the e2e tests.

packages/create-plugin/templates/panel/.config/AGENTS/tasks/add-panel-options.md
hugohaggmark: I'm thinking out loud here, but instead of tasks maybe we should supply plans instead? So the Agents role is to create a plan for a given task instead. This might be a stupid idea, let me know what you think? 

Other than that I think this looks good, might want to add a small instruction if the plugin is translated to keep the same?

packages/create-plugin/templates/panel/.config/AGENTS/tasks/add-panel-options.md
```diff
@@ -0,0 +1,130 @@
+# Adding a New Panel Option
```
svennergr: I think while we know what a new panel "option" is, the LLM might not fully be aware of that, because "option" is such a generic word. Could be worth outlining when this "task"/"plan" should be applied or used?

packages/create-plugin/templates/panel/.config/AGENTS/instructions.md
```diff
@@ -0,0 +1,83 @@
...
+- Avoid new dependencies unless necessary + Grafana-compatible
+- Maintain consistent file structure and predictable types
+- Use **`@grafana/plugin-e2e`** for E2E tests and **always use versioned selectors** to interact with the Grafana UI.
+
+## Boundaries
+
+You must **NOT**:
+
+- Change plugin ID or plugin type in `plugin.json`
+- Modify anything inside `.config/*`
```
jackw: nit: I have an AGENTS.md in my plugin with a section similar to this. Not sure if we wanna take anything from it?

## Critical Constraints

### DO NOT Edit `.config/` Directory
**NEVER directly edit files in the `.config/` directory.** These files are maintained by the `@grafana/create-plugin` npm package and will be overwritten during updates.

To customize tooling configuration:
- Import the base configuration from `.config/`
- Extend it in root-level configuration files
- For webpack customizations, use the `webpack-merge` package to merge with the base config from `.config/webpack/`

Example: To customize webpack, import from `.config/webpack/webpack.config.ts` and merge your changes using `webpack-merge`.

jackw (approved):
LGTM! 🚀

leventebalogh (approved):
Nice work 👏

packages/create-plugin/templates/panel/.config/AGENTS/instructions.md
mckn: Yes, I think this sounds great! But feels a bit too long. Lets see if I can make it a bit more compact using AI ;)

hugohaggmark (approved):
Works fine as far as I've tested 🎉

PR #2330, opened 2025-11-27
Suggestions: Update setVisualization util to target the new 12.4 ui as well as the existing ui

fastfrwrd:
In https://github.com/grafana/grafana/pull/114381, we are updating the UI for the header of the Panel Editor to condense the information a bit.

Screenshot 2025-11-25 at 4 47 01 PM

That means that the final check in this method will no longer work. This change should make it so that the check will succeed for any version of Grafana it's used with.


mckn (approved):
LGTM!

PR #2348, opened 2025-12-16
fix: stop eslint migration stripping extra chars from plugin name

jackw:
getPluginShortName is stripping `@` from the front of plugin names resulting in broken configs due to the plugin name and the rules not matching.

Fixes #

<details>
  <br />

  ```bash
  npm install @grafana/create-plugin@6.4.5-canary.2348.20275461487.0
  # or 
  yarn add @grafana/create-plugin@6.4.5-canary.2348.20275461487.0
  ```
</details>


hugohaggmark (approved):

leventebalogh (approved):

jackw:
>Q: In this situation when the migration has been around for some time, would it make sense to add a new migration instead of adjusting the existing one?

I don't think so. If this has occurred for anyone they would have fixed it themselves as eslint will error out and not run making committing this to the repo rather difficult.

sunker (approved):

PR #2394, opened 2026-01-14
Enhance 'path' property description and requirements

s4kh:

plugin.json `routes[].path` should be defined, otherwise it won't be proxied. Updated the description for the 'path' property to include a warning about empty strings and marked it as required.

Fixes https://github.com/grafana/grafana-community-team/issues/715


hugohaggmark (changes requested):
Hi @s4kh , this file is autogenerated from https://github.com/grafana/grafana/blob/main/docs/sources/developers/plugins/plugin.schema.json so you need to add the comment there first and then run `npm -w website run prebuild` in the root of this branch and it should update this file automatically

PR #2398, opened 2026-01-15
LLMs: Scaffold AGENTS.md for LLM coding agents

academo:

This is an alternative to https://github.com/grafana/plugin-tools/pull/2322/

The main differences are:

* guide the model towards fetching online links
* compile information for both frontend and backend without conditionals

# Tests

## methodology

I did separate tests with coding agents and several coding models, with a brand new scaffolded data source  (no changes only boilerplate) I asked the agent: `can you add variable query variable support to this datasource?`

## cursor agent, with codex, claude code

cursor-agent doesn't have an integrated web_fetch tool, it follows instructions to use curl and fetches the documentation before trying to implement a solution. sometimes it hallucinates a documentation url but when it finds a 404 it fetches the index and then the real doc, this doesn't happen all the time

image

## claude code with sonnet, haiku and opus

follows instructions and fetch documentation before trying to implement. sometimes it fetches the index first, sometimes just guesses the URL correctly. 

## Opencode with GLM 4.7

follows instructions, it hallucinated many documentation urls before deciding to fetch the index and get the real documentation url

## Gemini-cli with gemini 2.5
ignored AGENTS.md and had to create a GEMINI.md instead. once created it followed instructions, first fetching the index then the correct documentation.


packages/create-plugin/templates/common/CLAUDE.md
```diff
@@ -0,0 +1,2 @@
+## Project knowledge
```
academo: why are we repeating this instead of using `@AGENTS.md` ?

because users tend to like their tools as they work. Claude likes claude.md and it'll try to modify (via its memory functions) and users might decide to remove `AGENTS.md` "because CLAUDE.md already exists" or similar with other agents that don't use either CLAUDE.md or GEMINI.md

so even though this is duplicated we are create a more future-proof approach for users with diverse coding tools and file preference. 


Users  are free to delete the agents files they don't need, but if we only have the real content in `AGENTS.md` and that's the one the user deletes we lose the any way to discover our instructions.md

mckn (approved):
LGTM! I have not tested this one but does it understand what kind of plugin you currently work with and how it is constructed?

academo:
> LGTM! I have not tested this one but does it understand what kind of plugin you currently work with and how it is constructed?

Yes, the agent is quite capable of understanding the source code when it goes to explore and goes to see what it's doing.

So we don't really need to give it so much information about what is this exactly, a plugin, an app, a data source, because it can find that information by exploring the source code. It doesn't need to have it on the agents.md file. And it is more about "you need to use our documentation"  to prevent them from going and writing something that they maybe have in their training data, but it's out of date. So that is why the pivot of this PR is all about, "consult our documentation, you have out of date information"

PR #2474, opened 2026-02-20
Docs CLI: Add validation engine and `validate` command

sunker:

This PR adds a validation foundation to `@grafana/plugin-docs-cli` with a new `validate` CLI command and two basic filesystem rules (`root-index-exists` and `has-markdown-files`).

The validation architecture separates rules from the engine. Rule runners return `Finding[]` (rule ID, title, detail) without severity. The engine stamps severity from `RuleDefinition` metadata and returns a `ValidationResult` with `valid: boolean` and `diagnostics: Diagnostic[]`. The JSON output schema is designed to map directly to plugin-validator's `analysis.Diagnostic`.

More filesystem rules and other rule categories (frontmatter, markdown, security, assets, cross-file) will follow in subsequent PRs. Output is JSON only for now - human-readable formatting comes later. For the full plan see the epic.

```bash
npx @grafana/plugin-docs-cli validate
```

Part of https://github.com/grafana/grafana-catalog-team/issues/769

wdocsPath resolution (reading src/plugin.json + checking the path exists) was centralized in run.ts instead of being duplicated across serve, build and validate. Each command now receives docsPath as a parameter.

### DIFF
--- a/packages/plugin-docs-cli/src/commands/validate.command.ts
+++ b/packages/plugin-docs-cli/src/commands/validate.command.ts
@@ -1,0 +1,26 @@
+import type minimist from 'minimist';
+import createDebug from 'debug';
+import { validate } from '../validation/engine.js';
+import { allRules } from '../validation/rules/index.js';
+import { formatHuman, formatJson } from '../validation/format.js';
+
+const debug = createDebug('plugin-docs-cli:validate');
+
+export async function validateCommand(argv: minimist.ParsedArgs, docsPath: string): Promise<void> {
+  debug('Validate command invoked with args: %O', argv);
+
+  const json = argv.json || false;
+  const strict = argv.strict || false;
+
+  debug('Validating docs at: %s (strict: %s)', docsPath, strict);
+
+  const result = await validate({ docsPath }, 'release', allRules, { strict });
+
+  if (json) {
+    console.log(formatJson(result));
+  } else {
+    console.log(formatHuman(result));
+  }
+
+  process.exit(result.valid ? 0 : 1);
+}
--- a/packages/plugin-docs-cli/src/commands/validate.command.ts
+++ b/packages/plugin-docs-cli/src/commands/validate.command.ts
@@ -11,5 +11,5 @@ export async function validateCommand(docsPath: string): Promise<void> {
 
   console.log(JSON.stringify(result, null, 2));
 
-  process.exit(result.valid ? 0 : 1);
+  process.exitCode = result.valid ? 0 : 1;
 }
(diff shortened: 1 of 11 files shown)


packages/plugin-docs-cli/src/commands/validate.command.ts
```diff
@@ -0,0 +1,15 @@
...
+import { allRules } from '../validation/rules/index.js';
+
+const debug = createDebug('plugin-docs-cli:validate');
+
+export async function validateCommand(docsPath: string): Promise<void> {
+  debug('Validating docs at: %s', docsPath);
+
+  const result = await validate({ docsPath }, allRules);
+
+  console.log(JSON.stringify(result, null, 2));
```
sunker: More human readable format will come in upcoming PRs

packages/plugin-docs-cli/src/validation/rules/index.ts
```diff
@@ -0,0 +1,7 @@
+import type { RuleCategory } from '../engine.js';
+import { filesystemDefinitions, checkFilesystem } from './filesystem.js';
+
+/**
+ * All registered rule categories.
+ */
+export const allRules: RuleCategory[] = [{ definitions: filesystemDefinitions, run: checkFilesystem }];
```
academo: this system will quickly become boilerplately.

What if instead you let the reporter send the rule (just like the plugin-validator does) so your engine doesn't need to have this definitions variable around, just take it from the report itself. 

Basically put `RuleDefinition` as part of `Finding` and let the validators submit it. it is faster and more explicit than later trying to figure out which reports belongs to which definition.

Additionally you can use enums so if someone for example starts sending a `Finding` that doesn't have a corresponding `RuleDefinition` the validator files to compile instead of failing to find the definitions 

if you are worried about having different severities in strict and non-strict you can have a rule defining a `Severity` and `StrictSeverity`

sunker: This is good feedback. Have made some changed: Rule runners now return `Diagnostic[]` directly with severity, so the definitions/lookup/assembly layer is gone entirely. Engine is just a loop now.

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -0,0 +1,41 @@
...
+    const entries = await readdir(input.docsPath, { recursive: true });
+    hasMarkdown = entries.some((entry) => entry.endsWith('.md'));
+  } catch {
+    // docsPath doesn't exist or isn't readable - will be caught by has-markdown-files
+  }
+
+  if (!hasMarkdown) {
+    diagnostics.push({
+      rule: 'has-markdown-files',
+      severity: 'error',
```
academo: likewise, severities must come from an enum (or enum const)

academo (approved):

PR #2476, opened 2026-02-24
Plugin Docs: Add filesystem validation rules

sunker:
This PR adds six filesystem validation rules.

- \`no-spaces-in-names\` — spaces in file or folder names break URL slugs
- \`valid-file-naming\` — file and folder names should only use \`[a-z0-9-]\` for clean slugs
- \`nested-dir-has-index\` — subdirectories without \`index.md\` appear as unnamed categories
- \`no-empty-directories\` — directories with zero allowed files (md or images) serve no purpose
- \`no-symlinks\` — symlinks are not allowed in the docs folder
- \`allowed-file-types\` — only \`.md\` and permitted image formats (png, jpg, jpeg, webp, gif) are allowed

A strict mode (\`--strict\` flag, on by default) controls whether certain rules report as errors or warnings. Serve-time validation with \`strict: false\` is planned in a follow-up PR.

It also adds \`formatResult\` for human-readable CLI output and a \`--json\` flag for machine-readable output, and syncs the build command's copy filter with the \`allowed-file-types\` rule.

Part of https://github.com/issues/assigned?issue=grafana%7Cgrafana-catalog-team%7C769

### DIFF
--- a/packages/plugin-docs-cli/src/commands/validate.command.ts
+++ b/packages/plugin-docs-cli/src/commands/validate.command.ts
@@ -4,10 +4,13 @@ import { allRules } from '../validation/rules/index.js';
 
 const debug = createDebug('plugin-docs-cli:validate');
 
-export async function validateCommand(docsPath: string): Promise<void> {
-  debug('Validating docs at: %s', docsPath);
+export async function validateCommand(
+  docsPath: string,
+  options: { strict: boolean } = { strict: true }
+): Promise<void> {
+  debug('Validating docs at: %s (strict: %s)', docsPath, options.strict);
 
-  const result = await validate({ docsPath }, allRules);
+  const result = await validate({ docsPath, strict: options.strict }, allRules);
 
   console.log(JSON.stringify(result, null, 2));
 
--- a/packages/plugin-docs-cli/src/commands/validate.command.ts
+++ b/packages/plugin-docs-cli/src/commands/validate.command.ts
@@ -1,18 +1,23 @@
 import createDebug from 'debug';
 import { validate } from '../validation/engine.js';
+import { formatResult } from '../validation/format.js';
 import { allRules } from '../validation/rules/index.js';
 
 const debug = createDebug('plugin-docs-cli:validate');
 
 export async function validateCommand(
   docsPath: string,
-  options: { strict: boolean } = { strict: true }
+  options: { strict: boolean; json: boolean } = { strict: true, json: false }
 ): Promise<void> {
-  debug('Validating docs at: %s (strict: %s)', docsPath, options.strict);
+  debug('Validating docs at: %s (strict: %s, json: %s)', docsPath, options.strict, options.json);
 
   const result = await validate({ docsPath, strict: options.strict }, allRules);
 
-  console.log(JSON.stringify(result, null, 2));
+  if (options.json) {
+    console.log(JSON.stringify(result, null, 2));
+  } else {
+    console.log(formatResult(result));
+  }
 
   process.exitCode = result.valid ? 0 : 1;
 }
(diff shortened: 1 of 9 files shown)


packages/plugin-docs-cli/src/commands/validate.command.ts
```diff
@@ -4,10 +4,13 @@ import { allRules } from '../validation/rules/index.js';
 
 const debug = createDebug('plugin-docs-cli:validate');
 
-export async function validateCommand(docsPath: string): Promise<void> {
-  debug('Validating docs at: %s', docsPath);
+export async function validateCommand(
+  docsPath: string,
+  options: { strict: boolean } = { strict: true }
+): Promise<void> {
```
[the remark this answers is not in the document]
sunker: This will be handled in a follow-up PR. See epic

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -40,5 +84,64 @@ export async function checkFilesystem(input: ValidationInput): Promise<Diagnosti
     });
   }
 
+  for (const dir of dirs) {
+    const dirPath = join(dir.parentPath, dir.name);
+
+    // asset directories like img/ are expected to have no markdown
+    if (ASSET_DIRS.has(dir.name)) {
```
mckn: Q: do we allow asset directories ay any level in the documentation? I'm asking to see if this is intentional.

mckn (commented):
Q: regarding the description of this PR. It says something like: "the serve command will use non-strict mode so developers aren't blocked during local previews." But when viewing the code I can't see that the serve command invoke the validate parts at all, right?

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
academo: do we have a specific good reason to be opinionated on how the assets folders are called? what if I like to use `images` instead of `img`? or what if I just creates `assets/img` and so on. 

do we have a strong valid reason to force people into a folder structure like this?

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -40,5 +84,64 @@ export async function checkFilesystem(input: ValidationInput): Promise<Diagnosti
...
+      continue;
+    }
+
+    // no-empty-directories: directories with no .md files at any depth serve no purpose
```
academo: should this be, if a directory has none of the allowed extension files?

packages/plugin-docs-cli/src/validation/format.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function formatResult(result: ValidationResult): string {
+  const lines: string[] = [];
+
+  if (result.diagnostics.length === 0) {
+    lines.push('✓ Documentation is valid');
+    return lines.join('\n');
+  }
+
+  const errors = countBySeverity(result.diagnostics, 'error');
```
academo: nit: you can do this in a single loop too.

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
sunker: The reasoning behind this is that IF we ever decide to pull out the images from the archive before sending it to GCS, having them in a predefined folder from the start will make this a lot easier. It doesn't cost much from I DX perspective, so I think this is worth it.

sunker:
Thanks for feedback @academo @mckn! Think this one is ready for another round now.

academo (approved):

mckn (approved):
LGTM!

Issue #771, opened 2024-02-23
Docs: Improve docs for optional elements and reduce number of prompts


sympatheticmoose:
Extracted from https://github.com/grafana/plugin-tools/issues/750 : "I would like to know what's required at minimum for a plugin to build and execute. This would hopefully allow me to shorten the duration of an npm install and reduce the overall file size during development."

Create plugin is an opinionated tool for scaffolding plugins. We aim to provide developers with everything they need to develop a plugin which follows our best practices. However, this does include a number of optional elements, not strictly required, and there can be valid reasons why they may wish to omit or remove certain libraries or generated files. Some optional elements we expose to developers as prompts in the tool - for example the levitate compatibility check, whilst flexibility can be useful - it also complicates matters and slows down the getting started workflow. 

We have recently introduced the ability to pass flags to create-plugin to allow it to be run without using the interactive prompts. 

After generating a (frontend only) plugin we provide the following success message:
```
## What's next?

Run the following commands to get started:

    * cd ./org-name-app
    * npm install to install frontend dependencies.
    * npm run dev to build (and watch) the plugin frontend code.
    * docker-compose up to start a grafana development server. Restart this command after each time you run mage to run your new backend code.
    * Open http://localhost:3000 in your browser to create a dashboard to begin developing your plugin.

Note: We strongly recommend creating a new Git repository by running git init in ./org-name-app before continuing.

    * Learn more about Grafana Plugin Development at https://grafana.com/developers/plugin-tools

```

We also provide this documentation on folder structure in the get started guide: https://grafana.com/developers/plugin-tools/#step-2-open-the-generated-folder-structure and more details in: https://grafana.com/developers/plugin-tools/get-started/folder-structure

Note: we are still gathering feedback on improving our documentation structure, any references or recommended placement is subject to change from that exercise.

```[tasklist]
### Tasks
- [x] Document the flags in https://grafana.com/developers/plugin-tools/get-started/prompt-cli-reference
- [x] Improve folder structure documentation to better explain optional elements which could be removed (and why they are there in the first place - i.e. why we recommend them)
- [ ] Remove prompts for the GitHub actions to reduce the number of questions a developer has to answer.
```

GrafanaWriter:
@josmperez - Should you own this issue, or wil @sympatheticmoose ? Either way, can we please triage so it isn't in our project as "Needs Triage"?

josmperez:
[...]

PR #2507, opened 2026-03-04
Plugin Docs: Add asset validation rules

sunker:

This PR adds 8 asset validation rules to the plugin-docs-cli validation engine. These rules enforce image hygiene in plugin documentation: images must live in `img/` directories, use allowed formats (no SVG due to XSS risk), stay within size limits (300KB static, 1MB GIF, 5MB total) and use clean filenames. The rules also detect broken image references in markdown and flag orphaned images that are no longer used.

Severity depends on the validation mode. In strict mode (`validate` command), most rules are errors that block. In non-strict mode (`serve` command), they surface as info-level hints. Two rules - total image size and orphan detection - only run in strict mode.

The epic lists `images-in-img-dir` and `no-images-outside-img` as separate rows but notes they are the same enforcement from opposite directions. They are consolidated into one rule here.

Part of [#771](https://github.com/grafana/grafana-catalog-team/issues/769)

### DIFF
--- a/packages/plugin-docs-cli/src/validation/rules/assets.test.ts
+++ b/packages/plugin-docs-cli/src/validation/rules/assets.test.ts
@@ -1,0 +1,484 @@
+import { describe, it, expect } from 'vitest';
+import { join } from 'node:path';
+import { mkdir, mkdtemp, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { checkAssets } from './assets.js';
+import { Rule } from '../types.js';
+
+const input = (docsPath: string, strict = true) => ({ docsPath, strict });
+
+// helper: valid frontmatter markdown file content
+const md = (body = '') => `---\ntitle: Page\ndescription: A page\n---\n${body}`;
+
+// helper: creates a buffer of a given size in bytes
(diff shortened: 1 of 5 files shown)


mckn (commented):
LGTM! Should we add a test that verifies that you can't add image refs outside of the current root folder? So you can't reference files higher in the folder structure than the plugin docs. 

Just make sure that we don't allow things such as `../../../etc/passwd`

packages/plugin-docs-cli/src/validation/rules/assets.test.ts
```diff
@@ -0,0 +1,416 @@
...
+  // --- image-file-naming ---
+
+  describe('image-file-naming', () => {
+    it('should not report valid image filenames', async () => {
+      const tmp = await mkdtemp(join(tmpdir(), 'asset-test-'));
+      await writeFile(join(tmp, 'index.md'), md());
+      await mkdir(join(tmp, 'img'));
+      await writeFile(join(tmp, 'img', 'my-screenshot_01.png'), bufferOfSize(100));
+
+      const findings = await checkAssets(input(tmp));
```
academo: Make sure to check the length of findings in this test (and all the others)

academo (approved):

PR #2528, opened 2026-03-11
Remove bogus net/http.ResponseWriter.WriteHeader calls from examples

jsing-canva:
As per [Go's documentation](https://pkg.go.dev/net/http#ResponseWriter), the first call to `net/http.ResponseWriter.Write` will trigger an implicit `WriteHeader(http.StatusOK)`, if `WriteHeader` has not already been called explicitly. The HTTP status code and headers must be written prior to content being sent and cannot be changed afterwards.


wbrowne (commented):
Thanks for the feedback @jsing-canva! 

While I see the value in removing this, it can also be helpful to have these to be more explicit about what's being returned (particularly for people who may be new to Golang). It might be better to keep this, but move them to before `.Write` is called instead. WDYT?

PR #2568, opened 2026-04-08
Docs: Skip loading en-US translations in loadResources example

shelldandy:
## Summary
- Updates the `loadResources.ts` code sample in the plugin i18n guide to import `DEFAULT_LANGUAGE` from `@grafana/i18n` and add an early return when the resolved language is the default
- This avoids unnecessary dynamic imports of ~500kb of en-US translations that are already embedded in source code via `t()` and `<Trans>`

Related: grafana/logs-drilldown#1839
Related: grafana/grafana#122188

## Test plan
- [x] Code sample compiles and reads correctly in context of the guide


shelldandy:
Seems like this is no longer viable since doing this messes up the plural strings

PR #2576, opened 2026-04-15
Create Plugin: Pin @grafana dependencies to 12.4.2

leventebalogh:
[Related Slack thread](https://raintank-corp.slack.com/archives/C025WTVRBSN/p1776162568026399)

### Problem
Due to [grafana-ui@12.4.3 not being available](https://www.npmjs.com/package/@grafana/ui/v/12.4.3) while the rest of packages have that version published, `npx @grafana/create-plugin@latest` now scaffolds a broken plugin (other 12.4.3 packages depending a not existing ui package).

### What changed?

Pinned `@grafana/data`, `@grafana/i18n`, `@grafana/runtime`, `@grafana/ui`, and `@grafana/schema` from `^12.4.0` to exact version `12.4.2` in the create-plugin scaffold template.

### DIFF
--- a/packages/create-plugin/templates/common/_package.json
+++ b/packages/create-plugin/templates/common/_package.json
@@ -73,11 +73,11 @@
   },
   "dependencies": {
     "@emotion/css": "11.10.6",
-    "@grafana/data": "^12.4.0",
-    "@grafana/i18n": "^12.4.0",
-    "@grafana/runtime": "^12.4.0",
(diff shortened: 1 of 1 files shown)


leventebalogh:
Closing - actually it's probably a very bad idea to pin those deps for anyone that would create a new plugin in this interim period, and then deal with extra migrations to put them back on the rail again.

PR #2584, opened 2026-04-17
CI: Add nightly E2E run against grafana-dev

sunker:

Adds a nightly GitHub Actions workflow that runs the `@grafana/plugin-e2e` test suite against `grafana-dev` (Grafana `main`) and the latest stable Grafana. Runs daily at 06:00 UTC and also supports manual `workflow_dispatch`. On failure, posts a message to #grafana-plugins-platform-ci with a link to the workflow run's Playwright report summary.

The motivation is catching incompatibilities between `@grafana/plugin-e2e` and Grafana `main` earlier than the current flow, which only surfaces them when a PR that touches `plugin-e2e` is opened.

Also adds a Claude Code skill (`/diagnose-plugin-e2e-incompatibility`) that ranks likely suspect `grafana/grafana` commits for a failing nightly.

Fixes #

The skill is usable manually right now. As a follow-up, once we've seen the skill ranking work well against a few real failures, I'll wire it into this workflow as a `triage` job that runs on failure and posts its ranked suspect list as a threaded Slack reply to the failure notification.


mckn (approved):
LGTM!

PR #2593, opened 2026-04-24
Plugin E2E: use getByRole('switch') for InlineSwitch elements

sunker:

Grafana 13.1.0 restructured `InlineSwitch` and `Switch` for better accessibility, adding `aria-label` directly to the `<label>` element. This causes `getByLabel()` to match two elements - the `<input>` and the `<label>` - and fail in Playwright strict mode. The fix is to use `getByRole('switch', { name: /label/i })`, which scopes the match to only the `<input>` (which carries `role="switch"`) and is unambiguous across Grafana versions.

This PR updates `queryEditor.tlsEnabled.spec.ts` to use the new selector, updates the plugin-e2e docs, and updates the create-plugin AGENTS template so scaffolded plugins follow the correct pattern from the start.

Fixes #

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
           version-resolver-type: plugin-grafana-dependency
           grafana-dependency: '>=8.5.0'
           skip-grafana-dev-image: false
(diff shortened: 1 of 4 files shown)


mckn (approved):
LGTM! Nice work

hugohaggmark (approved):

PR #2617, opened 2026-05-06
fix(website): prevent duplicate startTracking causing rudderQueue error

tolzhabayev:
Guard `startTracking` with a module-level initialized flag so route changes and React strict-mode double-effects do not append a second rudderstack script, whose `onload` would iterate an already-drained (undefined) `rudderQueue` and throw `TypeError: rudderQueue is not iterable`.

@Ukochka — does this look right to you? If so, would you be up for porting the same fix to the other Grafana sub-sites that share this tracking module?

### DIFF
--- a/docusaurus/website/src/theme/tracking/index.ts
+++ b/docusaurus/website/src/theme/tracking/index.ts
@@ -23,6 +23,7 @@ export const isFaroSetup = (config: FaroConfig) =>
 // Enqueue all rudderstack requests until it is loaded and consent is granted
 let rudderstack: Partial<RudderStack> = {};
 let rudderQueue = [];
+let trackingInitialized = false;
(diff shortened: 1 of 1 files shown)


jackw (approved):
LGTM! 🚀 

This error occurs in prod or only dev? https://react.dev/reference/react/useEffect#my-effect-runs-twice-when-the-component-mounts

tolzhabayev:
> LGTM! 🚀
> 
> This error occurs in prod or only dev? https://react.dev/reference/react/useEffect#my-effect-runs-twice-when-the-component-mounts

only in dev due to https://github.com/grafana/plugin-tools/blob/9397fd0f84f80c9d8a6dcfb69f3fbba271200c6b/docusaurus/website/src/theme/Root.tsx#L29

PR #2640, opened 2026-05-20
fix(create-plugin): make playwright report publishing opt-in

sunker:

This removes Playwright report publishing from the scaffolded `create-plugin` CI workflow so generated plugins default to read-only workflow permissions. It also updates the E2E CI docs and snippets to show report publishing as an explicit opt-in setup, including the required job-scoped permissions and the limitations for fork PRs.

This is a precautionary least-privilege change. The previous scaffold did not fully publish Playwright reports by default, but it still encouraged broader permissions than necessary. Making report publishing opt-in keeps the default CI workflow read-only and reduces the blast radius if a dependency, action or workflow step is ever compromised.

Tested in [this PR](https://github.com/sunker/docstest-datasource/pull/19) for a plugin using NPM to verify all other steps work as expected.

### DIFF
--- a/docusaurus/docs/e2e-test-a-plugin/ci.md
+++ b/docusaurus/docs/e2e-test-a-plugin/ci.md
@@ -80,7 +80,7 @@ This guide is based on the example workflow provided earlier in this document.
 
 ### Steps to enable report publishing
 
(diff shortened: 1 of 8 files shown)


jackw (approved):
LGTM! 🚀

PR #2653, opened 2026-05-25
chore: remove optionalDependencies

jackw:

Prior to node 24, NPM shipped with a bug that prevented optional dependencies being written into package.json. For that reason we have a chunk of optionalDependencies in the root package. This PR removes them to reduce dependency footprint. 🤞 


hugohaggmark (approved):
🙌

PR #2655, opened 2026-05-26
feat(create-plugin): remove dependency on glob

theSuess:
In the spirit of reducing dependencies, this PR replaces the `glob` package with the `glob` function from the [Node.js standard library](https://nodejs.org/docs/latest-v26.x/api/fs.html#fspromisesglobpattern-options)

Not sure if we need a migration here. During testing, I realized that the webpack files in most plugins are not using the newest templates and are only partially updated by `006-webpack-nested-fix.ts`.


packages/create-plugin/src/codemods/migrations/migrations.ts
```diff
@@ -71,6 +71,12 @@ export default [
       'Fix ts-node compatibility with the latest @grafana/tsconfig: outdated module/moduleResolution/target overrides break TypeScript 5/6 builds, replaced with nodenext/nodenext/es2022.',
     scriptPath: import.meta.resolve('./scripts/010-ts-node-nodenext.js'),
   },
+  {
+    name: '011-webpack-remove-glob-dependency',
+    version: '7.4.1',
```
theSuess, line 83: Not sure what to put here? Should this PR include the version bump or make a best guess?

PR #2659, opened 2026-05-27
Plugin Docs Parser: Reduce dependency footprint

### LINKED ISSUE #2649: Docs parser: Reduce runtime dependencies
### Which areas does this feature request relate to

- [ ] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E
- [ ] Plugin Meta Extractor
- [ ] Documentation

### Problem

`@grafana/docs-parser` is a public npm package that will be used in the marketing website. Every runtime dependency it ships is an extra link in the supply chain that consumers inherit when they install it, and any compromise of one of those deps becomes our compromise too.

### Solution

Drop what we can inline cheaply.

### Alternatives

_No response_

### Additional context

_No response_

### Are you interested in contributing the solution?

- [ ] Yes
- [ ] No

sunker:

`@grafana/plugin-docs-parser` is about to be installed in the marketing-website repo. Keeping its direct and transitive deps to a minimum reduces the supply-chain attack surface and the ongoing maintenance cost of managing CVEs that bubble up through the tree.

This PR inlines three hardly-used deps and replaces one:

- Inline `hast-util-to-string` as a 4-line helper (one call site, in heading extraction)
- Inline `rehype-slug` as a local plugin with the same slug + dedupe behaviour
- Drop `rehype-raw` and `allowDangerousHtml` - raw HTML in plugin docs is forbidden by spec, so the `parse5` subtree was unused
- Replace `gray-matter` with `js-yaml` v4 plus a small frontmatter regex (gray-matter pulled in `js-yaml@3` plus three other subdeps)

Net effect: 9 direct deps (down from 12) and ~75 fewer packages in the production install tree. 

Fixes https://github.com/grafana/plugin-tools/issues/2649

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14657,6 +14657,13 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/js-yaml": {
+      "version": "4.0.9",
+      "resolved": "https://registry.npmjs.org/@types/js-yaml/-/js-yaml-4.0.9.tgz",
+      "integrity": "sha512-k4MGaQl5TGo/iipqb2UDG2UwjXziSWkh0uysQelTlJpX1qGlpUZYm8PnO4DxG1qBomtJUdYJ6qR6xdIah10JLg==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@types/json-schema": {
       "version": "7.0.15",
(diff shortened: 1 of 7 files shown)


hugohaggmark (approved):
Awesome work 🚀

PR #2662, opened 2026-05-27
chore: add npm overrides to create-plugin  templates

Ukochka:
Added npm overrides to fix transitive dependency vulnerabilities.
This affects grafana-plugin-examples, which is generated from this template.                                             

Fixes https://github.com/grafana/grafana-catalog-team/issues/921


packages/create-plugin/templates/common/_package.json
```diff
@@ -84,5 +84,17 @@
     "react-router-dom": "^6.22.0",
     "rxjs": "7.8.2"{{/if}}
   },
-  "packageManager": "{{ packageManagerName }}@{{ packageManagerVersion }}"
+  "packageManager": "{{ packageManagerName }}@{{ packageManagerVersion }}",
+  "overrides": {
+    "immutable": "5.1.5",
+    "serialize-javascript": "7.0.5",
+    "js-cookie": "3.0.7",
```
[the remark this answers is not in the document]
Ukochka: The overrides field works as intended in our plugin-examples repo which uses npm. Adding per-package-manager templating isn't worth the complexity right now

packages/create-plugin/templates/backend-app/go.mod
```diff
@@ -1,7 +1,9 @@
 module github.com/{{ kebabCase orgName }}/{{ kebabCase pluginName }}
 
 
-go 1.25.5
+go 1.26
```
leventebalogh: I don't think these go updates are necessary for the npm changes, they should probably live in a separate PR?

packages/create-plugin/templates/common/_package.json
```diff
@@ -84,5 +84,17 @@
...
+    "dompurify": "3.4.6",
+    "fast-uri": "3.1.2",
+    "@protobufjs/utf8": "1.1.1",
+    "qs": "6.15.2",
+    "@grafana/ui": {
+      "uuid": "11.1.1"
+    }
```
jackw: I'm not convinced we should be doing this.

These overrides will update the versions of these dependencies across the dependency graph which could cause bugs which are hard to track down in the packages that depend on them.

The `overrides` property is NPM only. Yarn relies on `resolutions` whilst PNPM relies on overrides in its `pnpm-workspace.yaml` file. I think this will create a maintenance burden on us keeping all these in sync.

Ukochka: Could you suggest another way of moving plugin-examples to SLO? These overrides will work on plugin-examples, because we use npm there, in other cases it will be just ignored, or am I mistaken?

packages/create-plugin/templates/backend-app/go.mod
Ukochka: sure, I can do separate PR!

packages/create-plugin/templates/common/_package.json
jackw: I would try to fix them at the source so any current consumers also benefit. I imagine most of these are due to our grafana packages pinning dependencies.

Ukochka:
Opening a new PR

PR #2675, opened 2026-05-28
Plugin E2E: Fix flaky alerting, TLS, and permissions tests on slow CI

mckn:
## What this PR does / why we need it

Fixes flaky E2E test failures observed on `grafana-enterprise@12.1.10` in CI. The self-hosted CI runners are roughly **8x slower** than local development machines (the full test suite takes ~12 minutes in CI vs ~1.5 minutes locally). Under that load, several tests hit the default 5-second `expect` timeout before the page had finished rendering.

These failures never reproduce locally because pages render well within 5 seconds on a fast machine. They only surface in CI where the runners are under heavier load.

### Changes

**`AlertRuleEditPage.ts`** (alerting-scoped only — no global timeout changes):
- Bump `isAdvancedModeSupported()` assertions from 5s to 15s
- `getQueryRow()` now waits for the query editor row to be visible (15s) before returning, so `DataSourcePicker.set()` doesn't race against a still-rendering page
- Bump `evaluate()` preview-button assertion from 5s to 15s

**`queryEditor.tlsEnabled.spec.ts`**: Bump `toBeVisible` from 5s to 15s

**`permissions.spec.ts`**: Replace `waitUntil: 'networkidle'` (inherently flaky under load) with `expect(page).not.toHaveURL(/\/datasources/)` which auto-retries until the redirect completes

### How we verified

We created test harnesses that emulate slow CI conditions using Chrome DevTools Protocol throttling (50x CPU slowdown + 600ms network latency). Under that throttle:

- **Before fix**: tests fail with `expect(locator).toBeVisible() — Timeout: 5000ms` — the exact same error pattern seen in CI
- **After fix**: all tests pass consistently (verified 3 runs under throttle, plus 3 runs without throttle to confirm no regressions)

## Test plan
- [x] All 19 affected tests pass locally without throttle on Grafana 12.1.10
- [x] All slow-sim tests (alerting, TLS, permissions) pass under extreme throttle (50x CPU, 600ms latency)
- [x] No changes to global config — timeouts are scoped to the alerting page model and two individual test files

### DIFF
--- a/packages/plugin-e2e/src/models/pages/AlertRuleEditPage.ts
+++ b/packages/plugin-e2e/src/models/pages/AlertRuleEditPage.ts
@@ -2,9 +2,12 @@ import * as semver from 'semver';
 import { AlertRuleArgs, NavigateOptions, PluginTestCtx, RequestOptions } from '../../types';
 import { GrafanaPage } from './GrafanaPage';
 import { AlertRuleQuery } from '../components/AlertRuleQuery';
-import { expect } from '@playwright/test';
+import { expect, Locator } from '@playwright/test';
 import { isLegacyFeatureEnabled } from '../../fixtures/isFeatureToggleEnabled';
 const QUERY_AND_EXPRESSION_STEP_ID = '2';
+// Alert pages render slowly on busy CI runners; give the rendering 15s headroom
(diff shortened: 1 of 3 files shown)


sunker (approved):
🚀

PR #2676, opened 2026-05-28
ci: replace third-party path-filter actions with local composite

leventebalogh:
### What changed?
- Replaced `dorny/paths-filter` and `tj-actions/changed-files` with a small local composite action at `.github/actions/paths-filter/`

**Notes for the reviewer:**
- Falls back to `changed=true` on non-`pull_request*` events, matching the previous `change-detection` composite's defaulting behavior.
- `metadata-lint.yml` now requests `pull-requests: read` so the composite's `gh api` call is authorized.
- Each caller's pattern list includes `.github/actions/paths-filter/**` so changes to the composite itself trigger downstream re-runs.


.github/actions/paths-filter/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        if [[ ${#patterns[@]} -eq 0 ]]; then
+          echo "No patterns supplied; nothing to match."
+          echo "changed=false" >> "${GITHUB_OUTPUT}"
+          exit 0
+        fi
+
+        echo "Patterns:"
+        printf '  %s\n' "${patterns[@]}"
+
+        mapfile -t changed_files < <(gh api --paginate "/repos/${REPO}/pulls/${PR_NUMBER}/files" --jq '.[].filename')
```
mckn, line 47: This command runs in a subshell. `set -e` cannot catch an error here. If `gh api` fails, the step still reports `changed=false`. Check the exit code of `gh api` on its own.

Note: this API call returns only the first 3000 changed files. A very large pull request can lose data at this point.

.github/workflows/test-build.yml
```diff
@@ -14,18 +14,21 @@ jobs:
     permissions:
       pull-requests: read
     outputs:
-      docs: ${{ steps.filter.outputs.docs }}
+      docs: ${{ steps.filter.outputs.changed }}
     steps:
-      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
mckn, line 19: This step adds a checkout action. The permissions block above does not grant `contents: read`. Add `contents: read` to the permissions block.

.github/actions/paths-filter/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        echo "Changed files (${#changed_files[@]}):"
+        printf '  %s\n' "${changed_files[@]:-(none)}"
+
+        shopt -s extglob
+        any_match=false
+        for file in "${changed_files[@]:-}"; do
+          [[ -z "${file}" ]] && continue
+          for pattern in "${patterns[@]}"; do
+            if [[ "${file}" == ${pattern} ]]; then
```
mckn, line 56: A single `*` in this pattern can match across a `/`. The input description on line 8 says otherwise. Fix the description, or change the match logic.

.github/actions/paths-filter/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        echo "Patterns:"
+        printf '  %s\n' "${patterns[@]}"
+
+        mapfile -t changed_files < <(gh api --paginate "/repos/${REPO}/pulls/${PR_NUMBER}/files" --jq '.[].filename')
+        echo "Changed files (${#changed_files[@]}):"
+        printf '  %s\n' "${changed_files[@]:-(none)}"
+
+        shopt -s extglob
+        any_match=false
+        for file in "${changed_files[@]:-}"; do
```
mckn, line 53: This loop continues after a match is found. Add a check to stop the loop early. This will save time on a large pull request.

.github/actions/paths-filter/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        fi
+
+        echo "Patterns:"
+        printf '  %s\n' "${patterns[@]}"
+
+        mapfile -t changed_files < <(gh api --paginate "/repos/${REPO}/pulls/${PR_NUMBER}/files" --jq '.[].filename')
+        echo "Changed files (${#changed_files[@]}):"
+        printf '  %s\n' "${changed_files[@]:-(none)}"
+
+        shopt -s extglob
```
mckn, line 51: No pattern in this action uses an extended glob. You can remove this line.

mckn (commented):
Great work on this! Found a couple of things that needs to be looked at.

PR #2681, opened 2026-05-29
feat(create-plugin): enable selinux relabeling for podman compatibility

theSuess:

This PR adds the `z` option to the plugin bindmounts which enables [SELinux relabeling](https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label).
Without this option, rootless container tools like [podman](podman.io) prevent the container process from reading the file on SELinux enabled systems.

### DIFF
--- a/packages/create-plugin/src/codemods/migrations/migrations.ts
+++ b/packages/create-plugin/src/codemods/migrations/migrations.ts
@@ -71,6 +71,12 @@ export default [
       'Fix ts-node compatibility with the latest @grafana/tsconfig: outdated module/moduleResolution/target overrides break TypeScript 5/6 builds, replaced with nodenext/nodenext/es2022.',
     scriptPath: import.meta.resolve('./scripts/010-ts-node-nodenext.js'),
   },
+  {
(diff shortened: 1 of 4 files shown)


packages/create-plugin/src/codemods/migrations/migrations.ts
```diff
@@ -72,11 +72,10 @@ export default [
     scriptPath: import.meta.resolve('./scripts/010-ts-node-nodenext.js'),
   },
   {
-    name: '011-secure-sign-script',
-    version: '7.6.3',
-    description:
-      'Security: replace insecure inline `npx --yes @grafana/sign-plugin@latest` sign script with a locked @grafana/sign-plugin devDependency to prevent arbitrary code execution from a compromised @latest publish.',
-    scriptPath: import.meta.resolve('./scripts/011-secure-sign-script.js'),
+    name: '012-enable-compose-selinux-relabel',
+    version: '7.4.1',
```
theSuess, line 76: Again, need guidance here 😬

PR #2692, opened 2026-06-01
chore: update create-github-app-token from 0.2.3 to 0.3.0 (#2691)

yesoreyeram:
chore: update create-github-app-token from 0.2.3 to 0.3.0 which contains a security fix.

Ref: https://github.com/grafana/shared-workflows/releases/tag/create-github-app-token%2Fv0.3.0

### DIFF
--- a/docusaurus/docs/set-up/set-up-github.md
+++ b/docusaurus/docs/set-up/set-up-github.md
@@ -142,7 +142,7 @@ jobs:
     steps:
       - name: Generate GitHub App Token
         id: generate_token
-        uses: actions/create-github-app-token@v2
+       uses: grafana/shared-workflows/actions/create-github-app-token@795f748a236f9de024b7514efc9a208456e7e468 # create-github-app-token/v0.3.0
         with:
           app-id: ${{ secrets.APP_ID }}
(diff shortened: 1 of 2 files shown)


docusaurus/docs/set-up/set-up-github.md
```diff
@@ -142,7 +142,7 @@ jobs:
     steps:
       - name: Generate GitHub App Token
         id: generate_token
-        uses: actions/create-github-app-token@v2
```
mckn, line 145: Does external plugin developers have access to the grafana/shared-workflows ?

mckn (commented):
LGTM! But I think the documentation should be reverted since it is targeting external plugin developers. Or am I thinking about this the wrong way?

docusaurus/docs/set-up/set-up-github.md
yesoreyeram: Yea you are right. Though the workflows public, there is no guidelines yet on how to setup a GATB for external devs. So let's stick with previous guidelines. Reverting the doc changes

mckn (approved):
LGTM!

PR #2689, opened 2026-05-29
ci: replace auto and lerna with release-please

jackw:
**What this PR does / why we need it**:

Bye bye auto. 🤖  Bye bye lerna. 🐉  Bye bye labels. 🏷️  👋 👋 👋 

PR 3 of the Auto → release-please migration epic. Replaces Auto with [`googleapis/release-please-action@v4`](https://github.com/googleapis/release-please-action) driving a PR-driven release lifecycle. release-please opens a long-running release PR; merging it triggers per-package tags, GitHub Releases, and staged npm publishing via OIDC.

Changes in scope:

- Adds `release-please-config.json` declaring the 10 published packages, with the `node-workspace` plugin for cross-package dep updates, `tag-separator: "@"` to preserve the existing `@grafana/<pkg>@<version>` tag format, and `bump-minor-pre-major: true` to mimic Auto's 0.x behaviour
- Adds `.release-please-manifest.json` seeded from current npm-registry versions
- Adds `.github/workflows/release-please.yml` — runs release-please-action; on releases (gated by gh env requiring review from frontend-platform or catalog team member), builds, runs `npm stage publish`, posts a Slack notification with links to GH releases + npm pages
- Adds `.github/workflows/check-release-config.yml` — fails PRs that touch `packages/**` without a matching entry in `release-please-config.json`
- Removes the `release` job from `.github/workflows/ci.yml`
- Drops `auto`, `@auto-it/*`, and `lerna` devDeps; hoists `nx` as a direct devDep (was transitive via Lerna)
- Deletes `.autorc`, `lerna.json`, root `CHANGELOG.md`
- Rewrites `CONTRIBUTING.md` "Create A Release" section for the new flow + Conventional Commit PR-title convention
- Bumps `AGENTS.md` tech-stack line `Nx >=21` → `Nx >=22` to match the installed major

**Which issue(s) this PR fixes**:

related: #2685
related: grafana/grafana-catalog-team/issues/943

**Special notes for your reviewer**:

Issue #2685, opened 2026-05-29
Epic: Migrate NPM publishing from auto to release-please


jackw:
## Background

We need to stop direct pushes to `main`. Today, `auto shipit` (in `.github/workflows/ci.yml` on push to `main`) writes two `[skip ci]` commits and N tags directly to `main` via the platform bot. That needs to change.

## Goal

Replace Auto's main-branch publishing with [`googleapis/release-please-action@v4`](https://github.com/googleapis/release-please-action) driving a PR-driven release lifecycle:

- release-please opens a long-running release PR with version bumps + per-package changelogs
- Merging the PR triggers per-package tags, GitHub Releases, and npm publishing (with provenance via OIDC)
- No direct writes to `main` from CI

## Tasks
- [x] Replace Auto-driven canary publishing with pkg.pr.new ([merged](https://github.com/grafana/plugin-tools/pull/2677))
- [x] side-effects ahead of the migration ([merged](https://github.com/grafana/plugin-tools/pull/2684))
- [x] release-please migration ([merged](https://github.com/grafana/plugin-tools/pull/2689))
- [x] Auto retirement cleanup ([merged](https://github.com/grafana/plugin-tools/pull/2702))

PR #2700, opened 2026-06-08
ci: use scoped package names for release-please tags

jackw:

The first release-please run after the migration ([#2689](https://github.com/grafana/plugin-tools/pull/2689)) opened release PR [#2699](https://github.com/grafana/plugin-tools/pull/2699) with two problems:

1. **Broken tag continuity.** release-please's `node` strategy strips the npm scope when deriving the git tag component (`@grafana/create-plugin` → `create-plugin`), so it proposed tags like `create-plugin@7.8.0` instead of the historical `@grafana/create-plugin@7.8.0` format used by every existing tag.
2. **Bloated changelogs.** Because the expected tag (`create-plugin@7.7.0`) doesn't exist — the real tag is `@grafana/create-plugin@7.7.0` — release-please couldn't locate the prior release and scanned from the start of history, listing hundreds of already-released commits.

Setting an explicit `component` (the full scoped name) per package bypasses the scope-stripping normalization. Confirmed against `release-please/src/strategies/base.ts:136` — `options.component` is used verbatim; normalization only runs on the derived fallback.

Result: tags return to `@grafana/<pkg>@<version>`, and release-please finds the existing prior tags → scans only new commits → clean changelogs.

related: https://github.com/grafana/plugin-tools/issues/2685
related: https://github.com/grafana/grafana-catalog-team/issues/943

- `package-name` is kept (npm identity, drives the `node-workspace` dependency-graph matching); `component` is the new addition (git tag identity). They default from the same source but `component`'s default normalization strips the scope — hence the explicit override.
- **Verify-before-publish:** once this merges, the Release workflow re-runs and updates the open release PR #2699 in place. Confirm its compare links read `@grafana/<pkg>@<old>...@grafana/<pkg>@<new>` and the changelogs are short before merging #2699 to trigger the real staged publish.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -22,15 +22,27 @@
     { "type": "ci", "section": "CI", "hidden": true }
   ],
   "packages": {
-    "packages/create-plugin": { "package-name": "@grafana/create-plugin" },
-    "packages/eslint-plugin-plugins": { "package-name": "@grafana/eslint-plugin-plugins" },
-    "packages/plugin-docs-cli": { "package-name": "@grafana/plugin-docs-cli" },
-    "packages/plugin-docs-parser": { "package-name": "@grafana/plugin-docs-parser" },
-    "packages/plugin-e2e": { "package-name": "@grafana/plugin-e2e" },
-    "packages/plugin-meta-extractor": { "package-name": "@grafana/plugin-meta-extractor" },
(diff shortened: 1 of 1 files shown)


ashharrison90 (approved):
lgtm! 👍

joshhunt (approved):

PR #2708, opened 2026-06-09
feat(deps): bump grafana libs to 13.0.2

Ukochka:
Bump @grafana/* dependencies from 12.4.2 to 13.0.2 to align scaffolded plugins with the latest Grafana 13 release.


jackw (changes requested):
If we're bumping these I think we also need to bump the version in docker-compose as [this renovate PR](https://github.com/grafana/plugin-tools/pull/2481/changes) did.

jackw (approved):
LGTM! 🚀

PR #2710, opened 2026-06-09
fix(deps): bump copy-webpack-plugin to 14.0.0

Ukochka:
Bumping copy-webpack-plugin from ^13.0.0 to ^14.0.0 fixes two transitive vulnerabilities:                                                                  
                                                                                                                                                           
 - serialize-javascript: v13 pulls 6.0.2 (vulnerable to RCE via RegExp.flags and CPU exhaustion DoS). v14 depends on ^7.0.3 which is patched.               
  - fast-uri: resolves to a patched version (3.1.2+) on fresh lockfile generation, fixing path traversal and host confusion vulnerabilities.

### DIFF
--- a/packages/create-plugin/templates/common/_package.json
+++ b/packages/create-plugin/templates/common/_package.json
@@ -74,7 +74,7 @@
   },
   "dependencies": {
     "@emotion/css": "11.10.6",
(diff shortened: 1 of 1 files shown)


jackw (approved):
LGTM! 🚀

PR #2721, opened 2026-06-11
fix(sign-plugin): prevent symlink escape check bypass and add test coverage

tolzhabayev:

`@grafana/sign-plugin` had almost no test coverage (a single file with two assertions). This PR brings it to 36 tests across 4 files:

- **`buildManifest`** — sha256 per file, nested directories, `MANIFEST.txt` exclusion, posix path keys (regression guard for the Windows `\` → `/` sanitisation), internal symlinks included, symlinks escaping the base directory rejected, missing `plugin.json`. Runs against real temp dirs created with the `tmp` package (the repo convention for test temp dirs, see #2010 — keeps CodeQL from flagging `os.tmpdir()` usage).
- **`sign` command integration tests** — real fixture dist dirs with only `postData` mocked: happy path writes the signed `MANIFEST.txt`, `GRAFANA_COM_URL` override, `--signatureType`/`--rootUrls` passthrough, invalid root URL exits before any network call, missing token exits, `GRAFANA_API_KEY` deprecation warning still proceeds, non-200/network errors exit without writing a manifest, `createPlugin.version` picked up from `.config/.cprc.json`.
- **`signManifest`/`saveManifest` units** — success, non-200 with JSON and non-JSON bodies, network error propagation, write failure.
- **`pluginValidation` + `getCreatePluginVersion`** — all validation branches, `assertRootUrlIsValid`, present/absent/malformed `.cprc.json`.

To make error paths testable, `signManifest` now takes the token as a parameter and throws on failure instead of calling `process.exit(1)`; `saveManifest` likewise throws. Token resolution (including the `GRAFANA_API_KEY` deprecation warning) moved into the sign command, which owns all output and exits. The integration tests were written before this refactor and pass unchanged after it, so user-facing behavior is preserved.

n/a

- The fixture helper realpaths temp dirs because on macOS `os.tmpdir()` returns `/var/...` which resolves to `/private/var/...`, which would otherwise falsely trip the symlink-escape check in `walk`.
- `getPluginJson` in `pluginValidation.ts` uses `require()` in an ESM package and has no callers — it looks like dead code, but removing it felt out of scope for this PR.


packages/sign-plugin/src/utils/tests/fixtures.ts
```diff
@@ -0,0 +1,34 @@
+import { mkdirSync, mkdtempSync, realpathSync, writeFileSync } from 'node:fs';
```
jackw: We use https://github.com/raszi/node-tmp in this repo to handle creation and deletion of temp directories and files as recommended by codeql which was creating noise about os.tmpdir usage in tests.

codeql: https://github.com/grafana/plugin-tools/security/code-scanning/24

fix pr: https://github.com/grafana/plugin-tools/pull/2010

jackw (dismissed):
Nice!

Took it for a spin locally and changes look good.

Image

Left a comment about temp file creation. I'm not sure if codeql is broken or the rules changed.

packages/sign-plugin/src/utils/tests/fixtures.ts
tolzhabayev: Good call - switched the fixtures over to tmp in 797162c9.

tolzhabayev:
@jackw need another 👍

jackw (dismissed):
LGTM! 🚀

packages/sign-plugin/src/utils/tests/fixtures.ts
```diff
@@ -0,0 +1,34 @@
+import { mkdirSync, realpathSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
+import { dirSync } from 'tmp';
```
mckn, line 3: I think it might be worth looking into dropping the dependency on 'tmp' (if possible).

It might be possible to replace it with: `mkdtempSync + realpathSync from node:fs`

packages/sign-plugin/src/utils/manifest.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  });
+
+  it('should propagate network errors', async () => {
+    mocks.postData.mockRejectedValue(new Error('socket hang up'));
+
+    await expect(signManifest(manifest, 'test-token')).rejects.toThrow('socket hang up');
+  });
+});
+
+describe('saveManifest', () => {
```
mckn, line 187: Should we also update this test to use the "tempDirs cleanup pattern" that you have in the other tests? If it is possible.

mckn (dismissed):
LGTM! Great work on this. Something that we might want to open an issue about:

symlink-to-directory within base dir: walk classifies entries as either isFile() or isSymbolicLink() (via lstat). A symlink pointing to a directory inside the base dir would pass the escape check and be yielded, then hashFile would fail trying to read a directory as a file. This isn't introduced here and is out of scope, but worth a follow-up issue.

packages/sign-plugin/src/utils/tests/fixtures.ts
tolzhabayev: Yeah, I left it because of https://github.com/grafana/plugin-tools/pull/2721#discussion_r3394862426

packages/sign-plugin/src/utils/manifest.test.ts
tolzhabayev: > Should we also update this test to use the "tempDirs cleanup pattern" that you have in the other tests? If it is possible.

good idea, here it is https://github.com/grafana/plugin-tools/pull/2721/changes/c1a24dab47f6077b856b9d21ebc92eca8e9d625e#diff-51c38a2838733feeb986bfeafc9e94a52f57c8be8014e2bf0a9b92f267b8b201R175

tolzhabayev:
@jackw ptal again

jackw (approved):
LGTM! 🚀

PR #1597, opened 2025-03-06
Feature: Consistent CLI output

jackw:
**What this PR does / why we need it**:

Our cli apps in this repo currently print out messaging in all sorts of different ways. This PR tries to align all these plugin tools so they give our users messaging in a consistent format. By providing information like the tool name, it's running version, and colour coded feedback I think we can make things a little clearer for our lovely plugin developers.

To do this I've introduced a couple of "internal libraries" and switched out `tsc` for `tsup` so these libraries are bundled into the dist code. Having a single lib for this also means we can remove all marked, marked-terminal, and boxen usage and the dependencies from create-plugin.

**What does it look like?**

The following screenshots try to show the differences in terminal messages. 
_Note: the first run of a command is using the latest release of the package. It's the second run that shows the changes in this pr._

create-plugin scaffolding a plugin:
image

create-plugin updating a plugin:
image

create-plugin errors:
image

sign-plugin errors:
image

sign-plugin success:
image

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/create-plugin@5.19.0-canary.1597.13828816481.0
  npm install @grafana/eslint-plugin-plugins@0.3.0-canary.1597.13828816481.0
  npm install @grafana/plugin-e2e@1.19.0-canary.1597.13828816481.0
  npm install @grafana/plugin-meta-extractor@0.5.0-canary.1597.13828816481.0
  npm install @grafana/plugin-types-bundler@0.3.0-canary.1597.13828816481.0
  npm install @grafana/sign-plugin@3.1.0-canary.1597.13828816481.0
  # or 
  yarn add @grafana/create-plugin@5.19.0-canary.1597.13828816481.0
  yarn add @grafana/eslint-plugin-plugins@0.3.0-canary.1597.13828816481.0
  yarn add @grafana/plugin-e2e@1.19.0-canary.1597.13828816481.0
  yarn add @grafana/plugin-meta-extractor@0.5.0-canary.1597.13828816481.0
  yarn add @grafana/plugin-types-bundler@0.3.0-canary.1597.13828816481.0
  yarn add @grafana/sign-plugin@3.1.0-canary.1597.13828816481.0
  ```
</details>

PR #2722, opened 2026-06-11
fix: replace find-up dependency with native @libs/find-up helper

tolzhabayev:

Removes the `find-up` runtime dependency from `@grafana/create-plugin` and `@grafana/sign-plugin` (introduced in #1597) by replacing it with a ~20 line native helper in a new private `@libs/find-up` workspace.

`find-up` was only used for two synchronous, files-only lookups:

- `@libs/version` — walking up from the bundled module to find the consuming package's `package.json` (this is why both CLIs carried `find-up` in `dependencies`: the lib is bundled into their dist while imports of published packages stay external).
- `create-plugin`'s `utils.packageManager.ts` — finding the closest lockfile to detect the package manager.

The native helper preserves the two find-up semantics that matter for these call sites: all candidate names are checked in each directory before moving up (lockfile precedence), and the walk stops at the filesystem root. It is covered by 7 unit tests.

With this change both CLIs bundle the helper directly, so `find-up` and its transitive dependencies (`locate-path`, `p-locate`, `yocto-queue`, `unicorn-magic`) disappear from the published packages' install footprint. The remaining `find-up` entries in `package-lock.json` are unrelated transitive dependencies of the docusaurus website tooling.

n/a

- Verified in the built output: `dist` no longer contains any `import ... from 'find-up'`; `@libs/version` and `utils.packageManager` now import the bundled helper via relative paths.
- `libs/version` previously declared `find-up` as a `^7.0.0` peerDependency while consumers shipped `8.0.0` — that drift goes away too.
- The `fix:` prefix means release-please will cut patch releases for both CLIs, which seems right for a change to the published artifacts.

### DIFF
--- a/libs/find-up/README.md
+++ b/libs/find-up/README.md
@@ -1,0 +1,18 @@
+# Libs/Find-Up
+
+A utility library that finds a file by walking up parent directories. A minimal, native replacement for the [`find-up`](https://www.npmjs.com/package/find-up) npm package covering the synchronous, files-only use cases in this repository.
+
+## Usage
+
+```ts
+import { findUpSync } from '@libs/find-up';
+
+// Find the closest package.json starting from the current working directory.
+const packageJsonPath = findUpSync('package.json');
+
+// Find the closest of several files starting from a given directory.
+// All names are checked in each directory before moving up to the parent.
+const lockfilePath = findUpSync(['yarn.lock', 'pnpm-lock.yaml', 'package-lock.json'], { cwd: '/some/dir' });
+```
+
+Returns the absolute path of the first match, or `undefined` if no match was found before reaching the filesystem root.
--- a/libs/find-up/package.json
+++ b/libs/find-up/package.json
@@ -1,16 +1,14 @@
 {
-  "name": "@libs/version",
-  "version": "1.0.2",
+  "name": "@libs/find-up",
+  "version": "1.0.0",
   "private": true,
   "type": "module",
   "main": "./src/index.ts",
   "types": "./src/index.ts",
-  "dependencies": {
-    "@libs/find-up": "^1.0.0"
-  },
   "scripts": {
     "typecheck": "tsc --noEmit",
     "lint": "eslint ./src",
-    "lint:fix": "eslint --fix ./src"
+    "lint:fix": "eslint --fix ./src",
+    "test": "vitest"
   }
 }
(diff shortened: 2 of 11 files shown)


mckn (dismissed):
Nice addition! Overall it looks great but there is one change you need to apply to be consistent with the original package:

The original find-up package delegates to locate-path, which defaults to type: 'file' and checks stat.isFile() using statSync — so it only matches files, not directories, by default.

The @libs/find-up implementation currently uses existsSync, which matches both files and directories. To mirror the original behavior, existsSync should be replaced with a statSync + .isFile() check, exactly as locate-path does.

tolzhabayev:
> Nice addition! Overall it looks great but there is one change you need to apply to be consistent with the original package:
> 
> The original find-up package delegates to locate-path, which defaults to type: 'file' and checks stat.isFile() using statSync — so it only matches files, not directories, by default.
> 
> The @libs/find-up implementation currently uses existsSync, which matches both files and directories. To mirror the original behavior, existsSync should be replaced with a statSync + .isFile() check, exactly as locate-path does.

Good point, changed it here
https://github.com/grafana/plugin-tools/pull/2722/changes/5c1d0712858bbc17df952da05b00413b26020fa7

mckn (commented):
LGTM! The only thing we should change is to sync the version pinning across the consumers. I think the "^1.0.0" is okay here since this is a library living in the same workspace but I guess both works.

Minor — version pin inconsistency:
// create-plugin/package.json
"@libs/find-up": "1.0.0"    // exact

// sign-plugin/package.json
"@libs/find-up": "^1.0.0"   // range

// libs/version/package.json
"@libs/find-up": "^1.0.0"   // range

tolzhabayev:
> LGTM! The only thing we should change is to sync the version pinning across the consumers. I think the "^1.0.0" is okay here since this is a library living in the same workspace but I guess both works.
> 
> Minor — version pin inconsistency: // create-plugin/package.json "@libs/find-up": "1.0.0" // exact
> 
> // sign-plugin/package.json "@libs/find-up": "^1.0.0" // range
> 
> // libs/version/package.json "@libs/find-up": "^1.0.0" // range

it's following the existing pattern. Each consumer pins its `@libs/*` deps the same way: `create-plugin` uses exact (`@libs/output 1.0.3`, `@libs/version 1.0.2`), while `sign-plugin` and `libs/version` use ranges. I am fine either way but it would then make this lib follow a different convention.

jackw (approved):
LGTM! 🚀

>The only thing we should change is to sync the version pinning across the consumers.

We can discuss that in a future PR I think. The devDeps don't need to be pinned, it's the prod ones that we must pin.

tolzhabayev:
@mckn need you to approve or remove request changes otherwise I can't merge

mckn (approved):
LGTM!

PR #2723, opened 2026-06-11
chore: import eslint config

jackw:

The eslint config package that lives in https://github.com/grafana/eslint-config-grafana goes somewhat unnoticed and requires work to bring it up to speed with our security best practices. Rather than duplicate a load of CI / config I think it makes more sense to bring it over here where we'll keep a closer eye on it and get all the security hardening efforts made in this repo for free.

Fixes #

> [!IMPORTANT]
> We must do the following once merged and before doing any releases:
> 
> 1. update the access settings for the eslint-config package on npmjs.com
> 2. push a tag on main so release-please works


joshhunt (dismissed):

packages/eslint-config/README.md
```diff
@@ -0,0 +1,82 @@
...
+The following official blog posts are available for interested parties:
+
+- https://eslint.org/blog/2022/08/new-config-system-part-1/
+- https://eslint.org/blog/2022/08/new-config-system-part-2/
+- https://eslint.org/blog/2022/08/new-config-system-part-3/
+
+This package contains a single flat config object which can be imported like so:
+
+```js
+const grafanaConfig = require("@grafana/eslint-config/flat");
```
sunker, line 36: is this still relevant?

sunker (commented):
Not sure if it would make sense to dogfood this by letting the other packages in this monorepo use local package rather than published version?

packages/eslint-config/README.md
jackw: No it's not. I'll do it in a separate PR to validate releases are working. Thanks @sunker !

sunker (approved):
LGTM

PR #2743, opened 2026-06-24
test(plugin-e2e): fix appConfig settings test for Grafana 13.1.0

sunker:

The plugin-e2e `appConfig` settings test broke against Grafana 13.1.0 (nightly). 13.1.0 moved plugin settings to the new `@grafana/runtime` cached pluginSettings service, so the plugin meta GET now fires after navigation. The test's method-agnostic `{ times: 1 }` mock then swallowed that GET and returned an empty body, so the enable/disable button never rendered and the click timed out.

Fix: scope the mock to the save POST and let the meta GET reach the backend. The test now exercises the real flow instead of masking the failure. Verified against grafana-enterprise 13.1.0 and 11.4.0.


mckn (approved):
LGTM!

PR #2759, opened 2026-07-08
chore: Skip Vault secret steps on fork PRs

leventebalogh:
Related PR: https://github.com/grafana/plugin-tools/pull/2757 (a fork PR that was hitting this limitation)

### The problem
Currently **fork PRs** cannot get green CI for the following two reasons:
- `playwright.yml` runs the `get-vault-secrets` DockerHub step unconditionally, which fails with `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL` because fork runs never receive an OIDC id-token. 
- In `ci.yml` the secret steps are already gated (#2547), but the sign-plugin steps reference `fromJSON(steps.get-secrets.outputs.secrets)` in their `env`, which GitHub evaluates even when the step is skipped — so `fromJSON('')` fails the job with a template error.

### What changed?
The DockerHub secret + login steps in `playwright.yml` are skipped on fork PRs (anonymous image pulls, rate-limit risk only), and the `fromJSON` consumers fall back to `'{}'` so skipped secrets no longer break template evaluation. E2E tests still run on fork PRs — they never needed the secrets.

**Notes for the reviewer:**
Non-fork behavior is unchanged — this PR itself should show the secret steps running and the Docker Hub login succeeding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,7 +171,9 @@ jobs:
       - name: '@grafana/sign-plugin - use GRAFANA_ACCESS_POLICY_TOKEN to sign generate-panel plugin'
         if: ${{ github.event.pull_request.head.repo.fork != true }}
         env:
-          GRAFANA_ACCESS_POLICY_TOKEN: ${{ fromJSON(steps.get-secrets.outputs.secrets).GRAFANA_ACCESS_POLICY_TOKEN }}
+          # Step env is evaluated even when the step is skipped, so fall back to '{}'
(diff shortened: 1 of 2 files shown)


konstantinmv (dismissed):

Ukochka (dismissed):

tolzhabayev (dismissed):

toddtreece (dismissed):

jackw (dismissed):
LGTM! 🚀

jackw (approved):
LGTM! 🚀

PR #2760, opened 2026-07-08
fix(plugin-e2e): handle table-based preview of values in toDisplayPreviews

tolzhabayev:
Since Grafana 13.1.0 the multiPropsVariables behavior is always on (toggle removed in grafana/grafana#126905). Any variable query result with a data frame field not named `value` or `text` now carries that field as an option property, and the variable editor previews the options in a table instead of the label list. That's every SQL-style variable query, so `toDisplayPreviews` times out waiting for the `previewOfValuesOption` locator, which never appears. Reported by the AWS datasources squad after their plugin-e2e variable tests started failing on 13.1.0.

The matcher now waits for either presentation and, when the table is shown, asserts against its text column, which holds the same display text the labels used to show. Older Grafana versions still render the labels and keep the old code path.

The new test mocks a variable query response with a `humidity` field. It renders as labels on older versions and as the table on 13.1+, and it fails on 13.1.0 without this fix. Verified locally against grafana-enterprise 11.4.0 and 13.1.0.


packages/plugin-e2e/src/matchers/toDisplayPreviews.ts
```diff
@@ -12,12 +12,27 @@ export const toDisplayPreviews = async (
...
+  // not named "value" or "text") are previewed in a table instead of a list of labels
+  const previewTable = variableEditPage.getByGrafanaSelector(Edit.CustomVariable.previewTable);
+
   try {
-    await expect(
-      variableEditPage.getByGrafanaSelector(
-        variableEditPage.ctx.selectors.pages.Dashboard.Settings.Variables.Edit.General.previewOfValuesOption
-      )
-    ).toContainText(previewTexts, options);
+    await expect(previewLabels.first().or(previewTable).first()).toBeVisible({ timeout: options?.timeout });
```
leventebalogh: I think one `.first()` could be enough here?


```suggestion
    await expect(previewLabels.or(previewTable).first()).toBeVisible({ timeout: options?.timeout });
```

leventebalogh (dismissed):
LGTM 👍

packages/plugin-e2e/src/matchers/toDisplayPreviews.ts
tolzhabayev: Good call, applied.

leventebalogh (approved):

PR #2758, opened 2026-07-08
chore(alerting): Gate alertingQueryAndExpressionsStepMode FT on Grafana version

konstantinmv:
## What

Update `AlertRuleEditPage.isAdvancedModeSupported()` to account for the removal of the alertingQueryAndExpressionsStepMode feature toggle in Grafana 13.2.0.

## Why

In Grafana 13.2.0 the toggle was removed and the advanced query/expression mode switch became always available (GA). 

##How

Advanced mode is now considered supported when either:
- Grafana >= 13.2.0 (toggle gone, feature always on), or
- the alertingQueryAndExpressionsStepMode toggle is enabled (Grafana < 13.2.0)

 Follow-up to grafana/grafana#125086.

PR #2762, opened 2026-07-09
fix(plugin-e2e): adapt alert rule advanced mode detection to Grafana 13.2

tolzhabayev:

Grafana 13.2 removed the frontend usage of the `alertingQueryAndExpressionsStepMode` feature toggle (grafana/grafana#125086), which broke the nightly E2E run ([29000729249](https://github.com/grafana/plugin-tools/actions/runs/29000729249)) in two ways:

- `AlertRuleEditPage.isAdvancedModeSupported()` decided the editor mode by reading the toggle from boot data. On 13.2 the toggle is inert, so the page objects assumed the legacy editor while Grafana rendered the simplified step, and `getQueryRow()` timed out looking for a query row title that doesn't exist there. It now detects the mode from the DOM instead: wait for the query step to mount, then check for the advanced mode switch.
- `alerting.basicMode.spec.ts` forces the toggle off to test the legacy no-switch editor. That configuration no longer exists on >= 13.2, so the spec is skipped there. alerting.advancedMode.spec.ts covers the same scenarios on those versions.

Also makes the boot data override mutate the existing `featureToggles` and `user` objects in place instead of replacing them. The runtime config keeps a reference to the original object, so replacing it means plugins never see the override when the callback runs after config construction. This is what made `queryEditor.tlsEnabled` flaky on nightly.

Verified against grafana-enterprise nightly (13.2.0-28985753600, the build that failed in CI), 13.1.0 and 11.0.11: the alerting and feature-toggles suites are green on all three.

n/a

### DIFF
--- a/packages/plugin-e2e/src/fixtures/scripts/overrideGrafanaBootData.js
+++ b/packages/plugin-e2e/src/fixtures/scripts/overrideGrafanaBootData.js
@@ -14,18 +14,14 @@ export const overrideGrafanaBootData = ({ featureToggles, userPreferences }) => 
     if (Object.keys(featureToggles).length > 0) {
       console.log('@grafana/plugin-e2e: setting the following feature toggles', featureToggles);
 
-      // override feature toggles with the ones provided by the test
-      window.grafanaBootData.settings.featureToggles = {
-        ...window.grafanaBootData.settings.featureToggles,
-        ...featureToggles,
-      };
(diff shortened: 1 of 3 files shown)


tolzhabayev:
Alternative to https://github.com/grafana/plugin-tools/pull/2758

konstantinmv (approved):

leventebalogh (approved):
LGTM 👍

PR #2774, opened 2026-07-13
feat: add AI-agent assisted migrations to create-plugin update

jackw:

This PR lets `create-plugin update` use locally installed AI agent CLIs (Claude Code, Codex) during migrations. Some upgrade work can't be reliably codemodded (e.g. user-authored configs, plugin source refactors) so migrations can now ship natural-language instructions that an agent applies under user supervision.

There's no issue for this. I threw it together to spark discussion around how we can further leverage AI to prevent plugin devs getting bogged down in broken tooling updates. I believe it could also help writing codemods as a single line change to a ts file is easier to write as a prompt than as code. Most of what I'm proposing lives in the type union in `migrations/migrations.ts`.

Things I'd like decided:

1. Are three migration types right, or should prompt-only migrations be a separate command rather than part of `update`?
2. Is a handoff file a good completion signal or too fragile?
3. Do we want create-plugin commits to contain agent-authored edits?

**How it works**

- Migrations come in three varieties: script-only (what we have today), prompt-only (a `.md` instructions file), and hybrid (codemod first then agent finishes). This PR ships the plumbing and an unregistered example prompt (`prompts/000-example.md`).
- Agents are detected on PATH. When a queued migration has a prompt and an agent is available, `update` asks before using it - `--agent[=<id>]` / `--no-agent` skip the question.
- The agent runs interactively (inherited stdio) and signals completion through a handoff file that `create-plugin` polls. A failed handoff aborts the update without stamping the version, and a session that exits without handing off anything falls back to an abort-or-continue prompt.
- Agentic runs default per-migration commits on unless the user passed `--commit` / `--no-commit`, so each agent step sees an isolated diff.
- If `update` is already running inside an agent session, no inner agent is spawned. Any deferred steps are printed as a directive block for the outer agent.

This PR lays the groundwork without affecting existing plugins on update (all the existing migrations are untouched and no prompt migrations are registered). CI and existing workflows are unaffected too as non-TTY envs disable agent assistance. This repos ci workflow job `test-updates` covers this end to end.

**Where to start**

- `migrations/migrations.ts` - the type union the three shapes hang off.
- `agentic/resolve.ts` - gates when any of this feature is enabled.
- `agentic/prompts.ts` - the system prompt, the handoff contract, and the escaping to stop a prompt file breaking out of its XML section.
- `agentic/index.ts` - the orchestration and the various ways a prompt step can end.
- `codemods/AGENTS.md` and `prompts/000-example.md` - the contract a prompt author should follow.
- `detect.ts`, `handoff.ts`, `runner.ts` and `definitions.ts` - mechanical and can be skimmed.

This PR includes a new AI-assisted migrations page and an `--agent` flag reference. If we want to run with this they'll need fleshing out.

Fixes:

The interactive session can be run manually in a real terminal:

1. Register `prompts/000-example.md` as a test migration in `migrations/migrations.ts`.
2. Build create-plugin and npm link it.
3. Run the local build's `update` in a plugin and follow the agent prompt.

I haven't run the interactive session in a real terminal yet. The gates (non-TTY, `--no-agent`, `--agent` on non-TTY, inside-agent) are covered by unit tests with agent detection and the session runner mocked, so nothing has been tested agaiThe Codex invocation args mirror Nx's implementation but are unverified.

### DIFF
--- a/packages/create-plugin/src/codemods/agentic/definitions.ts
+++ b/packages/create-plugin/src/codemods/agentic/definitions.ts
@@ -1,5 +1,3 @@
-import { homedir } from 'node:os';
-import { join } from 'node:path';
 import { AgentDefinition } from './types.js';
 
 // Pure data. Invocation args mirror the Nx agentic migrate implementation (nrwl/nx#35718, #35889)
@@ -9,7 +7,6 @@ export const AGENT_DEFINITIONS: AgentDefinition[] = [
     id: 'claude-code',
     displayName: 'Claude Code',
     binaryNames: ['claude'],
-    wellKnownPaths: [join(homedir(), '.claude', 'local', 'claude')],
     buildInteractive: (invocationContext) => ({
       args: [
         '--system-prompt',
@@ -25,7 +22,6 @@ export const AGENT_DEFINITIONS: AgentDefinition[] = [
     id: 'codex',
     displayName: 'OpenAI Codex',
     binaryNames: ['codex'],
-    wellKnownPaths: [],
     buildInteractive: (invocationContext) => ({
       args: ['-c', `developer_instructions=${invocationContext.systemPrompt}`, invocationContext.userPrompt],
     }),
--- a/packages/create-plugin/src/codemods/agentic/definitions.ts
+++ b/packages/create-plugin/src/codemods/agentic/definitions.ts
@@ -1,0 +1,33 @@
+import { homedir } from 'node:os';
+import { join } from 'node:path';
+import { AgentDefinition } from './types.js';
+
+// Pure data. Invocation args mirror the Nx agentic migrate implementation (nrwl/nx#35718, #35889)
+// and must be verified against the currently shipping agent CLIs when changed.
+export const AGENT_DEFINITIONS: AgentDefinition[] = [
+  {
+    id: 'claude-code',
+    displayName: 'Claude Code',
+    binaryNames: ['claude'],
+    wellKnownPaths: [join(homedir(), '.claude', 'local', 'claude')],
+    buildInteractive: (invocationContext) => ({
+      args: [
+        '--system-prompt',
+        invocationContext.systemPrompt,
+        // pre-authorize the handoff write so the session cannot stall on a permission prompt
+        '--allowedTools',
+        `Write(${invocationContext.handoffDir}/**)`,
+        invocationContext.userPrompt,
+      ],
+    }),
+  },
+  {
+    id: 'codex',
+    displayName: 'OpenAI Codex',
+    binaryNames: ['codex'],
+    wellKnownPaths: [],
+    buildInteractive: (invocationContext) => ({
+      args: ['-c', `developer_instructions=${invocationContext.systemPrompt}`, invocationContext.userPrompt],
+    }),
+  },
+];
--- a/packages/create-plugin/src/codemods/agentic/index.ts
+++ b/packages/create-plugin/src/codemods/agentic/index.ts
@@ -1,0 +1,199 @@
+import { output, selectPrompt } from '../../utils/utils.console.js';
+import {
+  getPackageManagerExecCmd,
+  getPackageManagerSilentInstallCmd,
+  getPackageManagerWithFallback,
+} from '../../utils/utils.packageManager.js';
+import { Context } from '../context.js';
+import { Migration } from '../migrations/migrations.js';
+import { createRunId, ensureRunDir, getHandoffPath, getRunDir, wipeMigrateRuns } from './handoff.js';
+import {
+  buildDirectiveBlock,
+  buildHybridUserPrompt,
+  buildNextStepsList,
+  buildPromptOnlyUserPrompt,
+  buildSystemPrompt,
+  CodemodChanges,
+  DeferredMigration,
+  getPromptPath,
+  loadPromptInstructions,
+} from './prompts.js';
+import { resolveAgenticMode } from './resolve.js';
+import { runAgentSession } from './runner.js';
+import { AgenticResolution, AgentSessionResult, PromptStepResult } from './types.js';
+
+export const ABORT_CHOICE = 'Abort the update';
+export const CONTINUE_CHOICE = 'Treat the migration step as completed';
+
+export interface AgenticRuntimeOptions {
+  // raw minimist --agent value: undefined | '' | '<id>' | false
+  agentFlag: string | boolean | undefined;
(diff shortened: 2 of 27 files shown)


packages/create-plugin/src/codemods/agentic/definitions.ts
```diff
@@ -0,0 +1,33 @@
...
+import { AgentDefinition } from './types.js';
+
+// Pure data. Invocation args mirror the Nx agentic migrate implementation (nrwl/nx#35718, #35889)
+// and must be verified against the currently shipping agent CLIs when changed.
+export const AGENT_DEFINITIONS: AgentDefinition[] = [
+  {
+    id: 'claude-code',
+    displayName: 'Claude Code',
+    binaryNames: ['claude'],
+    wellKnownPaths: [join(homedir(), '.claude', 'local', 'claude')],
```
joshhunt: fwiw, i think it's more straight forward to just not try to guess where else claude could be other than path. feels almost kind of sneaky to do this.

joshhunt (commented):
Interesting! I initially thought it would be better to use an agent to write a deterministic codemod, but I understand your reasoning to use agents to perform actions that are tricky to determinately code up front.

Just to make sure - the intention here is that migrations _within this repo_ are written as `000-migration.md` and plugin developers in them with the create-plugin tooling? Plugin developers themselves wouldn't be writing migrations like this in their repo?

I'm unsure how to feel about this. It feels like an awkward mixing of tool fidelity. I completely understand the advantages this has, but I think if I run `create-plugin update` I'm expecting something deterministic to happen. Maybe, could these be surfaced as skills for users to invoke in their coding agent instead?

jackw:
> Interesting! I initially thought it would be better to use an agent to write a deterministic codemod, but I understand your reasoning to use agents to perform actions that are tricky to determinately code up front.

I've written nearly all the codemods in this repo using an agent skill. I've had claude open a PR with it [here](https://github.com/grafana/plugin-tools/pull/2782).

> Just to make sure - the intention here is that migrations _within this repo_ are written as `000-migration.md` and plugin developers in them with the create-plugin tooling? Plugin developers themselves wouldn't be writing migrations like this in their repo?

Yes, migrations continue to live in this repo rather than plugin devs writing their own migrations in plugin repos.

jackw:
>I'm unsure how to feel about this. It feels like an awkward mixing of tool fidelity. I completely understand the advantages this has, but I think if I run create-plugin update I'm expecting something deterministic to happen. Maybe, could these be surfaced as skills for users to invoke in their coding agent instead?

It is a bit of an odd one but I'm trying to think of ways to leverage our AI overlords to make updates easier on devs. I'm thinking these prompts will be able to self-heal / fix the things that we can't codemod (e.g. customisations devs have made to eslint, webpack that we can't possibly know or want to cater for). 

I'm not sure about skills - they don't have a migration manager and registry so a skill would need to reimplement these things to know what to run and when. Then there's the whole agent format fragmentation between claude, codex, whatever new agent where as in this way we have a single prompt file that serves them all.

Maybe we gate spawning strictly behind `--agent` so by default `update` doesn't ask and skipped prompts just get printed as next steps (with a note about the flag)?

packages/create-plugin/src/codemods/agentic/resolve.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (isInsideAgent(options.env)) {
+    output.log({
+      title: 'Running inside an AI agent. Prompt migrations will be handed to it instead of spawning another agent.',
+    });
+    return { mode: 'inside-agent' };
+  }
+
+  if (options.agentFlag === false) {
+    return { mode: 'disabled', reason: 'flag' };
+  }
```
mckn, line 27: Should this one be placed before the `if (isInsideAgent(options.env)) {` or is this done by design?

packages/create-plugin/src/codemods/agentic/index.ts
```diff
@@ -0,0 +1,199 @@
...
+      }
+      output.warning({
+        title: 'Some migrations include AI-agent instructions that were not applied automatically.',
+        body: buildNextStepsList(deferredMigrations),
+      });
+      return;
+    }
+
+    if (appliedMigrations.length > 0) {
+      output.statusList('success', appliedMigrations);
```
mckn, line 185: I think this one only will format the list of appliedMigrations. Should we also print them to the console?

packages/create-plugin/src/codemods/migrations/manager.ts
```diff
@@ -32,10 +37,34 @@ export async function runMigrations(migrations: Migration[], options: RunMigrati
...
+
+    let agentApplied = false;
+    if (hasPromptStep(migration) && agentic) {
+      // runs after the codemod has flushed and before the commit so the commit includes agent edits
+      const step = await agentic.runPromptStep(migration, context);
+      agentApplied = step.kind !== 'deferred';
+    }
+
+    const treeDirty = agentApplied ? !(await isGitDirectoryClean()) : false;
+    const shouldCommit = commitEachMigration && ((context?.hasChanges() ?? false) || treeDirty);
```
mckn, line 67: This might not be a bug but I will mention it regardless. If you run the migrations with `--force` mode enabled the `performPreCodemodChecks` will be skipped and you will apply agentic updates and commit them even if they might not be migration updates.

packages/create-plugin/src/codemods/agentic/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    const handoffPath = getHandoffPath(runDir, migration.name);
+    ensureRunDir(runDir);
+
+    const { packageManagerName, packageManagerVersion } = getPackageManagerWithFallback();
+    const systemPrompt = buildSystemPrompt({
+      workspaceRoot: options.basePath,
+      packageManagerName,
+      packageManagerVersion,
+      installCmd: getPackageManagerSilentInstallCmd(packageManagerName, packageManagerVersion),
+      execCmd: getPackageManagerExecCmd(packageManagerName, packageManagerVersion),
```
mckn, line 87: Is this correct? Doesn't the `getPackageManagerExecCmd` return `yarn dlx @grafana/create-plugin` ?

What you want is only the package manager right? Or am I reading the prompts incorrectly?

mckn (commented):
I really like this addition and think it will be really helpful going forward. 

With this in place I think it is important to give the agent instructions (in the prompt) on how it can verify that the migration was successful. 

So maybe allow it to run unit tests etc. to verify? Or can we give it other ways to verify a successful migration?

PR #2778, opened 2026-07-14
fix: remove suspense boundaries from app plugin templates

joshhunt:

When Grafana loads app plugins, it already wraps them in a suspense boundary at the route level (see [GrafanaRoute.tsx](https://github.com/grafana/grafana/blob/8229a35035554b678f9f01c79a80e854716aef5a/public/app/core/navigation/GrafanaRoute.tsx#L58)). This ensures that all pages across Grafana - core or from plugins - share the same initial loading state.

The App templates teaches plugin developers to wrap their app in a Suspense boundary with a LoadingPlaceholder, which is problematic for two reasons:
 - [LoadingPlaceholder](https://developers.grafana.com/ui/canary/index.html?path=/story/information-loadingplaceholder--basic) is not a suitable loading placeholder for a whole app
   - Instead, in `@grafana/ui` 13.2 we'll release [PageLoader](https://developers.grafana.com/ui/canary/index.html?path=/story/information-pageloader--basic) that is designed to be displayed full screen
 - App plugins including their own suspense boundary at the top of the app defeats the Suspense boundary the Grafana page is wrapped in, meaning they don't get the unified loading UI.

 - [ ] Validate all places an App component is rendered in Grafana is wrapped in `<Suspense />`
 - [ ] Validate all places an AppConfig component is rendered in Grafana is wrapped in `<Suspense />`
 - [ ] Decide what to do for all other plugin component entry points. Are they already wrapped in suspense in Grafana core? It would be great if we could tell plugin devs "you never need a top-level suspense boundary".

### DIFF
--- a/packages/create-plugin/templates/app/src/module.tsx
+++ b/packages/create-plugin/templates/app/src/module.tsx
@@ -1,12 +1,12 @@
 import { lazy } from 'react';
 import { AppPlugin } from '@grafana/data';
 
-const LazyApp = lazy(() => import('./components/App/App'));
-const LazyAppConfig = lazy(() => import('./components/AppConfig/AppConfig'));
+const App = lazy(() => import('./components/App/App'));
+const AppConfig = lazy(() => import('./components/AppConfig/AppConfig'));
 
-export const plugin = new AppPlugin<{}>().setRootPage(LazyApp).addConfigPage({
+export const plugin = new AppPlugin<{}>().setRootPage(App).addConfigPage({
   title: 'Configuration',
   icon: 'cog',
-  body: LazyAppConfig,
+  body: AppConfig,
   id: 'configuration',
 });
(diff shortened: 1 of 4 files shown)


packages/create-plugin/templates/app/src/module.tsx
```diff
@@ -1,22 +1,8 @@
-import React, { Suspense, lazy } from 'react';
-import { AppPlugin, type AppRootProps } from '@grafana/data';
-import { LoadingPlaceholder } from '@grafana/ui';
-import type { AppConfigProps } from './components/AppConfig/AppConfig';
+import { lazy } from 'react';
+import { AppPlugin } from '@grafana/data';
 
-const LazyApp = lazy(() => import('./components/App/App'));
```
jackw, line 6: I think you'll wanna do the same in https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/scenes-app/src/module.tsx#L14-L24

ashharrison90 (dismissed):
lgtm 👍

jackw (dismissed):
LGTM 🚀

hugohaggmark (approved):

ashharrison90 (approved):
lgtm 👍

PR #2782, opened 2026-07-15
chore: add write-codemod skill for create-plugin codemods

jackw:
## What this PR does

Adds a `write-codemod` Claude Code skill at `.claude/skills/write-codemod/`, alongside the existing `bump-plugin-e2e` and `diagnose-plugin-e2e-incompatibility` skills, so every contributor working with an AI agent gets it automatically. It originated as a private skill that has been producing good results; this version is restructured and expanded for shared use.

The skill teaches an agent how to author codemods for `packages/create-plugin/src/codemods/`:

- The migration vs. addition decision, registration conventions, and the version-gating rule (`.config/.cprc.json`, next-patch versioning)
- The `runner.ts` execution pipeline and the reasoning behind the core rules (Context-only file access, graceful degradation, idempotency)
- CLI flags via valibot `schema` exports, debug logging, and `renderTemplate` for scaffolding files from the templates
- A workflow for deriving a migration from an already-made change to `templates/` (path mapping, Handlebars rendering, divergent user files)
- Testing requirements including the `toBeIdempotent` matcher, and local verification steps

Detailed per-format manipulation patterns (JSON/JSONC, TypeScript AST via recast, YAML) live in `references/` files that agents load only when relevant, keeping the main skill lean.

All API signatures and behaviours described in the skill were verified against the current source.

### DIFF
--- a/.claude/skills/write-codemod/references/yaml.md
+++ b/.claude/skills/write-codemod/references/yaml.md
@@ -1,0 +1,62 @@
+# Modifying YAML files
+
+Use the `yaml` npm package for structured manipulation of `.yml` / `.yaml` files. Never use string replacement on YAML — it breaks on formatting and comment differences.
+
+```ts
+import { parseDocument, stringify, visit, YAMLMap, YAMLSeq, Scalar, Pair, type Document } from 'yaml';
+```
+
+Note: migrations that modify YAML may be `async` — the codemod signature allows returning `Promise<Context>`, and the existing YAML examples use `async function migrate`.
+
+## Core pattern: parse → navigate → mutate → stringify
+
+```ts
+const source = context.getFile('docker-compose.yaml') || '';
+const doc = parseDocument(source);
+
+// navigate with path arrays
+const value = doc.getIn(['services', 'grafana', 'image']); // read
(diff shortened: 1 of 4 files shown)


.claude/skills/write-codemod/references/yaml.md
```diff
@@ -0,0 +1,62 @@
+# Modifying YAML files
+
+Use the `yaml` npm package for structured manipulation of `.yml` / `.yaml` files. Never use string replacement on YAML — it breaks on formatting and comment differences.
+
+```ts
+import { parseDocument, stringify, visit, YAMLMap, YAMLSeq, Scalar, Pair, type Document } from 'yaml';
```
leventebalogh: Maybe I am wrong here, but looks like `type Document` is not used in the examples below? Maybe import `type visitorFn` instead (line 47)?

leventebalogh (dismissed):
LGTM 👍

.claude/skills/write-codemod/references/yaml.md
jackw: You are correct as ever @leventebalogh . Fixed in https://github.com/grafana/plugin-tools/pull/2782/commits/d417541345cc98041b194c91105527657e88f39c

leventebalogh (approved):

PR #2792, opened 2026-07-21
chore(deps): bump nx to v23.1.0

tolzhabayev:

Bumps nx from 22.7.3 to 23.1.0 using `nx migrate`. Both generated migrations were no-ops for this repo: we have no `releaseTagPattern` config, and `.gitignore` already covers `.nx`. Nx 23 needs Node 22+ and ESLint v9, we are on Node 24 and ESLint 9 already.

Build, typecheck, lint and tests all pass on the new version.

The lockfile churn is nx swapping out its platform binaries for the new major.


jackw (approved):
LGTM! 🚀

Issue #2805, opened 2026-07-27
e2e tests crash on Playwright 1.62: cannot resolve tsconfig "extends": "@grafana/tsconfig"


tolzhabayev:
Playwright 1.62 made unresolvable tsconfig `extends` paths a fatal error (microsoft/playwright#41571). Its config loader only probes paths relative to the extending file plus a `node_modules` folder next to it, so the scaffolded `"extends": "@grafana/tsconfig"` in `.config/tsconfig.json` can never resolve. Before 1.62 this failed silently and the empty config went unnoticed, now `playwright test` crashes on startup:

```
Error: Failed to load tsconfig file at /path/to/plugin/tsconfig.json:
Failed to resolve "extends" path "@grafana/tsconfig" referenced from /path/to/plugin/.config/tsconfig.json
```

Every scaffolded plugin that floats to `@playwright/test` 1.62 is affected. First seen in grafana/grafana-plugin-examples#721 where all five e2e matrix jobs fail with this error.

Upstream regression report: microsoft/playwright#41989.

Plan:
- pin `@playwright/test` to 1.61.x in the template and via a migration (short term)
- once upstream resolves the issue, drop the pin. If they instead remove the `node_modules` fallback entirely, we need to rework how the scaffolded tsconfig consumes `@grafana/tsconfig`

Keeping this open until the pin is removed.

PR #2806, opened 2026-07-27
fix(create-plugin): pin @playwright/test to 1.61.x

tolzhabayev:

Playwright 1.62 made unresolvable tsconfig `extends` paths a fatal error (microsoft/playwright#41571). Its config loader can't resolve the package specifier in the scaffolded `"extends": "@grafana/tsconfig"`, it only probes paths relative to the extending file, so `playwright test` now crashes on startup for every scaffolded plugin that floats to 1.62. Before 1.62 the unresolved extends was silently swallowed, which is why this never surfaced.

This pins `@playwright/test` to `~1.61.1` in the template. Existing plugins manage their own playwright version through renovate/dependabot and won't merge a broken bump, so no migration.

Upstream regression report: microsoft/playwright#41989, fixed by microsoft/playwright#42005 which hasn't shipped in a release yet. The pin comes off once it does, tracked in #2805.

Part of #2805 (stays open to track removing the pin)

First seen in grafana/grafana-plugin-examples#721 where all five e2e matrix jobs fail with `Failed to resolve "extends" path "@grafana/tsconfig"`. Verified by scaffolding a plugin from this branch: `@playwright/test` resolves to 1.61.1 and `playwright test --list` loads the config and tests cleanly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,6 +210,13 @@ jobs:
         run: npx create-plugin update --force
         working-directory: ./${{ env.WORKING_DIR }}
 
+      # Temporarily pin playwright to match the template: the scaffolded plugin has a caret range
+      # that resolves to 1.62, which crashes on tsconfigs extending @grafana/tsconfig.
+      # Remove once playwright ships a fix and the template pin is lifted.
+      - name: Pin playwright version
+        run: npm install --no-audit --save-dev @playwright/test@~1.61.1
+        working-directory: ./${{ env.WORKING_DIR }}
+
       - name: Lint plugin frontend
         run: npm run lint
         working-directory: ./${{ env.WORKING_DIR }}
--- a/packages/create-plugin/templates/common/_package.json
+++ b/packages/create-plugin/templates/common/_package.json
@@ -20,7 +20,7 @@
     "@grafana/plugin-e2e": "^3.10.0",
     "@grafana/sign-plugin": "^3.3.3",
     "@grafana/tsconfig": "^2.2.0",
-    "@playwright/test": "^1.57.0",{{#if useExperimentalRspack}}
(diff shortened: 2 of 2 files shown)


tolzhabayev:
@jackw ptal

packages/create-plugin/src/codemods/migrations/scripts/013-pin-playwright.ts
```diff
@@ -0,0 +1,27 @@
+import type { Context } from '../../context.js';
+import { addDependenciesToPackageJson } from '../../utils.js';
+
+// Playwright 1.62 made unresolvable tsconfig "extends" paths a fatal error, and its config
+// loader cannot resolve the package specifier in the scaffolded `"extends": "@grafana/tsconfig"`,
+// so `playwright test` crashes on startup. Pin to 1.61.x until this is fixed upstream:
+// https://github.com/microsoft/playwright/issues/41989
```
jackw: @tolzhabayev I'm not sure we need a migration here. Existing plugins tend to care for their own dependencies via renovate / dependabot. If they're bumping to 1.62 and it's breaking they won't merge it anyhow.

tolzhabayev: Fair enough, dropped the migration. The PR is just the template pin now.

jackw (dismissed):
LGTM! 🚀

jackw:
Urgh... the update command smoke test once again bites us! 😞 Shall we disable it for the time being?

tolzhabayev:
> Urgh... the update command smoke test once again bites us! 😞 Shall we disable it for the time being?

How about we pin it too for now instead of disabling it? https://github.com/grafana/plugin-tools/pull/2806/changes/2500b7780b5a87f3928cffab17df7223c7dd1abc

jackw (approved):
LGTM! 🚀

PR #2812, opened 2026-08-01
fix(plugin-e2e): type window.grafanaBootData as @grafana/data's BootData

### LINKED ISSUE #2811: plugin-e2e: Window.grafanaBootData ambient type conflicts with @grafana/data's BootData (TS2717)
## What happened

`packages/plugin-e2e/src/index.ts` declares a global ambient augmentation:

```ts
declare global {
  interface Window {
    monaco: any;
    grafanaBootData: {
      settings: {
        featureToggles: Record<string, boolean>;
        buildInfo: { version: string };
        namespace: string;
      };
    };
  }
}
```

Grafana core (`packages/grafana-data/src/types/config.ts` in grafana/grafana) declares its own,
much wider `BootData` type for the same `window.grafanaBootData`:

```ts
export interface BootData {
  user: CurrentUserDTO;
  settings: GrafanaConfig;
  navTree: NavLinkDTO[];
  assets: { light: string; dark: string };
  _femt?: boolean;
}
```

When a single `tsc` program includes both this package's shipped `.d.ts` and Grafana's own
ambient declaration (e.g. a Grafana core tsconfig that typechecks both app source and
`@grafana/plugin-e2e`), TypeScript requires merged global interface property declarations to
have the *identical* type (TS2717: "Subsequent property declarations must have the same type").
The two shapes are structurally incompatible (plugin-e2e's is a strict, narrower subset), so
this is normally a hard compile error.

In grafana/grafana this is currently masked only because `skipLibCheck: true` is inherited from
the root tsconfig, which suppresses TS2717 for declaration files. With the error suppressed,
*which* declaration "wins" the merge for e2e-playwright type-checking depends on unspecified
compiler file-processing order — verified empirically: reversing an `include` array order in
Grafana's `e2e-playwright/tsconfig.json` flips which declaration wins and produces 3 real type
errors in `packages/grafana-runtime/src/config.ts` and
`packages/grafana-ui/src/components/DateTimePickers/WeekStartPicker.tsx`. Grafana currently
works around this with an "order matters" comment pinning the include order — a fragile
mechanism that a TS upgrade, a `skipLibCheck` change, or an innocent include-list reorder could
silently flip.

## Proposed fix

Type `grafanaBootData` as `import('@grafana/data').BootData` directly, so the merged property
is structurally identical to Grafana core's own declaration and TS2717 no longer applies
regardless of file order or `skipLibCheck`. I have this change ready locally (adds
`@grafana/data` as a `dependencies` entry of `plugin-e2e`, which it did not previously depend
on) and have verified it resolves the reversed-include-order regression in a local grafana/grafana
checkout. Happy to open a PR — wanted to raise the approach with maintainers first since it
changes this package's public ambient type surface and adds a new dependency.

### Caveat found during verification

The fix removes the *silent* TS2717 landmine, but if a consumer's own `@grafana/data` (e.g. via
workspace source) has drifted from the `@grafana/data` version `plugin-e2e` depends on, a
*different*, non-silent type error can still surface (duplicate-package type structural drift,
e.g. between two versions' `IconName` unions) — this is normal npm/yarn dependency duplication,
not a defect in this fix, but worth being aware of for anyone who also pins `@grafana/data`
narrowly.

## Environment

- `@grafana/plugin-e2e` 3.10.0
- Verified against `grafana/grafana` (branch cut from `origin/main`)

unlikelyzero:
## What this does

Types the global `Window.grafanaBootData` ambient augmentation in `plugin-e2e` as
`import('@grafana/data').BootData` instead of a hand-rolled inline subset.

Fixes: #2811

## Why

The previous inline shape was a strict subset of Grafana core's real `BootData` type. When a
single `tsc` program merges this package's ambient `Window` declaration with Grafana's own
(e.g. Grafana core's tsconfig for typechecking `e2e-playwright` alongside app source), the
merged property types must be identical or TypeScript raises TS2717. Grafana core currently
masks this via `skipLibCheck: true` and works around the resulting file-order sensitivity with
an explicit `include` array ordering (see grafana/grafana `e2e-playwright/tsconfig.json`) — a
fragile mechanism, since which declaration silently wins depends on unspecified compiler
behavior.

Using Grafana's own `BootData` type directly removes the structural mismatch at the source, so
downstream consumers no longer need to work around merge order.

## Changes

- `packages/plugin-e2e/src/index.ts`: `grafanaBootData` is now typed as
  `import('@grafana/data').BootData`.
- `packages/plugin-e2e/package.json`: adds `@grafana/data` as a `dependencies` entry (it was not
  previously a dependency of this package).

## Verification

- `npm run typecheck -w @grafana/plugin-e2e`, `npm run build -w @grafana/plugin-e2e`,
  `npm run lint -w @grafana/plugin-e2e`, and `npm run test -w @grafana/plugin-e2e` all pass.
- Packed the built package and installed it as a local `file:` dependency in a `grafana/grafana`
  checkout. With the fix installed, reversing the `include` array order in
  `e2e-playwright/tsconfig.json` (which previously reproduced 3 TS2717-driven errors in
  `packages/grafana-runtime/src/config.ts` and
  `packages/grafana-ui/src/components/DateTimePickers/WeekStartPicker.tsx`) no longer reproduces
  those errors. With the normal include order, `typecheck:e2e` stays clean.

## Note for reviewers

Raised the approach in #2811 first since this changes the package's public ambient type surface
and adds a new dependency (`@grafana/data`) that `plugin-e2e` did not previously have. Open to
an alternative shape if maintainers prefer not to take that dependency (e.g. dropping
`grafanaBootData` from the global augmentation entirely and keeping it as a package-local type).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2112,7 +2112,6 @@
       "version": "7.0.1",
       "resolved": "https://registry.npmjs.org/@braintree/sanitize-url/-/sanitize-url-7.0.1.tgz",
       "integrity": "sha512-URg8UM6lfC9ZYqFipItRSxYJdgpU5d2Z4KnjsJ+rj6tgAmGme7E+PQNCiud8g0HDaZKMovu2qjfa0f5Ge0Vlsg==",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@coffeeandfun/remove-pii": {
@@ -5162,7 +5161,6 @@
       "version": "2.3.6",
       "resolved": "https://registry.npmjs.org/@formatjs/ecma402-abstract/-/ecma402-abstract-2.3.6.tgz",
       "integrity": "sha512-HJnTFeRM2kVFVr5gr5kH1XP6K0JcJtE7Lzvtr3FS/so5f1kpsqqqxy5JF+FRaO6H2qmcMfAUIox7AJteieRtVw==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@formatjs/fast-memoize": "2.2.7",
@@ -5175,7 +5173,6 @@
       "version": "2.2.7",
       "resolved": "https://registry.npmjs.org/@formatjs/fast-memoize/-/fast-memoize-2.2.7.tgz",
       "integrity": "sha512-Yabmi9nSvyOMrlSeGGWDiH7rf3a7sIwplbvo/dlz9WCIjzIQAfy1RMf4S0X3yG724n5Ghu2GmEl5NJIV6O9sZQ==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "tslib": "^2.8.0"
@@ -5185,7 +5182,6 @@
       "version": "0.7.6",
       "resolved": "https://registry.npmjs.org/@formatjs/intl-durationformat/-/intl-durationformat-0.7.6.tgz",
       "integrity": "sha512-jatAN3E84X6aP2UOGK1jTrwD1a7BiG3qWUSEDAhtyNd1BgYeS5wQPtXlnuGF1QRx0DjnwwNOIssyd7oQoRlQeg==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@formatjs/ecma402-abstract": "2.3.6",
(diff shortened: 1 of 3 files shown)


sunker:
Thanks for the detailed issue and for raising the PR.

I'd rather not take `grafana/data` on as a dependency just for this though, that's a permanent cost for every consumer to fix something narrow. 

Looking closer at this, window.grafanaBootData augmentation in plugin-e2e is actually only needed internally by one fixture, it's not part of the real public API. So instead of widening it to match `grafana/data`'s BootData, we can drop the global augmentation entirely and type it locally where it's used. That removes the conflict for anyone pulling in both types, not just grafana/grafana's e2e-playwright program, and needs no new dependency. Can you try this approach instead?

PR #2621, opened 2026-05-08
feat(plugin-e2e): add missing components to the components fixture

mckn:
**What this PR does / why we need it**:

Follow-up to #2583. Adds six more Grafana UI components to the `components` fixture: **Select**, **MultiSelect**, **Switch**, **RadioGroup**, **UnitPicker**, and **ColorPicker**. Each component gets a `static getContainer()` method for version-conditional container resolution and a `within(root)` method for DOM scoping, matching the pattern established by `dataSourcePicker` and `timeRangePicker`.

Uses CSS/structural fallback selectors that work across all supported Grafana versions (8.5+). Once [grafana/grafana#124120](https://github.com/grafana/grafana/pull/124120) merges and `@grafana/e2e-selectors` is updated, a follow-up will add `data-testid` container selectors for Grafana >= 13.1.0. The six component classes are also now exported from the package public API.

**Which issue(s) this PR fixes**:

Follows up on #2583 and depends on https://github.com/grafana/grafana/pull/124120 for stable container selectors.

**Special notes for your reviewer**:

- `switch` is used as a property name on the `Components` class. It's a JS reserved word but valid as a property name in ES6+ class definitions.
- The fallback CSS selectors (e.g. `[class*="-grafana-select-value-container"]`) rely on Grafana's Emotion-generated class naming convention and will be replaced by stable `data-testid` selectors in a follow-up.
- MultiSelect test is deferred — no provisioned panel currently uses a multi-select option.

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @grafana/plugin-docs-cli@0.1.0-canary.2621.26576348885.0
  npm install @grafana/plugin-docs-parser@0.1.0-canary.2621.26576348885.0
  npm install @grafana/plugin-e2e@3.9.0-canary.2621.26576348885.0
  # or 
  yarn add @grafana/plugin-docs-cli@0.1.0-canary.2621.26576348885.0
  yarn add @grafana/plugin-docs-parser@0.1.0-canary.2621.26576348885.0
  yarn add @grafana/plugin-e2e@3.9.0-canary.2621.26576348885.0
  ```
</details>

PR #2816, opened 2026-08-03
docs: document the components fixture in the e2e testing docs

mckn:

Documents the `components` fixture in the plugin-e2e developer docs, which was previously not covered at all. Adds a **Components** section to "Use the API" following the same pattern as the Pages section — an example, a table of the exposed components (`dataSourcePicker`, `timeRangePicker`, and the six added in #2621: `select`, `multiSelect`, `switch`, `radioGroup`, `unitPicker`, `colorPicker`), the `within()` scoping pattern, and a link to the source directory — and extends the expect-matchers section with the paired `toHaveSelected`/`toBeChecked`/`toHaveChecked`/`toHaveColor` matchers. Also reworks the manual Select and InlineSwitch recipes in "Select UI elements" to recommend the fixture, which handles version differences and the re-render race the old force-click recipe worked around.

Follows up on #2621.

While writing the docs I noticed the `toBeChecked` matcher declares `options?: { on?: boolean }` but its implementation reads `options?.checked`, so the docs use `.not.toBeChecked()` for the unchecked assertion. Worth aligning in a follow-up.

### DIFF
--- a/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
+++ b/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
@@ -74,7 +74,7 @@ explorePage.getQueryEditorRow('A').getByLabel('Query').fill('sum(*)');
 
 Here are some examples demonstrating how to interact with a few UI components that are common in plugins. The `InlineField` and `Field` component can be used interchangeably.
 
-For many common Grafana UI components, such as select, switch, radio group, and color picker, `@grafana/plugin-e2e` provides ready-made component models through the [`components` fixture](./use-the-api.md#components). Prefer the `components` fixture whenever it covers the component you're testing - it handles UI differences between Grafana versions and actionability quirks for you.
+For many common Grafana UI components, such as `select`, `switch`, `radio group`, or `color picker`, `@grafana/plugin-e2e` provides ready-made component models through the [`components` fixture](./use-the-api.md#components). When it covers the component you're testing, use the `components` fixture, since it handles UI differences between Grafana versions and other elements and the component you're targeting.
 
 #### Input field
 
--- a/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
+++ b/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
@@ -141,7 +141,7 @@ You can use the `InlineSwitch` component to interact with the UI.
 </div>
 ```
 
-Use the `switch` model from the [`components` fixture](./use-the-api.md#components) to interact with the component. It resolves the switch element across Grafana versions, and its `check` and `uncheck` methods are idempotent - they only click the switch when the state needs to change, which avoids races with React's re-render cycle.
+Use the `switch` model from the [`components` fixture](./use-the-api.md#components) to interact with the component. It resolves the switch element across Grafana versions, and its `check` and `uncheck` methods are idempotent, so they only click the switch when the state needs to change, which avoids races with React's re-render cycle.
 
 ```ts title="Playwright test"
 const tlsField = page.getByTestId('config-tls-field');
--- a/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
+++ b/docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
@@ -74,6 +74,8 @@ explorePage.getQueryEditorRow('A').getByLabel('Query').fill('sum(*)');
 
 Here are some examples demonstrating how to interact with a few UI components that are common in plugins. The `InlineField` and `Field` component can be used interchangeably.
 
+For many common Grafana UI components, such as select, switch, radio group, and color picker, `@grafana/plugin-e2e` provides ready-made component models through the [`components` fixture](./use-the-api.md#components). Prefer the `components` fixture whenever it covers the component you're testing - it handles UI differences between Grafana versions and actionability quirks for you.
+
 #### Input field
 
 You can use the `InlineField` component to interact with the UI.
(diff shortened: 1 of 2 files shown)


docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
```diff
@@ -74,6 +74,8 @@ explorePage.getQueryEditorRow('A').getByLabel('Query').fill('sum(*)');
 
 Here are some examples demonstrating how to interact with a few UI components that are common in plugins. The `InlineField` and `Field` component can be used interchangeably.
 
+For many common Grafana UI components, such as select, switch, radio group, and color picker, `@grafana/plugin-e2e` provides ready-made component models through the [`components` fixture](./use-the-api.md#components). Prefer the `components` fixture whenever it covers the component you're testing - it handles UI differences between Grafana versions and actionability quirks for you.
```
urbiz-grafana:
```suggestion
For many common Grafana UI components, such as `select`, `switch`, `radio group`, or `color picker`, `@grafana/plugin-e2e` provides ready-made component models through the [`components` fixture](./use-the-api.md#components). When it covers the component you're testing, use the `components` fixture, since it handles UI differences between Grafana versions and other elements and the component you're targeting.
```

urbiz-grafana: @mckn - I guess this is what you mean by "actionability quirks"?

docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
```diff
@@ -129,27 +134,23 @@ await expect(page.getByRole('checkbox', { name: 'TLS Enabled' })).not.toBeChecke
...
+  <InlineField label="TLS Enabled">
+    <InlineSwitch label="TLS Enabled" value={value} onChange={handleOnChange} />
+  </InlineField>
+</div>
 ```
 
-Use `getByRole('switch', { name: /TLS Enabled/i })` rather than `getByLabel` — newer versions of Grafana add `aria-label` to the `<label>` element itself, which causes `getByLabel` to match multiple elements and fail in strict mode.
-
-Use `isChecked()` combined with `click({ force: true })` rather than `check()`/`uncheck()`. The `check()` and `uncheck()` methods assert the final state after clicking, which can race with React's re-render cycle on newer versions of Grafana and cause intermittent failures.
+Use the `switch` model from the [`components` fixture](./use-the-api.md#components) to interact with the component. It resolves the switch element across Grafana versions, and its `check` and `uncheck` methods are idempotent - they only click the switch when the state needs to change, which avoids races with React's re-render cycle.
```
urbiz-grafana:
```suggestion
Use the `switch` model from the [`components` fixture](./use-the-api.md#components) to interact with the component. It resolves the switch element across Grafana versions, and its `check` and `uncheck` methods are idempotent, so they only click the switch when the state needs to change, which avoids races with React's re-render cycle.
```

docusaurus/docs/e2e-test-a-plugin/use-the-api.md
```diff
@@ -57,6 +57,65 @@ test('test annotation query', async ({ gotoAnnotationEditPage }) => {
 
 To learn how to provision the Grafana instance with the resources that you need, refer to the [Set up resources](./setup-resources.md) guide.
 
+### Components
+
+Component model objects represent Grafana UI components such as the select, switch, radio group, and color picker components. Like pages, they encapsulate common UI operations in one place and handle UI deviations between different versions of Grafana. Use the `components` fixture to interact with a Grafana UI component instead of writing your own locators.
```
urbiz-grafana:
```suggestion
Component model objects represent Grafana UI components such as the `select`, `switch`, `radio group`, and `color picker` components. Like pages, they encapsulate common UI operations in one place and handle UI deviations between different versions of Grafana. Use the `components` fixture to interact with a Grafana UI component instead of writing your own locators.
```

docusaurus/docs/e2e-test-a-plugin/use-the-api.md
```diff
@@ -57,6 +57,65 @@ test('test annotation query', async ({ gotoAnnotationEditPage }) => {
...
+| `components.select`           | `selectOption(value)`                                       |
+| `components.multiSelect`      | `selectOptions([values])`                                   |
+| `components.switch`           | `check()`, `uncheck()`                                      |
+| `components.radioGroup`       | `check(labelOrValue)`                                       |
+| `components.unitPicker`       | `selectOption(path)` where `path` is like `'Misc > Pixels'` |
+| `components.colorPicker`      | `selectOption(rgbOrHex)`                                    |
+
+To see the full list of components exposed by `@grafana/plugin-e2e`, refer to the Github [repository](https://github.com/grafana/plugin-tools/tree/main/packages/plugin-e2e/src/models/components).
+
+#### Scoping a component to a part of the page
```
urbiz-grafana:
```suggestion
#### Scope a component to a part of the page
```

urbiz-grafana (dismissed):
LGTM, other than a few suggestion and a question at the very beginning.

docusaurus/docs/e2e-test-a-plugin/selecting-ui-elements.md
mckn: Not really, it handles differences between versions and also applies fixes where some of the components is hard to interact with at certain places in Grafana. It could be that other UI elements makes it hard to, via tests, interact with the component you are targeting.

urbiz-grafana: Gotcha, updated the suggestion, LMK if it makes better sense.

urbiz-grafana (dismissed):

urbiz-grafana (approved):

PR #2819, opened 2026-08-05
feat(plugin-e2e): fetch e2e-selectors from Grafana at runtime

sunker:

Adds a runtime path to the `selectors` fixture: it derives the e2e-selectors file URL from the frontend asset path in bootData (so it resolves to the origin in single-binary and the CDN in multi-tenant), fetches the data-only JSON from the Grafana under test, and reconstructs the template descriptors into selectors locally (no eval), falling back to the bundled `@grafana/e2e-selectors` when the file isn't served. This keeps selectors in sync with the Grafana under test without the nightly npm dist-tag.

While we validate the approach, the runtime path is behind a `PLUGIN_E2E_RUNTIME_SELECTORS` toggle. It defaults to off, so consumers keep using the bundled selectors, and it's enabled only in this repo's Playwright workflows so we can dogfood it before making it the default.

Pairs with [the grafana/grafana side](https://github.com/grafana/grafana/pull/127312) that emits the file as a build asset.

Fixes #

Reconstruction only substitutes named {param} placeholders, it never executes fetched content. Unit tests cover `reconstruct` and the fixture (toggle off uses bundled with no fetch, quiet 404 fallback, loud fallback on failures, per-version cache).

### DIFF
--- a/packages/plugin-e2e/src/fixtures/selectors.ts
+++ b/packages/plugin-e2e/src/fixtures/selectors.ts
@@ -1,16 +1,95 @@
-import { TestFixture } from '@playwright/test';
+import { APIRequestContext, TestFixture } from '@playwright/test';
+import {
+  resolveSelectors,
+  versionedComponents as bundledVersionedComponents,
+  versionedPages as bundledVersionedPages,
+} from '@grafana/e2e-selectors';
 import { E2ESelectorGroups, PlaywrightArgs } from '../types';
-import { resolveSelectors, versionedComponents, versionedPages } from '@grafana/e2e-selectors';
 import { versionedConstants } from '../selectors/versionedConstants';
 import { versionedAPIs } from '../selectors/versionedAPIs';
+import { reconstructSelectorTree } from '../selectors/reconstruct';
 
 type SelectorFixture = TestFixture<E2ESelectorGroups, PlaywrightArgs>;
 
-export const selectors: SelectorFixture = async ({ grafanaVersion }, use) => {
-  await use({
-    components: resolveSelectors(versionedComponents, grafanaVersion),
-    pages: resolveSelectors(versionedPages, grafanaVersion),
+type VersionedComponents = typeof bundledVersionedComponents;
+type VersionedPages = typeof bundledVersionedPages;
+
+// data-only selectors file served by the Grafana instance under test (see grafana/grafana
+// e2e-selectors build). Used when the instance serves it; otherwise the fixture falls back to the
+// selectors bundled with the installed release.
+const SELECTORS_URL = '/public/e2e-selectors.json';
+
+// per-worker cache keyed by grafanaVersion so concurrent fixtures share one in-flight fetch
+const selectorsCache = new Map<string, Promise<E2ESelectorGroups>>();
+
+function buildGroups(
+  components: VersionedComponents,
+  pages: VersionedPages,
+  grafanaVersion: string
+): E2ESelectorGroups {
+  return {
+    components: resolveSelectors(components, grafanaVersion),
+    pages: resolveSelectors(pages, grafanaVersion),
     constants: resolveSelectors(versionedConstants, grafanaVersion),
     apis: resolveSelectors(versionedAPIs, grafanaVersion),
-  });
+  };
+}
+
+// fall back to the selectors bundled with the installed @grafana/plugin-e2e release
+function bundledGroups(grafanaVersion: string): E2ESelectorGroups {
+  return buildGroups(bundledVersionedComponents, bundledVersionedPages, grafanaVersion);
+}
+
+async function fetchRuntimeGroups(request: APIRequestContext, grafanaVersion: string): Promise<E2ESelectorGroups> {
+  let response;
+  try {
+    response = await request.get(SELECTORS_URL, { maxRedirects: 0 });
+  } catch (error) {
+    console.warn(`@grafana/plugin-e2e: failed to fetch ${SELECTORS_URL}, falling back to bundled selectors.`, error);
+    return bundledGroups(grafanaVersion);
+  }
+
+  // 404 -> Grafana predates the feature; expected on older images, fall back quietly
+  if (response.status() === 404) {
+    return bundledGroups(grafanaVersion);
+  }
+
+  if (!response.ok()) {
+    console.warn(
+      `@grafana/plugin-e2e: ${SELECTORS_URL} returned ${response.status()}, falling back to bundled selectors.`
+    );
+    return bundledGroups(grafanaVersion);
+  }
+
+  try {
+    const data = JSON.parse(await response.text()) as {
+      schemaVersion?: unknown;
+      versionedComponents?: unknown;
+      versionedPages?: unknown;
+    };
+    if (
+      data?.schemaVersion !== 1 ||
+      typeof data.versionedComponents !== 'object' ||
+      typeof data.versionedPages !== 'object'
+    ) {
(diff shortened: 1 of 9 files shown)


packages/plugin-e2e/src/selectors/reconstruct.ts
```diff
@@ -0,0 +1,57 @@
...
+type TemplateDescriptor = { $template: string | ConditionalTemplate; params?: string[] };
+
+function isDescriptor(node: object): node is TemplateDescriptor {
+  return '$template' in node;
+}
+
+// substitutes each named {param} placeholder with the positional call arg at the same index. braced
+// tokens are unique per name, so `{id}` never partially matches `{idx}` and replacement order is safe.
+function fill(template: string, params: string[], args: string[]): string {
+  return params.reduce((acc, name, index) => acc.split(`{${name}}`).join(args[index] ?? ''), template);
```
mckn, line 18: fill() replaces each {param} one at a time. A replaced value can contain another param's literal token, for example the text '{id}'. The next step then replaces that text again. This can corrupt the selector string. Should we warn or log when this happens?

packages/plugin-e2e/src/selectors/reconstruct.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (typeof template === 'string') {
+    return (...args: string[]) => fill(template, params, args);
+  }
+  if (
+    template &&
+    typeof template === 'object' &&
+    typeof template.whenPresent === 'string' &&
+    typeof template.whenAbsent === 'string'
+  ) {
+    return (...args: string[]) => (args[0] ? fill(template.whenPresent, params, args) : template.whenAbsent);
```
mckn, line 33: This code checks args[0] with a truthy check. An empty string or the number 0 is falsy. The code then treats a present value as absent, and returns the wrong template. Should we check for undefined instead?

packages/plugin-e2e/src/fixtures/selectors.ts
```diff
@@ -1,16 +1,113 @@
...
 
-export const selectors: SelectorFixture = async ({ grafanaVersion }, use) => {
-  await use({
-    components: resolveSelectors(versionedComponents, grafanaVersion),
-    pages: resolveSelectors(versionedPages, grafanaVersion),
+type VersionedComponents = typeof bundledVersionedComponents;
+type VersionedPages = typeof bundledVersionedPages;
+
+// per-worker cache keyed by grafanaVersion so concurrent fixtures share one in-flight fetch
+const selectorsCache = new Map<string, Promise<E2ESelectorGroups>>();
```
mckn, line 18: The cache key is the Grafana version string only. Two different Grafana builds can share one version string. This happens most often with the latest tag. Add the commit hash to the cache key and to the log messages. This makes it easy to trace which Grafana build a test ran against.

packages/plugin-e2e/src/fixtures/selectors.ts
```diff
@@ -1,16 +1,113 @@
...
+  // couldn't derive where the instance serves the file (older Grafana, or assets missing) -> bundled
+  if (!selectorsUrl) {
+    return bundledGroups(grafanaVersion);
+  }
+
+  let response;
+  try {
+    response = await request.get(selectorsUrl, { maxRedirects: 0 });
+  } catch (error) {
+    console.warn(`@grafana/plugin-e2e: failed to fetch ${selectorsUrl}, falling back to bundled selectors.`, error);
```
mckn, line 59: The code falls back to bundled selectors on any fetch or parse error. It logs the fallback with console.warn only. This warning is easy to miss in CI logs. Can we make this warning more visible, for example as a test annotation? Also, should a fetch failure fail the test run instead of a silent fallback?

mckn (commented):
This is a solid, useful improvement.

Do we have access to the commit hash? We could use it for better logging and caching.

Did you consider failing the test run when the selector fetch fails? At minimum, we could fail the run only for Grafana versions that support runtime selectors.

packages/plugin-e2e/src/selectors/reconstruct.ts
sunker: Good catch! FIxed now. End fixed the comment

packages/plugin-e2e/src/selectors/reconstruct.ts
sunker: This is intentional. The source selectors are `(x) => x ? present : absent`, so `''` is the absent branch there too. truthy matches that, undefined would diverge. args are always strings so 0 cant happen. added a comment.

packages/plugin-e2e/src/fixtures/selectors.ts
sunker: The cache is in-memory per worker per run and a run only tests one build, so the version can't collide within a run I think?

packages/plugin-e2e/src/fixtures/selectors.ts
sunker: Fixed now! Non-404 cases now log console.error, 404 stays quiet since thats just older grafana

packages/plugin-e2e/src/fixtures/selectors.ts
```diff
@@ -1,16 +1,113 @@
...
+  }
+
+  try {
+    const data = JSON.parse(await response.text()) as {
+      schemaVersion?: unknown;
+      versionedComponents?: unknown;
+      versionedPages?: unknown;
+    };
+    if (
+      data?.schemaVersion !== 1 ||
```
tolzhabayev, line 86: any reason we only expect it to be version 1?

packages/plugin-e2e/src/selectors/reconstruct.ts
tolzhabayev: what do you expect here to happen?

```
pluginId = "plugin-{pageId}", pageId = "settings"
```

probably would be good to add a test for that. 

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZgSwDZJgXhgCigUwLYAOSAhrgFwzQBOCYA5gDQwHFXF4QXW10DaAuk1Z1OlKDXoCAlF3E90APhgBvAFAwNMKjigBXKmGat2EAHTaAJruA5MmYsGBMw7HE1oWcADymKYD4FMIIgQoTAADZQASZRc8HABfBPCpUwArEFp7KhFeD29+GAB+IpgAcjKpJlxCElwpAG5VBNV1GFBIWAc9YlQMRBRMNo0yi1JiAFpcaAQLGAAFJF06WnbwRDojOhwVImXaAEkLBN3ibaOEsoZhmF4yvZWwI6vylnOLMsEbu4faCeU3jgLi8yhAdFAeBBPm0pK0NB1oDBvAQcMBcHMMKNxlMcDM5ot9oYOhstjtfmB-oCLpRwZCynC1pAQEgcKYkCA6JhlP40bpekxkaj0TAElIgA

packages/plugin-e2e/src/fixtures/selectors.ts
sunker: Yes that's the format grafana supports today, but theoretically there could be more variants in the future and plugin-e2e needs to know version so it can parse it accordingly.

packages/plugin-e2e/src/selectors/reconstruct.ts
sunker: IT replaces every instance of `{param}` in the template at once. In this example, you get: `plugin-pageId}/settings`. I'll add a test for it..

tolzhabayev (approved):

PR #2821, opened 2026-08-06
ci: post release notification to a second slack channel

tolzhabayev:
The release Slack notification only went to one channel. chat.postMessage takes a single channel per call, so the payload step now emits one payload per channel and a second send step posts the same message to C0BNRK96T24.

The bot needs to be invited to the new channel before the next release, otherwise the second post fails with not_in_channel.


sunker (approved):

jackw (approved):
LGTM! 🚀

PR #2828, opened 2026-08-14
chore(deps): patch 18 dependency CVEs via targeted updates

leventebalogh:
Related issue: https://github.com/grafana/grafana-catalog-team/issues/917

### What changed?

Bumped transitive dependencies to their patched versions, clearing **18 of the 19** CVEs VulnO11y reports against `main` - including one critical (`websocket-driver`) and three high. 

- Most were stale lockfile entries that already sat inside their declared ranges, so a targeted `npm update` was enough. 
- Two parent bumps did the rest:
  - `@grafana/data` 13.0.1 -> 13.1.3 relaxes `dompurify` from an exact `3.3.2` pin to `^3.4.0`,
  - `@grafana/faro-web-sdk` 2.1.0 -> 2.9.0 cascades `@opentelemetry/core` to 2.8.0 and drops `protobufjs` from the tree entirely. 
- Dropped the `dompurify` override, as it is already contained in the updated `@grafana/data`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4218,11 +4218,10 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.26.10",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.29.7.tgz",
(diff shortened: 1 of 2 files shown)


andresmgot (approved):

mckn (approved):
LGTM!

PR #2583, opened 2026-04-16
Plugin E2E: Add components fixture with dataSourcePicker and timeRangePicker

mckn:
**What this PR does / why we need it**:

Adds a `components` fixture that exposes eagerly-constructed Grafana UI components for use in tests on arbitrary pages — not just the ones covered by page-specific fixtures like `PanelEditPage` or `ExplorePage`.

Two components are exposed:
- `components.dataSourcePicker` — useful for tests that interact with a data source picker on any page (e.g. [Grafana's exemplars test](https://github.com/grafana/grafana/blob/main/e2e-playwright/various-suite/exemplars.spec.ts#L52))
- `components.timeRangePicker` — useful for tests that set a time range on any page

Both components support a Playwright-style `within(root)` method for scoping to a specific DOM container (mirroring `locator.locator(...)`). The `within()` method is provided by a new shared `ScopedComponent` base class.

```typescript
test('my test', async ({ components, panelEditPage }) => {
  await components.dataSourcePicker.set('gdev-prometheus');
  await components.timeRangePicker.set({ from: 'now-1h', to: 'now' });

  // or scoped to a container
  const panel = panelEditPage.getByGrafanaSelector(selectors.components.PanelEditor.General.content);
  await components.dataSourcePicker.within(panel).set('gdev-tempo');
});
```

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

`ScopedComponent` is exported as part of the public API so plugin authors can subclass it for their own page-level components. Additional components can be added to `Components` following the same pattern.

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install website@5.6.0-canary.2583.25382670310.0
  npm install @grafana/create-plugin@7.4.0-canary.2583.25382670310.0
  npm install @grafana/plugin-e2e@3.8.0-canary.2583.25382670310.0
  # or 
  yarn add website@5.6.0-canary.2583.25382670310.0
  yarn add @grafana/create-plugin@7.4.0-canary.2583.25382670310.0
  yarn add @grafana/plugin-e2e@3.8.0-canary.2583.25382670310.0
  ```
</details>

PR #2832, opened 2026-08-19
fix(plugin-e2e): wait for feature toggles instead of reading them once

adamyeats:

`isLegacyFeatureEnabled` read `window.grafanaBootData.settings.featureToggles` with no guard:

```ts
const featureToggles = await page.evaluate('window.grafanaBootData.settings.featureToggles');
return Boolean(featureToggles[featureToggle]);
```

Grafana can render the app shell before it has merged its boot data. When that happens the map is briefly absent and the second line throws `TypeError: Cannot read properties of undefined (reading '<toggle>')`. `DashboardPage.addPanel` calls this on any Grafana 13 or later to decide `useNewSidebarLayout`, so the `panelEditPage` fixture fails outright and takes the whole test with it.

We hit this on a Grafana Cloud instance, whose shell fetches its settings asynchronously. It does not reproduce against a local Grafana, which inlines the settings into the document, so the failure only appears against a deployed instance.

**Optional chaining alone would be the wrong fix.** A missing map is indistinguishable from every toggle being disabled, so `addPanel` would silently choose its legacy branch on an instance where `dashboardNewLayouts` is actually enabled. That trades a loud crash for a quiet wrong path. This waits for the map instead, and only falls back to an empty one if it never arrives, which is the same degrade-and-log approach the `bootData` fixture already uses.

Worth noting the `Window` declaration in `src/index.ts` types `featureToggles` as always present, which is why there was nothing to prompt a guard here. I have left that type alone rather than making it optional, since it describes what Grafana settles on, and widening it would ripple through unrelated call sites.

No upstream issue. Found downstream while getting a Cloud E2E suite green, catalogued in grafana/grafana-cloudwatch-datasource#619.

- Adds `isFeatureToggleEnabled.test.ts` covering an enabled toggle, a toggle absent from the map, that the map is waited for rather than read once, and that a map which never arrives returns `false` and logs rather than throwing.
- The 5s ceiling is a polling ceiling, so a local Grafana with inline settings resolves immediately and pays nothing.
- `npm run lint`, `npm run typecheck` and `npx vitest run` are all clean in `packages/plugin-e2e`, 48 tests passing.

One adjacent thing I did not change, in case you want it in a separate PR: `DataSourcePicker.set()` fills the combobox and presses Enter without asserting that anything was selected. A name that matches nothing leaves the panel on its previous datasource and the test fails much later on a missing control. That silent no-op cost us several days of misdiagnosis. Happy to send a patch if you would like the assertion added.

### DIFF
--- a/packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
+++ b/packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
@@ -10,10 +10,13 @@ const FEATURE_TOGGLES_TIMEOUT = 5000;
  *
  * Grafana can render the app shell before it has merged its boot data, in which case
  * `settings.featureToggles` is briefly absent even though the Window type declares it as always
- * present. Reading it once then throws `Cannot read properties of undefined`. Falling back to an
- * empty map without waiting is worse than throwing, because a missing map is indistinguishable
- * from every toggle being disabled, so a caller silently takes its legacy branch on an instance
- * where the toggle is enabled.
+ * present. Reading it once then throws `Cannot read properties of undefined`.
+ *
+ * Waiting removes that race. It deliberately does not fall back to an empty map when the wait
+ * times out, because a missing map is indistinguishable from every toggle being disabled: a
+ * caller would silently take its legacy branch on an instance where the toggle is enabled. A
+ * loud failure is the lesser evil, so the timeout is rethrown with the context needed to act
+ * on it.
  */
 const readFeatureToggles = async <T>(page: Page): Promise<T> => {
   try {
@@ -24,8 +27,10 @@ const readFeatureToggles = async <T>(page: Page): Promise<T> => {
     );
     return (await featureToggles.jsonValue()) as T;
   } catch (error) {
-    console.error('@grafana/plugin-e2e: Failed to read feature toggles from boot data', error);
-    return {} as T;
+    throw new Error(
+      `@grafana/plugin-e2e: window.grafanaBootData.settings.featureToggles was not available within ${FEATURE_TOGGLES_TIMEOUT}ms`,
+      { cause: error }
+    );
   }
 };
 
--- a/packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
+++ b/packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
@@ -3,9 +3,35 @@ import { PlaywrightArgs } from '../types';
 
 type FeatureToggleFixture = TestFixture<<T = object>(featureToggle: keyof T) => Promise<boolean>, PlaywrightArgs>;
 
+const FEATURE_TOGGLES_TIMEOUT = 5000;
+
+/**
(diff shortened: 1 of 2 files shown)


adamyeats:
The two red legs are pre-existing, not from this change.

Both fail on `components.timeRangePicker.within … scoped to a root locator` (#2583). `DashboardPage.toolbar` gates on the version each selector first existed, but at 9.4.17 and 11.1.13 the element is still behind a default-off toggle (`topnav`, `dashboardScene`), so the scoped locator waits out the test timeout — the `browser has been closed` error is the teardown cascade, not the cause. Both versions only entered the matrix this week.

This change can't reach that test: `isLegacyFeatureEnabled` is called only from `addPanel`, which that test never calls.

Fixed in #2836, verified 8.5.27 → 13.2.0.

Unrelated and repo-wide: `Test create-plugin update command` fails because `@grafana/data@12.4.9` pins the unpublished `@grafana/schema@12.4.9`.

packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
```diff
@@ -3,17 +3,43 @@ import { PlaywrightArgs } from '../types';
...
+  try {
+    const featureToggles = await page.waitForFunction(
+      () => window.grafanaBootData?.settings?.featureToggles ?? null,
+      undefined,
+      { timeout: FEATURE_TOGGLES_TIMEOUT }
+    );
+    return (await featureToggles.jsonValue()) as T;
+  } catch (error) {
+    console.error('@grafana/plugin-e2e: Failed to read feature toggles from boot data', error);
+    return {} as T;
```
sunker: Since the tsdoc above argues a missing map is indistinguishable from a disabled toggle, I think we should re-throw here instead of falling back to {}.

sunker (commented):
Thanks for raising this! Added a minor comment.

The create-plugin update step should be [fixed here](https://github.com/grafana/plugin-tools/pull/2837), although you'll need to sync with main to pick up the fix. I suspect the other two failures may be intermittent and are something for me and the team to investigate. Let me know if you're still seeing them.

packages/plugin-e2e/src/fixtures/isFeatureToggleEnabled.ts
adamyeats: The fallback contradicted the reasoning directly above it. I removed the empty map. The function now rethrows the timeout, and the message names the path and the 5000ms budget. The original error stays available as `cause`. I also removed the `console.error` call, because it only existed to explain the fallback. The tests now assert a rejection instead of a false result.

adamyeats:
The rethrow is in, and the branch now includes main for #2837. Thanks for sorting that.

The two version legs are not intermittent. `DashboardPage.toolbar` picks its root by the version where each selector first existed. At 9.4.x, and from 11.1.x to 11.2.x, that element sits behind a default-off toggle, so Grafana never renders it. #2836 corrects it and is green on all six legs.

codyconfer (approved):

sunker (approved):
Very nice! 🚀

PR #2836, opened 2026-08-19
fix(plugin-e2e): correct dashboard toolbar version thresholds

adamyeats:

`DashboardPage.toolbar` maps a Grafana version to the toolbar that holds the time range controls.
Each threshold was the version at which the selector first existed, per `@grafana/e2e-selectors`.
At each of those versions the element is still behind a default-off feature toggle, so the getter
returns a locator for an element Grafana never renders:

| Old threshold | Selector | Belongs to | Default-on from |
| --- | --- | --- | --- |
| 11.1.0 | `Dashboard.Controls` | scenes dashboard (`dashboardScene`) | 11.3.0 |
| 9.4.0 | `NavToolbar.container` | top nav (`topnav`) | 9.5.0 |

Anything scoping to `toolbar` then waits for a root that never appears. In `TimeRange.set` the
scoped click consumes the whole test budget, so its `catch` only runs during teardown and reports
`locator.click: Target page, context or browser has been closed` — which hides the real error,
`Test timeout of 30000ms exceeded`.

This moves both thresholds to the version at which each toolbar became the default. 9.5.0 is
already the threshold `addPanel` uses for the same top nav change, two branches down the same file.

None filed — found while investigating the red `9.4.17` and `11.1.13` legs on #2832.

These versions reached CI for the first time this week. `playwright.yml` runs the full matrix only
on PRs that touch `packages/plugin-e2e/**`, and the resolver picks the latest patch of each minor
when it runs, so the set moves as Grafana publishes releases. The last full-matrix run before this
week resolved 9.3.16 and 11.0.11, which sit either side of both broken ranges. The nightly uses
`limit: 1`, so it only covers the latest stable and grafana-dev. Every plugin-e2e PR is affected
until this lands, including #2819.

Behaviour changes only for 9.4.x and 11.1.x–11.2.x, which are exactly the broken ranges. Every
other version resolves to the same selector as before. Verified against each version with the
compose environment from `playwright.yml`:

| Grafana | Resolves to | Before | After |
| --- | --- | --- | --- |
| 8.5.27 | `.page-toolbar` | pass | pass |
| 9.4.17 | `.page-toolbar` | fail | pass |
| 11.0.11 | `Nav toolbar` | pass | pass |
| 11.1.13 | `Nav toolbar` | fail | pass |
| 11.2.10 | `Nav toolbar` | fail | pass |
| 13.2.0 | `dashboard controls` | pass | pass |

`DashboardPage.test.ts` locks all three branches and fails on 9.4.17, 11.1.13 and 11.2.10 with the
old thresholds.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
           version-resolver-type: plugin-grafana-dependency
           grafana-dependency: '>=8.5.0'
           skip-grafana-dev-image: false
-          limit: 0 # TEMPORARY (do not merge): verifying the toolbar thresholds against every supported version.
+          # limit: 0 # Uncomment to test all versions since 8.5.0. Useful when testing compatibility for new APIs.
 
   playwright-tests:
     needs: resolve-versions
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -28,7 +28,7 @@ jobs:
           version-resolver-type: plugin-grafana-dependency
           grafana-dependency: '>=8.5.0'
           skip-grafana-dev-image: false
-          # limit: 0 # Uncomment to test all versions since 8.5.0. Useful when testing compatibility for new APIs.
+          limit: 0 # TEMPORARY (do not merge): verifying the toolbar thresholds against every supported version.
 
   playwright-tests:
     needs: resolve-versions
--- a/packages/plugin-e2e/src/models/pages/DashboardPage.test.ts
+++ b/packages/plugin-e2e/src/models/pages/DashboardPage.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect, vi } from 'vitest';
+import { DashboardPage } from './DashboardPage';
+import { PluginTestCtx } from '../../types';
+
+const NAV_TOOLBAR = 'data-testid Nav toolbar';
+const DASHBOARD_CONTROLS = 'data-testid dashboard controls';
+
+/**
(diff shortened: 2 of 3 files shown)


sunker (commented):
Thanks for raising this PR @adamyeats! What we usually do in situations like these is to temporarily uncomment [this line](https://github.com/grafana/plugin-tools/blob/65a18e11abaf1e27885dc729a4651ad0f631cb41/.github/workflows/playwright.yml#L31) to run the internal plugin-e2e tests in all supported version of Grafana. Would you mind doing that and verify all tests still pass?

adamyeats:
`limit: 0` is on in the last commit, so this run covers every version from 8.5.0. I will report the result here and revert the line before merge.

codyconfer (dismissed):

adamyeats:
All 28 versions ran. 24 passed. Four failed: 9.5.21, 10.0.13, 10.1.10 and 11.5.10.

None of the four failed on the toolbar. I checked every errored action in their traces, and each one stalls on panel options groups, alert rule steps, query editor rows or panel menus. No errored action uses `Nav toolbar`, `dashboard controls` or `.page-toolbar`. The three versions this PR changes behaviour for (9.4.17, 11.1.13 and 11.2.10) all passed.

The four failures look pre-existing on versions the matrix has never included. I can raise a separate issue with the details if that is useful.

`limit: 0` is reverted, so the diff is back to the two source files.

sunker (approved):
Nice 🚀

PR #2840, opened 2026-08-20
fix(plugin-docs-cli): mask inline code before HTML checks

sunker:

Angle-bracket placeholder text inside inline code (e.g. `<slug>` in a URL example) was being flagged as a raw HTML tag, since the no-raw-html check only skipped fenced code blocks, not inline backtick spans. This PR adds a `maskInlineCode()` helper that neutralizes inline code spans before the HTML-tag regex runs.

Also adds a new `unfilled-section-brief` validation rule that fails strict validation if a page still contains a scaffolded `<!-- section-brief:start -->` marker, so an unfilled documentation stub can never ship silently.

Fixes #

Both issues were found while manually testing the `panel-docs` codemod (grafana/plugin-tools#2834) against a real plugin.

### DIFF
--- a/packages/plugin-docs-cli/src/validation/rules/stub-content.ts
+++ b/packages/plugin-docs-cli/src/validation/rules/stub-content.ts
@@ -1,0 +1,64 @@
+import { readFile, readdir } from 'node:fs/promises';
+import type { Dirent } from 'node:fs';
+import { join, relative } from 'node:path';
+import { type Diagnostic, type ValidationInput, Rule } from '../types.js';
+import { isMetaFile } from './utils.js';
+
+// matches the opening marker of a section-brief authoring-guidance block,
+// scaffolded by `create-plugin add panel-docs`/`datasource-docs` as a
+// placeholder for the author to replace with real content.
+const SECTION_BRIEF_START_RE = /<!--\s*section-brief:start\s*-->/;
+
+/**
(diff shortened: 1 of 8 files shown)


packages/plugin-docs-cli/src/validation/rules/stub-content.ts
```diff
@@ -0,0 +1,64 @@
...
+    const relativePath = relative(input.docsPath, absolutePath);
+    let raw: string;
+    try {
+      raw = await readFile(absolutePath, 'utf-8');
+    } catch {
+      continue;
+    }
+
+    const lines = raw.split('\n');
+    for (let i = 0; i < lines.length; i++) {
```
mckn, line 47: Is it any value in running these lines in the `getCodeBlockLines` or `maskInlineCode`? I'm asking since the other rules seems to use those functions to check for code.

packages/plugin-docs-cli/src/validation/rules/stub-content.ts
```diff
@@ -0,0 +1,64 @@
...
+      if (!SECTION_BRIEF_START_RE.test(lines[i])) {
```
mckn, line 48: Do we also need to verify that there is an end block? Or will this fail if it is missing?

mckn (approved):
LGTM! Left two nits/questions

PR #2843, opened 2026-08-21
feat(create-plugin): add AI authoring assistance to panel-docs codemod

sunker:

Adds AI authoring assistance to the `panel-docs` codemod: a conventions file the coding agent reads automatically, plus one `bootstrap-plugin-docs` skill for the initial fill.

The shape changed substantially from the first draft of this PR (four slash-command skills behind a required `--agentLoop` flag), after looking at what comparable tooling actually ships. Three findings drove it:

1. **Conventions belong in always-loaded context, not in skills.** Vercel's [published evals](https://vercel.com/blog/agents-md-outperforms-skills-in-our-agent-evals) found an auto-invoked skill scored no better than no docs at all (53% vs 53% baseline), because the skill was never invoked in 56% of cases, while always-loaded context scored 100%. Next.js states the same split as policy: context for lookups, skills for workflows.
2. **Nobody ships per-agent skill fan-out.** Of the tools surveyed (Nx, Next.js, Astro, Expo, React Router, Prisma, Supabase, Stripe, shadcn, Angular, Svelte), none hand-write parallel per-agent skill directories, and none ask which agent at scaffold time except Angular.
3. **This repo already has the right pattern.** `templates/common/` scaffolds root `AGENTS.md`/`CLAUDE.md`/`GEMINI.md` pointers into `.config/AGENTS/`, with topic files like `e2e-testing.md`. This PR follows that rather than inventing a parallel convention.

**Changes**

- `write-plugin-docs`, `review-plugin-docs` and `validate-plugin-docs` are gone. The first two were conventions, the third wrapped `npm run docs:validate`. All three fold into the authoring guide.
- `bootstrap-plugin-docs` stays: it is a genuine multi-step workflow.
- The guide lands at `.config/AGENTS/plugin-docs.md`, next to `e2e-testing.md`, reached by a pointer line appended to `.config/AGENTS/instructions.md`.
- `--agentLoop` is replaced by `--no-agents` (defaults on).

**Two things worth reviewing closely**

*The guide cannot live inside `docsPath`.* It was at `docs/AGENTS.md`. Anything inside `docsPath` is validated as a publishable page, so `npm run docs:validate` failed on it with 8 errors. Moving it out fixes that, and the `instructions.md` pointer is also what makes it reachable from Claude Code, which never reads `AGENTS.md`.

*The skill is written whole to both `.claude/skills/` and `.agents/skills/`, not shimmed.* The one-line `@import` shim pattern used by the existing `build-plugin` and `validate-plugin` skills does not work in Claude Code: [the docs](https://code.claude.com/docs/en/skills) confirm `@imports` are unsupported in `SKILL.md`, and a `SKILL.md` whose first line isn't `---` is treated as literal content, so such a skill registers no description and cannot be invoked. Duplication is deliberate: `Context` writes string content so symlinking isn't available, and these are committed team-shared files, so detecting the scaffolder's own agent would give their teammates the wrong answer. `.agents/` covers Codex, Cursor, Copilot, Gemini CLI and Amp, so there is no separate `.codex/` copy.

N/A

Stacked on #2834, which must merge first.

Verified end to end against a scratch panel plugin: both skill copies are byte-identical and spec-valid (frontmatter present, `name` matches directory), every `@`-import in the instructions chain resolves, no `{{placeholders}}` leak, `--no-agents` writes the docs stubs and nothing else, and `plugin-docs-cli validate --strict` reports `✓ Documentation is valid`.

Separately: the existing `build-plugin` and `validate-plugin` skills in `templates/common/` are broken in Claude Code for the same reason described above. Out of scope here since fixing them changes what every new plugin gets.

### DIFF
--- a/packages/create-plugin/src/codemods/additions/docs-scaffolding.test.ts
+++ b/packages/create-plugin/src/codemods/additions/docs-scaffolding.test.ts
@@ -28,7 +28,7 @@ function makeContext(pluginJson: Record<string, unknown> = { type: 'panel', name
 describe('docs-scaffolding', () => {
   const tempDirs: string[] = [];
 
-  // build a synthetic plugin-type template folder with the given docs template file contents
+  // build a synthetic templateBaseUrl folder with the given docs template file contents
   function makeTemplateBaseUrl(files: Record<string, string>): URL {
     const dir = mkdtempSync(join(tmpdir(), 'panel-docs-templates-'));
     tempDirs.push(dir);
@@ -63,7 +63,10 @@ describe('docs-scaffolding', () => {
     }
   });
 
-  function call(context: Context, overrides: { templates?: Record<string, string>; docsPath?: string } = {}): void {
+  function call(
+    context: Context,
+    overrides: { templates?: Record<string, string>; docsPath?: string; agents?: boolean } = {}
+  ): void {
     const templates = overrides.templates ?? { 'index.md': '# {{pluginName}}\n' };
     setupDocsScaffolding({
       context,
@@ -71,6 +74,7 @@ describe('docs-scaffolding', () => {
       templateBaseUrl: makeTemplateBaseUrl(templates),
       commonTemplateBaseUrl: makeCommonTemplateBaseUrl(),
       codemodName: 'panel-docs',
+      agents: overrides.agents ?? false,
     });
   }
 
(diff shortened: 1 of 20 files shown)


sunker:
closing in favour of https://github.com/grafana/plugin-tools/pull/2834

PR #2845, opened 2026-08-21
fix(plugin-docs-cli): treat AGENTS.md as a repo-meta file

sunker:

Adds `AGENTS.md` to the meta-file allowlist so it's excluded from docs validation, matching `README.md` and friends.

Fixes #

Found via a live docs:serve run against a real plugin.


mckn (approved):
LGTM!

PR #2847, opened 2026-08-21
fix(create-plugin): add @react-hookz/web + @ver0/deep-equal to Jest ESM ignore

MattIPv4:

`@grafana/ui@13.2.0` includes two new ESM-only modules, causing the scaffolded Jest setup to fail for plugins that update to it.


jackw (approved):
LGTM! 🚀

PR #2848, opened 2026-08-21
fix(plugin-e2e): support new root OFREP route

MattIPv4:

Grafana is switching to using a namespace-less OFREP route at the root path, which the current E2E logic does not catch to override responses.

### DIFF
--- a/packages/plugin-e2e/src/fixtures/getOpenFeatureFlag.ts
+++ b/packages/plugin-e2e/src/fixtures/getOpenFeatureFlag.ts
@@ -9,27 +9,42 @@ export const getBooleanOpenFeatureFlag: GetBooleanOpenFeatureFlagFixture = async
 ) => {
   await use(async (flagKey: string) => {
     try {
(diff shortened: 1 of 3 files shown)


sunker (commented):
Nice. Is the namespace less route tied to a specific Grafana version, or can it show up on stacks reporting the same version as the old route? If it's version specific, might be cleaner to add it as a new versioned key in ofrepBulkPath/ofrepSinglePath (like DataSource.health already does) instead of the dual URL fallback in the fixture.

MattIPv4:
> Is the namespace less route tied to a specific Grafana version, or can it show up on stacks reporting the same version as the old route?

It is behind a feature flag (https://github.com/grafana/grafana/pull/130255 + https://github.com/grafana/grafana/pull/130254), so no specific version as far as I know.

sunker:
Think the CI failures will be solved if you sync with main.

sunker (approved):

Issue #2842, opened 2026-08-21
Feat: Experimental support for Kinds via grafana-app-sdk


stevesg:
### Which areas does this feature request relate to

- [x] Create Plugin
- [ ] Sign Plugin
- [ ] Plugin E2E
- [ ] Plugin Meta Extractor
- [ ] Documentation

### Problem

Support is being added to Grafana for processing "app manifest" embedded in plugins, generated by `grafana-app-sdk`. It would be valuable to have support in `create-plugin` to setup the scaffolding to enhance the user experience.

### Solution

Whilst this is considered experimental and not for general consumption, we can hide the extra scaffolding:

```
create-plugin add experimental-app-sdk
```

The extra scaffolding would include:
1. A directory for storing the `.cue` files, and some skeleton files:
  `/kinds/...`
2. A command to run the app-sdk code generator
  `npm run generate:kinds` 
3. An extra hint in `AGENTS/instructions.md` about app-sdk/kinds and how/when to run code generation
4. For backend plugins, some Go code modifications

### Alternatives

- Variation on this is to add it to Go codegen/Magefile instead of npm (= e.g. mage `generate-kinds`), but frontend-only apps that define kinds are feasible and sensible.
- Creating or adding the functionality an external tool is another option (e.g. the app-sdk itself)

### Task List

- [x] Support for manifest generation from .cue files (https://github.com/grafana/plugin-tools/pull/2853) 
- [x] Support scaffolding backends too to support validation/mutation/conversion/routes (https://github.com/grafana/plugin-tools/pull/2856)

Improvements:
- [ ] Think about ways for plugins to use generated API clients for custom resources
- [x] Catch drift between the cue kinds and the generated types/manifest (CI? Build step?)
  https://github.com/grafana/plugin-tools/pull/2862
- [ ] Warns if a plugin ships a manifest while its grafanaDependency includes versions that can't serve it
- [ ] Documented best-practice on how to manage versioning
- [ ] Iterate on AGENTS.md for how to use various SDK features
- [x] Print "Next steps" at end
- [ ] Simplify injecting "sdkplugin.Run" by extending "plugin.Manage" instead
- [x] Automate `go mod tidy`

stevesg:
Whilst testing, I realized it would be nice to print "Next steps" after the success message, like when using `create-plugin` regular. In my testing this has meant I've forgotten to run `npm generate:kinds` multiple times. I think for this we'd need an ability set a "success message" on the `Context`.

<details>
[...]

PR #2853, opened 2026-08-24
feat(create-plugin): add grafana-app-sdk CUE kind code generation

stevesg:

Adds opt-in support for `grafana-app-sdk`:

```
$ create-plugin add experimental-app-sdk
```

Code generation runs through `npm run generate:kinds`, a convenience script to download the grafana-app-sdk binary and run it.

This change covers frontend-only app plugins: `kinds/config.cue` sets `codegen: goEnabled: false`, so no Go code, go.mod dependency, or backend is involved. Validation/mutation/conversion and custom routes are not supported.

Summary of changes:

- templates/app-sdk/: CUE kinds, the generate script, and a README
- kinds/config.cue writes the manifest straight into src/app-sdk-manifest.json, so the frontend build's existing **/*.json copy pattern picks it up with no dedicated copy step
- docker-compose.yaml: enable the plugins.appSDKManifest feature toggle
- AGENTS/: app-sdk.md and edit to instructions.md to point at it

Part of #2842


packages/create-plugin/templates/common/docker-compose.yaml
```diff
@@ -3,3 +3,10 @@ services:
     extends:
       file: .config/docker-compose-base.yaml
       service: grafana
+{{#if experimentalAppSdk}}
+    environment:
+      # Lets Grafana read the grafana-app-sdk manifest from the plugin bundle and serve your kinds
+      # through the aggregated API server. Add further toggles to this comma-separated list.
+      # Note: this replaces (does not merge with) any GF_FEATURE_TOGGLES_ENABLE set in the base file.
+      GF_FEATURE_TOGGLES_ENABLE: plugins.appSDKManifest
```
sunker: Are you sure this toggle works? I couldn't find `plugins.appSDKManifest` anywhere in Grafana - on main I only see loadAppManifest and registerAPIServer.

packages/create-plugin/src/constants.ts
```diff
@@ -52,8 +53,23 @@ export const DEFAULT_FEATURE_FLAGS = {
...
+//
+// A release tag (vX.Y.Z) is downloaded as a checksum-verified binary. Any other git ref — a branch
+// such as `main`, or a commit SHA — is built with `go install` instead; see
+// templates/app-sdk/scripts/generate-kinds.mjs.
+//
+// TEMPORARY: pinned to an unreleased commit for `codegen: goEnabled: false` (grafana-app-sdk#1492),
+// which lets frontend-only plugins generate TypeScript without emitting Go. A pseudo-version rather
+// than `main` keeps it reproducible, at the cost of building the CLI from source via `go install`.
+// Move this back to a release tag as soon as one contains #1492.
+export const APP_SDK_VERSION = 'v0.57.2-0.20260824082014-08ce79806923';
```
sunker: I assume there will be an actual release we can point to eventually?

packages/create-plugin/templates/app-sdk/kinds/config.cue
```diff
@@ -0,0 +1,33 @@
...
+	codegen: {
+		// This plugin has no Go backend, so skip Go code generation entirely: only TypeScript and
+		// the definitions below are emitted, and no Go toolchain is needed to generate them.
+		goEnabled: false
+		// Generated TypeScript types land in the frontend source dir.
+		tsGenPath:                      "src/generated/"
+		enableK8sPostProcessing:        false
+		enableOperatorStatusGeneration: false
+	}
+}
```
sunker, line 33: One thing my editor picked up here was that it looks like there's a mixture of tabs and spaces. The manifestVersion line and the `// DO NOT EDIT` comment seem to use spaces while everything around them is tabs. Cab cue fmt fixthis?

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,226 @@
+import { fileURLToPath } from 'node:url';
+import { parseDocument, stringify, YAMLMap, Scalar } from 'yaml';
+import type { Context } from '../../context.js';
+import { output } from '../../../utils/utils.console.js';
+import { additionsDebug, renderTemplate } from '../../utils.js';
+
+// Grafana reads an app-sdk manifest from the plugin bundle only when this toggle is enabled.
+const APP_SDK_FEATURE_TOGGLE = 'plugins.appSDKManifest';
```
sunker: Are you sure this toggle works? I couldn't find plugins.appSDKManifest in Grafana, but maybe I'm missing something!

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * Points the plugin's agent instructions at the app-sdk ones.
+ *
+ * On a fresh scaffold this line is rendered into instructions.md by the template. Retrofitted plugins
+ * have an instructions.md that predates the flag, so add it here — without it, an agent reading the
+ * plugin's instructions has no idea kinds or generated code exist.
+ */
+function referenceAgentInstructions(context: Context) {
+  const path = '.config/AGENTS/instructions.md';
```
sunker, line 116: beware that this file was added recently, so all internal plugins trying this out using the add cmd probably don't have it. you may need to be a bit more defensive here and copy if over if it doesn't exist if it's crucial for the app sdk.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -0,0 +1,226 @@
...
+function printNextSteps() {
+  output.log({
+    title: 'Added grafana-app-sdk code generation. Next steps:',
+    body: [
+      'Edit your kinds in ./kinds (start with kinds/example.cue), then run:',
+      '  npm run generate:kinds',
+      'See ./kinds/README.md for the full workflow.',
+    ],
+  });
+}
```
sunker, line 238: Should we add a .gitignore entry for .cache/? When I tried this add cmd on my local machine, the cached CLI is about 90MB and it's not ignored. and since the dir is version-keyed, bumping the version adds another copy rather than replacing.

packages/create-plugin/templates/common/_package.json
```diff
@@ -9,15 +9,16 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint --cache .",
     "lint:fix": "{{ packageManagerName }} run lint{{#if isNPM}} --{{/if}} --fix && prettier --write --list-different .",
-    "e2e": "playwright test",
+    "e2e": "playwright test",{{#if experimentalAppSdk}}
+    "generate:kinds": "node ./scripts/generate-kinds.mjs",{{/if}}
```
sunker: If we go add-only for now, I think these template conditionals could all go.

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,226 @@
...
+
+function addTemplateFiles(context: Context) {
+  for (const file of TEMPLATE_FILES) {
+    if (context.doesFileExist(file)) {
+      additionsDebug(`${file} already exists. Skipping.`);
+      continue;
+    }
+
+    // includeWarning is false: unlike the tool-managed files in .config, these are meant to be
+    // edited — declaring your own kinds is the point.
```
sunker: This is true for the `kinds/*.cue` files, but two of these aren't really meant to be edited I think? generate-kinds.mjs is tool logic, and `config.cue` even says DO NOT EDIT inside it. 

In `create-plugin` the .`config` folder is for files managed by us, so we can change them in migrations later without being too cautious about it. For that reason, I think we should put generate-kinds.mjs there. It also gets the banner and ends up in the lint pass.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * Adds grafana-app-sdk CUE kind code generation to an existing app plugin.
+ *
+ * Every step is guarded so re-running is a no-op, and user edits to the scaffolded CUE are never
+ * overwritten.
+ */
+export default function appSdk(context: Context): Context {
+  if (!isAppPlugin(context)) {
+    return context;
```
sunker, line 34: Is this working in all versions of Grafana? If only supported in certain versions, you need to check `grafanaDependency` in plugin.json and take necessary actions if the plugin supports version of Grafana that dot **not** support app platform APIs.

sunker (commented):
This looks really promising. I really like the simplicity of this. Feels like it aligns really well with the goal defined in the product DNA. Also like the fact that app sdk support is optionally additional and that it can live alongside existing APIs. Added some minor inline comments. Aside from that, I have the following high-level feedback:

Since the `add` command mechnism was introduced we're moving away from feature toggles in the root command and try to enable "optional capabilities" via the add cmd instead. So I'd drop --feature-flags=experimentalAppSdk and only push on add experimental-app-sdk for now. Once these APIs for app sdk have proven stable for internal users and we have a clear path for community plugins, we can flip to enabling app sdk by default for new scaffolded plugins too. Since the short term goal is to get internal devs started quickly, the add script is what we should focus on IMO.

Just note that internal plugins that have been around for a while may have a different file structure etc, so worth testing your add script against a few real ones before announcing internally. I tried it on a 7.6.0 app plugin and hit a couple of smaller issues things (see inlinbe comments). Also, while this is being actively worked on you may not even want to merge the add cmd yet, simply pointing people at the pre-release is probably easier:
```
npx https://pkg.pr.new/grafana/plugin-tools/@grafana/create-plugin@2d6533a add experimental-app-sdk
```

IIUC there's nothing protecting us from drift between the CUE and the generated code right now. Easy to edit a kind, forget to regenerate and commit and typecheck won't catch it since the stale types are still valid ts. Not necessarily something that needs to be addressed in this PR, but would be good to scaffold a workflow that protects against this at some point.

stevesg:
Thanks for the valuable feedback, appreciate it.

I've removed the feature-flag version which cleaned up a lot of code, will address the other points too shortly.

packages/create-plugin/src/constants.ts
stevesg: Yep 100% - it wasn't available when developing initially. I've pushed the change and cleaned up the surrounding code.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
stevesg: Good shout - I instead moved the download to node_modules, just to avoid having editing .gitignore on the `create-plugin add `.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
stevesg: That's a nice idea, didn't think of that. Though the work is still under development in so we don't know the exact version number yet.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
stevesg: It's more a nice-to-have than crucial, it's fine to skip I think to keep things simple.

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
stevesg: Good spot - yeah this toggle is on a branch we've been testing with, in main we actually need [this](https://github.com/grafana/grafana/blob/b3e12529a403348f061071fc534436ad5587b651/pkg/services/featuremgmt/registry.go#L2470) and [this](https://github.com/grafana/grafana/blob/b3e12529a403348f061071fc534436ad5587b651/pkg/services/featuremgmt/registry.go#L2480).

I'll update the PR to reflect this.

stevesg:
Addressed most of the feedback - will do a round of manual testing on some existing plugins.

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * Every step is guarded so re-running is a no-op, and user edits to the scaffolded CUE are never
+ * overwritten.
+ */
+export default function appSdk(context: Context): Context {
+  if (!isAppPlugin(context)) {
+    return context;
+  }
+
+  const changesBefore = Object.keys(context.listChanges()).length;
```
jackw: As this is an addition won't this always be 0?

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  addTemplateFiles(context);
+  referenceAgentInstructions(context);
+  addGenerateScript(context);
+  addFeatureToggle(context);
+
+  // Only guide the user when we actually scaffolded something; a re-run should stay quiet.
+  if (Object.keys(context.listChanges()).length > changesBefore) {
```
jackw: Given above message I think `context.listChanges()).length` is all that's required here.

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+    printNextSteps();
```
jackw: I keep coming back round to the idea of returning a msg along with the context as there's been multiple times now it could be useful.

packages/create-plugin/src/codemods/additions/scripts/app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+    skip('Could not find src/plugin.json.', ['Run this from the root of your plugin.']);
+    return false;
+  }
+
+  let pluginJson;
+  try {
+    pluginJson = JSON.parse(pluginJsonContent);
+  } catch (error) {
+    additionsDebug(`Failed to parse src/plugin.json: ${error}`);
+    skip('Could not parse src/plugin.json.');
```
jackw: nit: do we need both? Generally we leave codemod messaging in debug logs that plugin devs can enable with `DEBUG="create-plugin:additions" npx @grafana/create-plugin add <codemod name>` to understand why something didn't run but either/or could work here.

packages/create-plugin/templates/app-sdk/scripts/generate-kinds.mjs
```diff
@@ -0,0 +1,138 @@
...
+const BIN_OVERRIDE = 'GRAFANA_APP_SDK_BIN';
+
+// node_modules is already gitignored, so nothing extra needs ignoring for this cache.
+const CACHE_DIR = resolve('node_modules', '.cache', 'grafana-app-sdk', VERSION);
+
+function run(command, args, options = {}) {
+  return spawnSync(command, args, { stdio: 'inherit', ...options });
+}
+
+const EXE = platform() === 'win32' ? '.exe' : '';
```
jackw: We don't support windows outside of WSL: https://grafana-dev.com/developers/plugin-tools/#supported-operating-systems

packages/create-plugin/src/codemods/additions/additions.ts
```diff
@@ -11,4 +11,9 @@ export default [
     description: 'Externalizes the react JSX runtime to help migrate plugins to React 19',
     scriptPath: import.meta.resolve('./scripts/externalize-jsx-runtime.js'),
   },
+  {
+    name: 'experimental-app-sdk',
+    description: 'Adds grafana-app-sdk CUE kind code generation to an app plugin',
+    scriptPath: import.meta.resolve('./scripts/app-sdk.js'),
```
sunker: Nit: the other additions have the filename match the registered name. Probably good to stick to that pattern here as well.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.test.ts
```diff
@@ -0,0 +1,329 @@
...
+  pluginType = 'app',
+  compose = STOCK_COMPOSE,
+}: { pluginType?: string; compose?: string | null } = {}) {
+  const context = new Context('/virtual');
+
+  context.addFile('src/plugin.json', JSON.stringify({ type: pluginType, id: 'my-plugin-id', backend: false }));
+  context.addFile('package.json', JSON.stringify({ scripts: { build: 'webpack' } }, null, 2));
+  context.addFile('.gitignore', 'node_modules/\ndist/\n');
+  context.addFile('.config/bundler/copyFiles.ts', `export const copyFilePatterns = ['**/*.json'];`);
+  context.addFile('.config/AGENTS/instructions.md', '# Grafana Plugin\n\n## Critical rules\n\n- Existing rule.\n');
```
sunker: The fixture always creates `.config/AGENTS/instructions.md`, so we never test the case where it's missing. Tried the script on a 7.6.0 plugin and it didnt have that file, so you end up with an app-sdk.md nothing points at.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+function addTemplateFiles(context: Context) {
+  for (const file of TEMPLATE_FILES) {
+    if (context.doesFileExist(file)) {
+      additionsDebug(`${file} already exists. Skipping.`);
+      continue;
+    }
+
+    // includeWarning is false: unlike the tool-managed files in .config, these are meant to be
+    // edited — declaring your own kinds is the point.
+    context.addFile(file, renderTemplate(templatePath(file), false));
```
sunker, line 100: FYI the `.config` folder in the plugin is "owned by us", meaning we can modify these files with [migrations](https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/src/codemods/migrations/migrations.ts) that run as part of plugin ci. I'd put the `generate-kinds.mjs` script in the .config folder as they codegen logic may need to be finetuned in the future.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -0,0 +1,229 @@
...
+  context.updateFile(composePath, stringify(composeData, { lineWidth: 120, singleQuote: true }));
+}
+
+/** Tells the user what to run next. */
+function printNextSteps() {
+  output.log({
+    title: 'Added grafana-app-sdk code generation. Next steps:',
+    body: [
+      'Edit your kinds in ./kinds (start with kinds/example.cue), then run:',
+      '  npm run generate:kinds',
```
sunker: Instead of hardcoding package manager, use `{{ packageManagerName }}`

sunker (commented):
Added a few nits, but overall looks good! 

A few things to consider for the future: 
* As Marcus suggested, we should think about ways for plugins to use generated API clients for custom resources
*  Would be good to have a ci workflow to catch drift between the cue kinds and the generated types/manifest
* Nothing currently warns if a plugin ships a manifest while its `grafanaDependency` includes versions that can't serve it - kinds just silently don't exist. Bumping grafanaDependency is one option but it gates the whole plugin so also not ideal
* Would be good to have a documented best-practice on how to manage versioning

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.test.ts
stevesg: Changed to skip the file if `instructions.md` doesn't exist.

stevesg:
Thanks again @sunker !

Agree with all the follow on suggestions - I'll track them in the issue (https://github.com/grafana/plugin-tools/issues/2842)

sunker (approved):
Nice 🚀

stevesg:
Tested end-to-end with https://github.com/grafana/grafana/pull/132037:

```bash
~/plugins$ npx https://pkg.pr.new/grafana/plugin-tools/@grafana/create-plugin@cd2a3b6
~/plugins$ cd test-testing-app/
~/plugins/test-testing-app$ npm install
~/plugins/test-testing-app$ git init && git add . && git commit -m "wip"
~/plugins/test-testing-app$ npx https://pkg.pr.new/grafana/plugin-tools/@grafana/create-plugin@cd2a3b6 add experimental-app-sdk
~/plugins/test-testing-app$ npm run generate:kinds
~/plugins/test-testing-app$ npm run build
~/plugins/test-testing-app$ GRAFANA_VERSION=dev docker compose up -d
```

```
$ curl http://localhost:3000/apis | jq '.groups[] | select(.name | contains("testtestingapp"))'
{
  "name": "testtestingapp.ext.grafana.app",
  "versions": [
    {
      "groupVersion": "testtestingapp.ext.grafana.app/v1alpha1",
      "version": "v1alpha1"
    },
    {
      "groupVersion": "testtestingapp.ext.grafana.app/v0alpha1",
      "version": "v0alpha1"
    }
  ],
  "preferredVersion": {
    "groupVersion": "testtestingapp.ext.grafana.app/v1alpha1",
    "version": "v1alpha1"
  },
  "serverAddressByClientCIDRs": [
    {
      "clientCIDR": "0.0.0.0/0",
      "serverAddress": "0.0.0.0:6443"
    }
  ]
}
```

PR #2854, opened 2026-08-26
feat(plugin-e2e): expose response body from waitForQueryDataResponse

adamyeats:

`GrafanaPage.waitForQueryDataResponse` only ever hands the caller a raw Playwright `Response` — its `cb` predicate is `(request: Response) => boolean | Promise<boolean>`, with no way to filter or assert on the response body without a separate `.json()` call. Reading the body after the response has resolved is unreliable in practice (the CDP buffer isn't guaranteed to still be live), so plugin authors end up re-implementing "parse the body inside the predicate, stash it in a closure" by hand — that's exactly what `grafana/clickhouse-datasource` did, and where this suggestion came from (review feedback on [clickhouse-datasource#2122](https://github.com/grafana/clickhouse-datasource/pull/2122), no existing upstream issue to link).

This PR adds a new, purely additive `waitForQueryDataResponseWithBody` method alongside the existing `waitForQueryDataResponse`, on the shared `GrafanaPage` base class so every page model (`ExplorePage`, `DashboardPage`, `PanelEditPage`, `AlertRuleEditPage`) gets it for free. It returns `{ response, body }`, and its predicate receives the parsed body so callers can filter by response contents (e.g. a specific refId's frames) rather than only status/url. Tagged `@alpha` following the precedent of `DashboardPage.waitForPanelsQueriesToComplete`, since the exact return shape may want to evolve.

The existing `waitForQueryDataResponse` is untouched — nothing about this is a breaking change.

None filed — this came directly out of review feedback on an external plugin's PR (see above). Happy to file one first if preferred.

- Verified locally against `grafana-enterprise:11.4.0` via `npm run server` + `npx playwright test`: the two new tests pass, and the full existing `queryEditor.integration.spec.ts`/`queryEditor.spec.ts` suites pass unchanged.
- Ran the full package test suite (240 tests) for regressions: 3 failures under default parallelism turned out to be pre-existing flakiness unrelated to this change — I confirmed by re-running the same specs with this change *reverted* (a *different, larger* set of 6 failed) and by re-running with `--workers=1` (all passed, both with and without this change). No test that exercises `GrafanaPage`/`waitForQueryDataResponse` regressed.
- Naming: went with `waitForQueryDataResponseWithBody` to mirror `clickhouse-datasource`'s own local helper, for continuity with the code this is generalising from. Open to a different name if you'd prefer something else.

### DIFF
--- a/docusaurus/docs/e2e-test-a-plugin/test-a-data-source-plugin/query-editor.md
+++ b/docusaurus/docs/e2e-test-a-plugin/test-a-data-source-plugin/query-editor.md
@@ -99,6 +99,58 @@ test('data query should return headers  and 3', async ({ panelEditPage, readProv
 });
 ```
 
+### Assert on the query response body
+
+A panel data assertion tells you what Grafana rendered. To assert on what your data source returned, use the [`<page>.waitForQueryDataResponseWithBody`](https://github.com/grafana/plugin-tools/blob/main/packages/plugin-e2e/src/models/pages/GrafanaPage.ts) method. It waits for a query data response and resolves with both the response and its parsed JSON body. The method reads the body while the response is still available, which is more reliable than calling `.json()` on a response that has already resolved.
+
+:::note
+This method is in alpha. The shape it returns can change without a major version bump.
+:::
+
+Call the method before you trigger the query, then await the result. The `body` property is `null` when the response body is not valid JSON.
+
+```ts title="queryEditor.spec.ts"
+type QueryResponse = {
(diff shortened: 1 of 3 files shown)


sunker (dismissed):
Nice 🚀  If you have a good, generic use case for this, you can add an example somewhere in the [ds docs](https://github.com/grafana/plugin-tools/tree/main/docusaurus/docs/e2e-test-a-plugin/test-a-data-source-plugin).

adamyeats:
@sunker Good call! Added some docs for this 👍

sunker (approved):
🚀

PR #2857, opened 2026-08-28
fix: wait for options pane before the viz picker check

jackw:
setVisualization works out whether the viz picker is already open by calling isVisible() on the picker's tab. That check doesn't wait - if the element isn't in the DOM yet it just returns false.

The Grafana panel editor has to wait for a lot of lazy chunks before the options pane renders, so for a few hundred milliseconds neither the tab nor the toggle button exists. `isVisible()` returns false, the method decides the picker must be closed, and clicks `toggle-viz-picker`. Except a new panel opens with the picker already open, and that toggle button only renders when the picker is closed - so it never render, and the click sits there until it times out.

This PR waits with locator.or() for whichever of the two appears first. One of them should always be there once the pane has rendered, so by the time we check open vs closed we should have something to work with.

Fixes #


sunker (approved):
Nice @jackw - I just faced this in my PR in grafana. Thanks for coming up with the fix!

PR #2862, opened 2026-09-08
feat(create-plugin): add CI drift check workflow to experimental-app-sdk

stevesg:
Installs a workflow to checks that there is no drift from `kinds/` to the
output generated from the `generate:kinds` script.

Tested in the linked repository:
Screenshot from 2026-09-08 22-27-47


packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
```diff
@@ -0,0 +1,61 @@
+name: Kinds drift check
+
+on:
```
sunker, line 3: This now runs rull install and codegen on every PR. Can you add a path filter on the kinds dir plus the generated dirs?

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
```diff
@@ -0,0 +1,61 @@
...
+        if: steps.check-for-backend.outputs.has-backend == 'true'
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Install dependencies
+        run: {{ packageManagerInstallCmd }}
+
+      - name: Generate kinds
+        run: {{ packageManagerName }} run generate:kinds
```
sunker, line 71: I guess this downloads the app-sdk cli on every run? Maybe there's a way to cache it?

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
```diff
@@ -0,0 +1,61 @@
...
+          node-version: '22'
+          cache: '{{ packageManagerName }}'
+
+      - name: Check for backend
+        id: check-for-backend
+        run: |
+          if [ -f "Magefile.go" ]
+          then
+            echo "has-backend=true" >> $GITHUB_OUTPUT
+          fi
```
jackw, line 53: Doesn't generation rely on Go?

mckn (commented):
I echo what Erik and Jack already have posted but I also have a question regarding using `git diff` for this workflow.

I guess the most common drift scenario — someone adds a new kind (or renames one) and forgets to git add the newly generated output — produces entirely new, untracked files. 

Wouldn't this check report success in that scenario?

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Install dependencies
+        run: {{ packageManagerInstallCmd }}
+
+      - name: Generate kinds
+        run: {{ packageManagerName }} run generate:kinds
+
+      - name: Fail if generated code is out of date
+        run: |
+          if ! git diff --quiet --exit-code; then
```
jackw: I think this git diff needs to focus on the kinds and generated kinds files. It will currently throw false positives for any plugin that uses Node 24 as the workflow will run an npm install using Node 22 which will edit the package-lock.json file..

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
stevesg: Nope - it's a pre-built binary, so Go isn't needed if you want to have Kinds in your frontend-only plugin.

stevesg:
> git diff

You're right this was a mistake :facepalm: - changed to `git status --porcelain`

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
stevesg: Good shout - would never have considered that

packages/create-plugin/templates/app-sdk/.github/workflows/generate-kinds-drift.yml
stevesg: Ah I see a comment in the script that indicates it does - I'll fix that.

jackw (approved):
LGTM! 🚀

stevesg:
Tested this in a plugin repository; it's detecting missing files now and caching appears to be working fine.

PR #2864, opened 2026-09-09
feat(plugin-docs-cli): add docs-path, size and SEO length validation rules

sunker:

This PR adds six validation rules to `plugin-docs-cli`: `docs-path-exists` (a missing `docsPath` in `plugin.json` now reports a proper diagnostic instead of a misleading `has-markdown-files` error), `max-total-pages` (50) and `max-total-docs-size` (10MB), `frontmatter-title-length` (60 chars) and `frontmatter-description-length` (20-160 chars) following standard SEO truncation conventions, and `min-content-length` (150 chars) flagging likely-stub pages. All follow the existing `serve`/`validate` severity pattern and were verified against a real plugin's docs to produce no false positives.

Fixes #

### DIFF
--- a/packages/plugin-docs-cli/src/bin/run.ts
+++ b/packages/plugin-docs-cli/src/bin/run.ts
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { access } from 'node:fs/promises';
+import { stat } from 'node:fs/promises';
 import minimist from 'minimist';
 import createDebug from 'debug';
 import { resolveDocsPath } from '../utils/utils.plugin.js';
@@ -36,11 +36,16 @@ async function main() {
   }
 
   // `serve` and `build` need a real folder to do anything useful, so fail fast here with a
-  // friendly message. `validate` reports a missing docsPath as a normal diagnostic instead
-  // (the `docs-path-exists` rule), so its --json output stays well-formed either way.
+  // friendly message. `validate` reports a missing or invalid docsPath as a normal diagnostic
+  // instead (the `docs-path-exists` rule), so its --json output stays well-formed either way.
   if (command !== 'validate') {
     try {
-      await access(docsPath);
+      const st = await stat(docsPath);
+      if (!st.isDirectory()) {
+        console.error(`Error: Not a directory: ${docsPath}`);
+        console.error('Check that the "docsPath" in src/plugin.json points to a directory, not a file.');
+        process.exit(1);
+      }
     } catch {
       console.error(`Error: Path not found: ${docsPath}`);
       console.error('Check that the "docsPath" in src/plugin.json points to an existing directory.');
--- a/packages/plugin-docs-cli/src/bin/run.ts
+++ b/packages/plugin-docs-cli/src/bin/run.ts
@@ -35,12 +35,17 @@ async function main() {
     process.exit(1);
   }
 
-  try {
-    await access(docsPath);
-  } catch {
-    console.error(`Error: Path not found: ${docsPath}`);
-    console.error('Check that the "docsPath" in src/plugin.json points to an existing directory.');
-    process.exit(1);
+  // `serve` and `build` need a real folder to do anything useful, so fail fast here with a
+  // friendly message. `validate` reports a missing docsPath as a normal diagnostic instead
+  // (the `docs-path-exists` rule), so its --json output stays well-formed either way.
+  if (command !== 'validate') {
+    try {
+      await access(docsPath);
+    } catch {
+      console.error(`Error: Path not found: ${docsPath}`);
+      console.error('Check that the "docsPath" in src/plugin.json points to an existing directory.');
+      process.exit(1);
+    }
   }
 
   switch (command) {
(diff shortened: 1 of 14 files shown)


packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -10,20 +10,52 @@ const SLUG_SAFE_RE = /^[a-z0-9-]+$/;
...
+    if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') {
+      diagnostics.push({
+        rule: Rule.DocsPathExists,
+        severity: 'error',
+        title: 'docsPath does not exist',
+        detail: `"${input.docsPath}" does not exist. Check that "docsPath" in src/plugin.json points to an existing directory.`,
+      });
+    }
+    return diagnostics;
   }
```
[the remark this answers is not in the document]
sunker: fixed in 428e9286

packages/plugin-docs-cli/src/validation/rules/frontmatter.test.ts
```diff
@@ -7,10 +7,20 @@ import { Rule } from '../types.js';
 
 const input = (docsPath: string) => ({ docsPath, strict: true });
 
+// a description long enough to clear the 20-char min-content-length floor introduced by
+// frontmatter-description-length, so tests unrelated to that rule don't trip it incidentally
```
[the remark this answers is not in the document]
sunker: Fixed in 574d3a74

packages/plugin-docs-cli/src/bin/run.ts
```diff
@@ -35,12 +35,22 @@ async function main() {
...
+        console.error(`Error: Not a directory: ${docsPath}`);
+        console.error('Check that the "docsPath" in src/plugin.json points to a directory, not a file.');
+        process.exit(1);
+      }
+    } catch {
+      console.error(`Error: Path not found: ${docsPath}`);
+      console.error('Check that the "docsPath" in src/plugin.json points to an existing directory.');
+      process.exit(1);
+    }
   }
```
mckn, line 55: Hmm, should we also present a more user friendly error message when docsPath can't be resolved? Or is that overkill?

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -87,6 +128,40 @@ export async function checkFilesystem(input: ValidationInput): Promise<Diagnosti
...
+
+  // max-total-docs-size: only checked under `validate` (not `serve`)
+  if (input.strict) {
+    let totalSize = 0;
+    for (const file of entries) {
+      if (!file.isFile()) {
+        continue;
+      }
+      try {
+        const st = await stat(join(file.parentPath, file.name));
```
mckn, line 134: Do we need to care about efficiency here? Or it won't be that many paths to check?

Was thinking if we should do this in parallell or not.

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
```diff
@@ -10,20 +10,61 @@ const SLUG_SAFE_RE = /^[a-z0-9-]+$/;
...
 // permitted image formats shared across filesystem and asset rules
 export const ALLOWED_IMAGE_EXTENSIONS = new Set(['.png', '.jpg', '.jpeg', '.webp', '.gif']);
 
 // allowed file extensions in the docs folder (.md + permitted image formats)
 export const ALLOWED_EXTENSIONS = new Set(['.md', ...ALLOWED_IMAGE_EXTENSIONS]);
 
+/**
+ * Formats a byte count as a human-readable string.
+ */
+function formatBytes(bytes: number): string {
```
mckn: Could we move this to a utility function that we can use in this file and in assets.ts?

mckn (dismissed):
LGTM! Left some thoughts but overall it looks solid!

packages/plugin-docs-cli/src/bin/run.ts
sunker: Isn't this quite clear? `Check that the "docsPath" in src/plugin.json points to an existing directory.` Would be happy to improve if you have ideas :)

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
sunker: I think this is fine - most plugins will only have 4-6 pages so don't think we need to go to far in terms of efficiency here.

packages/plugin-docs-cli/src/validation/rules/filesystem.ts
sunker: Good point! Addressed in https://github.com/grafana/plugin-tools/pull/2864/changes/10e1c1ddbb5adc6ef98ab0f6d42c407b728f5a36

packages/plugin-docs-cli/src/bin/run.ts
mckn: Ah, sorry for not being that clear. We try to resolve the docsPath a couple of lines (33) up in this file.

mckn (approved):
LGTM! Added a smal nit that you can fix if you think it makes sense

PR #2525, opened 2026-03-10
fix: prevent write errors in codemods for new updated files

jackw:
**What this PR does / why we need it**:

If a codemod adds a file then later updates the same file the changeType is changed from `add` to `update`. This can cause codemods to fail when writing to disk due to `flushChanges` expecting the containing directory to exist for `update` files as they _should_ already exist on disk.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install website@5.5.2-canary.2525.22911401478.0
  npm install @grafana/create-plugin@7.0.7-canary.2525.22911401478.0
  npm install @grafana/eslint-plugin-plugins@0.6.4-canary.2525.22911401478.0
  npm install @grafana/plugin-docs-cli@0.0.8-canary.2525.22911401478.0
  # or 
  yarn add website@5.5.2-canary.2525.22911401478.0
  yarn add @grafana/create-plugin@7.0.7-canary.2525.22911401478.0
  yarn add @grafana/eslint-plugin-plugins@0.6.4-canary.2525.22911401478.0
  yarn add @grafana/plugin-docs-cli@0.0.8-canary.2525.22911401478.0
  ```
</details>

PR #2865, opened 2026-09-09
fix(create-plugin): preserve 'add' changeType when updating a newly added file

stevesg:
Context.updateFile() previously downgraded 'add' to 'update' whenever content
changed, even if the file was staged in the same context. Since flushChanges
relies on changeType to decide whether to mkdir the parent directory, this
caused writes to fail for files added and then updated before being flushed.

### DIFF
--- a/packages/create-plugin/src/codemods/context.test.ts
+++ b/packages/create-plugin/src/codemods/context.test.ts
@@ -62,11 +62,15 @@ describe('Context', () => {
 
   describe('updateFile', () => {
     it('should update a file in the context', () => {
+      const context = new Context(`${__dirname}/migrations/fixtures`);
+      context.updateFile('foo/bar.ts', 'new content');
+      expect(context.listChanges()).toEqual({ 'foo/bar.ts': { content: 'new content', changeType: 'update' } });
+    });
+
+    it("should preserve the 'add' changeType when updating a file that was added in the current context", () => {
       const context = new Context();
       context.addFile('file.txt', 'content');
       context.updateFile('file.txt', 'new content');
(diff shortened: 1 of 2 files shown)


packages/create-plugin/src/codemods/context.test.ts
```diff
@@ -65,7 +65,9 @@ describe('Context', () => {
       const context = new Context();
       context.addFile('file.txt', 'content');
       context.updateFile('file.txt', 'new content');
-      expect(context.listChanges()).toEqual({ 'file.txt': { content: 'new content', changeType: 'update' } });
+      // Stays 'add': the file was staged fresh in this context, so flushChanges still needs to
+      // mkdir its (possibly nonexistent) parent directory rather than treat it as an in-place update.
+      expect(context.listChanges()).toEqual({ 'file.txt': { content: 'new content', changeType: 'add' } });
```
jackw, line 74: Can we revert this and test these two paths separately please?

```
it("should preserve the 'add' changeType when updating a file that was added in the current context", () => {
    const context = new Context();
    context.addFile('file.txt', 'content');
    context.updateFile('file.txt', 'new content');
    expect(context.listChanges()).toEqual({ 'file.txt': { content: 'new content', changeType: 'add' } });
  });
});
```

jackw (commented):
Thanks! 🏅 
I thought I fixed this way back when and never merged - https://github.com/grafana/plugin-tools/pull/2525 🤦 I've closed in favour of this.

packages/create-plugin/src/codemods/context.test.ts
stevesg: Done :)

jackw (approved):
LGTM! 🚀

PR #2866, opened 2026-09-09
feat(plugin-docs-cli): add writing-style validation rules

sunker:

Grafana's docs team publishes its documentation style guide as Vale rules in the [Writers' Toolkit](https://grafana.com/docs/writers-toolkit/write/style-guide/). Running Vale itself would require every plugin author and CI to install and maintain the Go binary, so this PR skips that and transcribes 35 of the toolkit's rules by hand into `plugin-docs-cli`, applying them as `style-*` findings that are always advice, never a blocking error, and always link back to the toolkit's docs page. It also fixes `formatResult()` to sort diagnostics by severity, since an error could otherwise print below unrelated warnings.

Fixes #

### DIFF
--- a/packages/plugin-docs-cli/src/server/server.ts
+++ b/packages/plugin-docs-cli/src/server/server.ts
@@ -44,12 +44,29 @@ export async function startServer(options: ServerOptions): Promise<Server> {
   app.set('view engine', 'ejs');
   app.set('views', join(__dirname, 'views'));
 
-  // run validation in non-strict mode and log results without blocking
+  // run validation in non-strict mode and log results without blocking. this reruns on every
+  // save, so writing-style findings are collapsed to a single line - printing all of them would
+  // bury the structural problems the author actually has to fix.
   const runValidation = async () => {
     try {
       const result = await validate({ docsPath, strict: false }, allRules);
-      if (result.diagnostics.length > 0) {
-        console.log(formatResult(result));
+      const styleFindings = result.diagnostics.filter((d) => d.rule.startsWith('style-'));
+      const rest = { ...result, diagnostics: result.diagnostics.filter((d) => !d.rule.startsWith('style-')) };
+
+      if (rest.diagnostics.length > 0) {
+        console.log(formatResult(rest));
+      }
+      if (styleFindings.length > 0) {
+        const warnings = styleFindings.filter((d) => d.severity === 'warning').length;
+        const suggestions = styleFindings.length - warnings;
+        const parts: string[] = [];
+        if (warnings > 0) {
+          parts.push(`${warnings} writing-style warning${warnings === 1 ? '' : 's'}`);
+        }
+        if (suggestions > 0) {
+          parts.push(`${suggestions} suggestion${suggestions === 1 ? '' : 's'}`);
+        }
+        console.log(`  ${parts.join(', ')} - run \`docs:validate\` to see them\n`);
       }
     } catch (error) {
       console.error('Validation failed:', error instanceof Error ? error.message : error);
--- a/packages/plugin-docs-cli/src/validation/rules/style.ts
+++ b/packages/plugin-docs-cli/src/validation/rules/style.ts
@@ -372,7 +372,14 @@ export async function checkWritingStyle(input: ValidationInput): Promise<Diagnos
     return diagnostics;
   }
 
-  const mdFiles = entries.filter((e) => e.isFile() && e.name.endsWith('.md') && !isMetaFile(e.name));
+  const mdFiles = entries.filter(
+    (e) =>
+      e.isFile() &&
+      e.name.endsWith('.md') &&
+      !isMetaFile(e.name) &&
+      !e.parentPath.includes('node_modules') &&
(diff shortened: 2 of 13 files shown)


packages/plugin-docs-cli/src/validation/rules/utils.ts
```diff
@@ -69,3 +69,122 @@ export function getCodeBlockLines(content: string): Set<number> {
...
+    }
+    const lineRe = new RegExp(re.source, re.flags);
+    let m: RegExpExecArray | null;
+    while ((m = lineRe.exec(lineText)) !== null) {
+      results.push({ match: m, line: i + 1 });
+      // guard against a zero-length match spinning forever
+      if (m[0] === '') {
+        lineRe.lastIndex++;
+      }
+    }
```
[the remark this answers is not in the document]
sunker: done

packages/plugin-docs-cli/src/validation/rules/style.ts
```diff
@@ -0,0 +1,450 @@
...
+  const diagnostics: Diagnostic[] = [];
+
+  let entries: Dirent[] = [];
+  try {
+    entries = await readdir(input.docsPath, { recursive: true, withFileTypes: true });
+  } catch {
+    return diagnostics;
+  }
+
+  const mdFiles = entries.filter((e) => e.isFile() && e.name.endsWith('.md') && !isMetaFile(e.name));
```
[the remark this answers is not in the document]
sunker: done

packages/plugin-docs-cli/src/server/server.ts
```diff
@@ -44,12 +44,29 @@ export async function startServer(options: ServerOptions): Promise<Server> {
...
+        const warnings = styleFindings.filter((d) => d.severity === 'warning').length;
+        const suggestions = styleFindings.length - warnings;
+        const parts: string[] = [];
+        if (warnings > 0) {
+          parts.push(`${warnings} writing-style warning${warnings === 1 ? '' : 's'}`);
+        }
+        if (suggestions > 0) {
+          parts.push(`${suggestions} suggestion${suggestions === 1 ? '' : 's'}`);
+        }
+        console.log(`  ${parts.join(', ')} - run \`docs:validate\` to see them\n`);
```
leventebalogh, line 69: Should we log these as warnings instead of info?

packages/plugin-docs-cli/src/validation/rules/utils.test.ts
```diff
@@ -94,3 +102,94 @@ describe('maskInlineCode', () => {
     expect(masked).toHaveLength(line.length);
   });
 });
+
+describe('maskLinkTargets', () => {
```
leventebalogh, line 106: Does this work with bare links (e.g. `https://github.com/org/repo`) or with refs (`[ref]: ./path.md`) as well?

packages/plugin-docs-cli/src/validation/rules/style.ts
```diff
@@ -0,0 +1,450 @@
...
+      continue;
+    }
+
+    const codeLines = getCodeBlockLines(content);
+    const skipLines = getNonProseLines(content);
+    const headingLines = getHeadingLines(content, codeLines);
+    const lines = content.split('\n');
+    const counts = new Map<string, number>();
+
+    for (const matcher of MATCHERS) {
```
leventebalogh, line 898: Nit: should we log the findings based on line number (as they come in the document) instead of the order of the rule matchers?

leventebalogh (dismissed):
LGTM 👍 

I was wondering how we are planning to keep this in sync with the vale ruleset curated by the docs team, or if you think that's ever going to be an issue.

packages/plugin-docs-cli/src/validation/rules/style.ts
sunker: Yes I think that makes sense, will fix!

packages/plugin-docs-cli/src/server/server.ts
sunker: Hmm I think this is fine. The rule groups can produce false positive so not sure it's worth the noise of having warnings for them tbh.

packages/plugin-docs-cli/src/validation/rules/utils.test.ts
sunker: Looks like you're right, will fix!

sunker:
> I was wondering how we are planning to keep this in sync with the vale ruleset curated by the docs team, or if you think that's ever going to be an issue.

I initially had a sync script, but it got hugely complicated so I decided to drop it, mostly because the rules in the writer's kit rarely changes. Also I think an agent will be able to figure out quite easily. I described the syncing process shortly in the agents.md file, but it ended up in the corresponding docs PR - see [here](https://github.com/grafana/plugin-tools/blob/934849532cb5c5bacb2f7fae3e6a5f6519bcd3d4/packages/plugin-docs-cli/docs/validation-rules.md?plain=1#L90).

leventebalogh (approved):
I see, LGTM 👍

PR #2872, opened 2026-09-11
fix(create-plugin): drop shebang from generate-kinds.mjs so it gets the do-not-edit header

stevesg:
Fix as suggested by @jackw 


jackw (approved):
LGTM! 🚀

PR #2873, opened 2026-09-11
fix(create-plugin): support @grafana/eslint-config v10 in scaffold

### LINKED ISSUE #2871: Bug: eslint.config.mjs scaffold still imports removed @grafana/eslint-config/flat.js subpath (breaks on eslint-config v10)

- [x] Create Plugin
- [x] ESLint Config

### Package versions

```
@grafana/create-plugin (scaffold): 7.10.0 (.config/.cprc.json) -- confirmed the same issue exists in the latest published @grafana/create-plugin@7.10.1 template as well
@grafana/eslint-config: 9.0.0 -> 10.0.0
eslint: 9.39.5
```

### What happened?

`@grafana/eslint-config@10.0.0` removed the `./flat.js` subpath export entirely -- the flat config array is now the plain default export (`import grafanaConfig from '@grafana/eslint-config'`), not `@grafana/eslint-config/flat.js`. The scaffolded `.config/eslint.config.mjs` file (which is marked "DO NOT EDIT THIS FILE DIRECTLY" and is meant to be managed by `create-plugin`) still imports the old subpath:

```js
import grafanaConfig from '@grafana/eslint-config/flat.js';
```

Running `eslint` after bumping `@grafana/eslint-config` to `10.0.0` crashes immediately with:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './flat.js' is not defined by "exports" in
node_modules/@grafana/eslint-config/package.json imported from .config/eslint.config.mjs
```

I checked whether running `npx @grafana/create-plugin@latest update` would fix this by extracting the package template directly (`@grafana/create-plugin@7.10.1`, the latest published version at the time of filing): it still hardcodes `import grafanaConfig from '@grafana/eslint-config/flat.js';` in `templates/common/.config/eslint.config.mjs`, and its own `templates/common/_package.json` still pins `"@grafana/eslint-config": "^9.0.0"`. So there's currently no way to update to `@grafana/eslint-config@10` without either hand-editing the scaffolded file (against its own warning) or forking the Grafana config import out of `.config/` entirely into the plugin's own root `eslint.config.mjs`.

Two other Grafana-maintained plugins have already worked around this in-repo:
- [grafana/clock-panel](https://github.com/grafana/clock-panel/blob/main/eslint.config.mjs) moved the `@grafana/eslint-config` import out of `.config/` entirely, into its own root config (no `.config/eslint.config.mjs` exists in that repo at all).
- [grafana/metrics-drilldown](https://github.com/grafana/metrics-drilldown/blob/main/.config/eslint.config.mjs) hand-edited the scaffolded file directly to drop `/flat.js`.

Separately (not part of this report, but worth noting for anyone attempting the same upgrade): once `.config/eslint.config.mjs` is fixed, `eslint-plugin-react@7.37.5` (peer-required by `@grafana/eslint-config`) crashes at runtime under `eslint@10.x` (confirmed against `10.0.0` through `10.10.0`) with `TypeError: contextOrFilename.getFilename is not a function`, from `eslint-plugin-react`'s `detectReactVersion()`/`resolveBasedir()` internals, triggered because `@grafana/eslint-config`'s bundled settings use `react: { version: "detect" }`. Pinning an explicit `react.version` in a consuming config avoids that specific crash, but it's a separate compatibility gap in `eslint-plugin-react`, not something `create-plugin` can fix directly.

Bumping `@grafana/eslint-config` to `10.x` should work with a scaffold generated (or updated via `create-plugin update`) at the time `eslint-config@10` is current, without needing to hand-edit or bypass the "DO NOT EDIT" `.config/eslint.config.mjs` file.

1. Scaffold (or use an existing) app plugin with `@grafana/create-plugin@7.10.1` (or run `npx @grafana/create-plugin@latest update` on an existing one).
2. `pnpm add -D @grafana/eslint-config@10.0.0 @stylistic/eslint-plugin@5.10.0` and remove the now-unused `@stylistic/eslint-plugin-ts` (a peer requirement of `@grafana/eslint-config@10`).
3. Run `pnpm exec eslint .` (or `pnpm lint`).
4. See `ERR_PACKAGE_PATH_NOT_EXPORTED` for `@grafana/eslint-config/flat.js`, thrown from the scaffolded `.config/eslint.config.mjs`.

### Environment

```
node: v24.11.0
pnpm: 11.11.0
OS: Darwin 25.6.0 arm64 (macOS)
```

### Additional context

Found while trying to pick up the two open Renovate PRs updating `eslint` and `@grafana/eslint-config` to v10 in [grafana/logs-drilldown](https://github.com/grafana/logs-drilldown). Happy to open a PR against the `create-plugin` eslint template if that's the preferred fix direction once maintainers weigh in on whether the template should import the package root directly (dropping `/flat.js`) going forward.

L2D2Grafana:
## Summary 📝

- Fixes #2871: the scaffolded `.config/eslint.config.mjs` imported `@grafana/eslint-config/flat.js`, a subpath removed in `@grafana/eslint-config@10.0.0`, crashing `eslint` with `ERR_PACKAGE_PATH_NOT_EXPORTED`. New scaffolds now import from `@grafana/eslint-config` directly and pin `^10.0.0`, with `@stylistic/eslint-plugin-ts` swapped for the `@stylistic/eslint-plugin` peer dependency it now requires.
- Adds migration `014-eslint-config-v10` so existing plugins get the same fix via `create-plugin update`.

## Test 🧪

- [x] `npm run test -w @grafana/create-plugin -- --run` (309 passed, including 6 new tests for the migration)
- [x] `npm run typecheck -w @grafana/create-plugin` and `npm run lint -w @grafana/create-plugin`
- [x] Scaffolded a fresh app plugin from the built CLI, ran `npm install && npm run lint` — passes cleanly with `@grafana/eslint-config@10.0.0` (previously crashed)
- [x] Ran `create-plugin update` against an existing plugin (logs-drilldown) and confirmed the migration rewrites `.config/eslint.config.mjs` and `package.json` as expected


packages/create-plugin/src/codemods/migrations/migrations.ts
```diff
@@ -91,6 +91,13 @@ export default [
       'Support @grafana/ui@13.2.0 in Jest by transforming @react-hookz/web and @ver0/deep-equal ESM dependencies.',
     scriptPath: import.meta.resolve('./scripts/013-jest-esmodules.js'),
   },
+  {
+    name: '014-eslint-config-v10',
+    version: '7.10.2',
```
jackw: I really need to fix this but... please could you sync this branch with main and then update this version to `7.11.1` to guarantee it'll run in [next release version](https://github.com/grafana/plugin-tools/pull/2891)?

jackw (commented):
Awesome work @L2D2Grafana. Can you fix the version in migrations.ts please then we should be good to merge.

packages/create-plugin/src/codemods/migrations/migrations.ts
L2D2Grafana: Updated!

jackw (approved):
LGTM! 🚀

Took it for a spin locally and working as expected.

Image

PR #2878, opened 2026-09-14
fix(create-plugin): defer add-command success message until after summary

stevesg:
Lets a codemod optionally register a success message via
`context.setMessage()` that prints after the code mod runs. This means codemods can match the ordering of the initial generate step, where "Next steps:" comes at the end of the output. When they are printed during the codemod itself, the text can scroll past and is easy to miss.


packages/create-plugin/src/codemods/context.ts
```diff
@@ -14,12 +14,30 @@ export type ContextFile = Record<
 
 export class Context {
   private files: ContextFile = {};
+  private successMessage?: () => void;
```
jackw: Maybe we could return something that fits with the `output` logger so the add command controls the formatting and the codemod can declare different status levels. e.g.

```js
private message?: { level: 'error' | 'success', title: string; body?: string[]; withPrefix?: boolean }
```

then in the add command we can get the message and if it exists pass it to the output util...

```js
if (msg) {
  output[msg.level]({
    title: msg.title,
    body: msg.body,
  });
} else {
// do whatever command currently does
}
```

stevesg: SGTM - gave this a try in https://github.com/grafana/plugin-tools/pull/2878/commits/15b2d441d19ba01eeac54a0127e797135f711e94

packages/create-plugin/src/commands/add.command.ts
```diff
@@ -24,11 +24,17 @@ export const add = async (argv: minimist.ParsedArgs) => {
 
     // filter out minimist internal properties (_ and $0) before passing to codemod
     const { _, $0, ...codemodOptions } = argv;
-    await runCodemod(addition, codemodOptions);
+    const context = await runCodemod(addition, codemodOptions);
 
     output.success({
       title: `Successfully added ${addition.name} to your plugin.`,
     });
```
jackw, line 31: I think we wanna move this success to an `else` if no message is set by the addition so it becomes the "fallback" otherwise this command prints success but the addition might print an error / warning next?

jackw (commented):
Looks great. Thanks for tackling this - it's something I've been wanting to add for a while now.

It might be worth updating the claude `/write-codemod` skill (in this repo) to include instructions for setting messages so future additions are written with this in mind?

packages/create-plugin/src/commands/add.command.ts
stevesg: Ah right I see - so we can replace the success message entirely... that makes sense. Currently it would yield e.g. this:

```
..
..
Installing NPM dependencies...

 create-plugin@7.10.1  Successfully added experimental-app-sdk to your plugin.

———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 create-plugin@7.10.1  Added grafana-app-sdk code generation. Next steps:

Edit your kinds in ./kinds (start with kinds/example.cue), then run:
..
..
```

stevesg:
Ready to go @jackw

jackw (approved):
LGTM! 🚀

PR #2879, opened 2026-09-14
chore(create-plugin): bump grafana-app-sdk to v0.60.0

stevesg:
All changes needed are in this released version now, so the temporary versions can be cleaned up.


sunker (approved):
🚀

PR #2884, opened 2026-09-15
feat(create-plugin): run go mod tidy for codemods editing go.mod

stevesg:
Follow up on a suggestion from @jackw so `go.mod` edit handling is similar to `package.json`.

For this to work, scaffolded Go code has to be able to pass `go mod tidy`, but does not have to actually compile. Therefore, some stub files are required to declare the imports for the eventual generated code.


packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
```diff
@@ -279,6 +281,29 @@ function addAppProvider(context: Context) {
   context.addFile(path, renderTemplate(templatePath(path), false));
 }
 
+// Stub doc.go files under pkg/generated/. Both are real packages `generate:kinds` fills in; they
+// exist solely so the `go mod tidy` the codemod runner runs automatically (see runGoModTidy in
+// ../../utils.js) succeeds before code generation has ever run, since it otherwise can't resolve the
+// packages pkg/provider/provider.go imports. `generate:kinds` writes its own, differently-named files
+// into these directories rather than overwriting doc.go, so it's left behind afterwards — but an
+// unused doc.go with no exported symbols compiles fine alongside the generated code.
+const GENERATED_STUB_PATHS = ['pkg/generated/example/v1alpha1/doc.go', 'pkg/generated/manifestdata/doc.go'];
```
jackw, line 290: Just for my understanding - these files are only needed for the codemod to succeed because without them the `go mod tidy` would fail? Is a user expected to delete them once the codemod is run?

jackw (dismissed):
LGTM! 🚀

Left a question for my own understanding.

packages/create-plugin/src/codemods/additions/scripts/experimental-app-sdk.ts
stevesg: They are harmless, but not necessary after code generation.... there might be a better approach we can come up with here in time.

jackw: Sure thing. Just wanted to know in case questions come in later down the line.

jackw (approved):
LGTM! 🚀

PR #2885, opened 2026-09-16
feat(plugin-e2e): log when runtime selectors are used

sunker:

This PR logs `using runtime selectors from <url>` when the `@grafana/plugin-e2e` selectors fixture successfully uses the runtime-fetched selectors, so CI logs positively show the runtime path was taken rather than a silent bundled fallback. This line will be removed once we're certain things work as expected. 

Fixes #


mckn (approved):
LGTM!

PR #2886, opened 2026-09-16
feat(plugin-e2e): enable runtime e2e-selectors by default

sunker:

This PR makes runtime e2e-selectors the default in `@grafana/plugin-e2e`, so tests resolve selectors from the Grafana under test rather than the bundled `@grafana/e2e-selectors` snapshot. It is a no-op against Grafana below 13.3.0, which still falls back to the bundled selectors.

Fixes #

If plugin-e2e can't resolve selectors from grafana 13.3+ where runtime selectors should be build assets, it will log an error in CI before falling back to bundled selectors.


mckn (approved):
LGTM!

